PR #10, opened 2013-04-21
FreezeOperation

lecoqlibre:
FreezeOperation class basicly implemented. Look at the method Editor::unfreeze_route from gtk2_ardour/editor_ops.cc.


x42:
This no longer applies :(

PR #56, opened 2013-11-11
French translation reviewed

jpeg729:
All mis-translations removed or corrected. It isn't yet what I'd
call a good translation, but at least it isn't wrong.


x42:
closed per request in #105

PR #63, opened 2013-12-31
added feature to automatically apply a delay to panned channels

SReichelt:
The feature comes in the form of a new session property that specifies the delay time in ms/100%, 0 being the default so nothing changes for existing sessions. The delay is applied to the output channel with the lower percentage, proportionally to the difference in percentage between the two channels.

At the moment, the feature has been implemented for the 1in2out and 2in2out panners. The width control of the 2in2out panner is ignored. VBAP support can easily be added later.

Other minor changes made along the way:
- Use sqrt(1/2) instead of -3 dB for the pan law, as it is more accurate and available as a compile-time constant.
- The gain interpolation that is used when the panner is moved no longer remembers any state across calls to distribute_one(). This seemed to be unintentional and slightly broken.

To try this feature, open an existing session that includes panned tracks. Start playback, preferably through headphones, open the "Misc" tab of the session properties, and drag the slider to the right.


SReichelt:
I just rebased my cloned repository to Ardour 4. Not sure if GitHub handled this correctly. There should be just two commits now. Any chance of getting this into mainline?

x42:
The issue with this is that it modifies existing panners.
This should really be a dedicated new panner module (now that ardour allows to select and choose panners).

SReichelt:
That was true in the original implementation, but I changed it during the discussion above. The panners are selectable now. The remaining modifications to existing code are just to avoid code duplication.

Anutrix:
Status?

luzpaz:
Is this PR still relevant ?

SReichelt:
Not sure if you are asking me. I can't really answer that question, it depends on whether this feature is wanted or not.

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #104, opened 2015-01-05
Old session loading

sbergen:
Fixed some issues I found with loading an old session file.

The file had &lt;Session version="0.908.2" ... &gt; in it and it was recorded with some version of A2 IIRC. Not sure if the code is 100% correct even like this...

While at it, I added a utility class for finding Statefuls by id with standard algorithms, as the code is much more readable like this IMO.

### DIFF
--- a/libs/ardour/location.cc
+++ b/libs/ardour/location.cc
@@ -29,6 +29,7 @@
 #include "pbd/stl_delete.h"
 #include "pbd/xml++.h"
 #include "pbd/enumwriter.h"
+#include "pbd/statefulidpredicates.h"
 
 #include "ardour/location.h"
 #include "ardour/midi_scene_change.h"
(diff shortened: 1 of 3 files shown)


mojofunk:
I'm not sure what the policy is on supporting loading of older sessions especially very old sessions that may cause other issues with current versions. If this is accepted it would be nice if we could add the session file to the libardour tests.

The refactoring looks OK but I'm not sure that you are using the correct visibility/export macro though, I don't think you need any for non-template header only classes. I think I'd also prefer that class was defined in the pbd/stateful.h header.

x42:
Do you still have this ancient session?

In general Ardour should be able to load and convert old sessions, but  v0.x.x sessions are likely not reasonable to support. Still detecting the old session format would still be useful.

luzpaz:
ping @sbergen

sbergen:
I found some old sessions in my archives with a matching session version, but I can't be sure if one of them is this specific one. @x42 are the xml session files enough for what you'd like to do, or do you need something else also?

sdevisch (approved):

sdevisch (approved):
Approved.

PR #106, opened 2015-03-13
Add --program-exe-name option to waf build options

mojofunk:
Use an option to determine executable name(prefix) and also use that for config dir names. Also fixed with this branch should be some hard coded version numbers. With --program-name and --program-exe-name left at the default options this should be a no-op change.

I have not tested the packaging scripts with this change, although there shouldn't be any problems with the defaults.

### DIFF
--- a/gtk2_ardour/ardour.sh.in
+++ b/gtk2_ardour/ardour.sh.in
@@ -23,7 +23,7 @@ fi
 
 ## Glib atomic test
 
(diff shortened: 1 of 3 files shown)


pauldavisthefirst:
i'm confused about the purpose of this change set.

mojofunk:
Although I think this might have been useful the issue that was addressing has since been fixed in another/similar fashion

PR #108, opened 2015-04-15
Export chapter marks in mp4chaps

johannes-mueller:
The mp4 file format supports chapter marks using the so called mp4chaps format to enable chapter wise navigation in an mp4 file. The format is like hh:mm:ss.sss Chapter Title
    
This branch adds the ability to export those kind of chapter marks  along with TOC and CUE marks. The filename extension for the chapter mark file is "chapters.txt". The format specification description is "MP4ch".

An "Intro" chapter is introduced at position zero as some players (like vlc) seem to rely on the existence of a chapter mark at 00:00:00.000

### DIFF
--- a/gtk2_ardour/export_format_dialog.cc
+++ b/gtk2_ardour/export_format_dialog.cc
@@ -69,6 +69,7 @@ ExportFormatDialog::ExportFormatDialog (FormatPtr format, bool new_dialog) :
 
   with_cue (_("Create CUE file for disk-at-once CD/DVD creation")),
   with_toc (_("Create TOC file for disk-at-once CD/DVD creation")),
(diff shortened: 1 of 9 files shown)


x42:
looks good. thanks!

Given that 4.0-rc4 was just tagged, I think we should postpone pulling this after 4.0 is out (next 24-48 h) though.

johannes-mueller:
I added one tiny commit that tweaks the file name of the chapter mark file a little bit, so that handling with the mp4chaps tool is a bit more convenient.

x42:
rebased onto master as acd1ee1 .. 4.0-60-g9ba98ae   for now the 'not strip part' is included, it does seem convenient (and we can easily revert or amend it if needed.).

thank you.

PR #109, opened 2015-04-17
Update editor_rulers.cc

ahellquist:
Change ruler context actions from "Make Loop range" to "New Loop range"
change ruler context actions from "Make Punch range" to "New Loop range"

This to make it consistent with other ruler context actions eg. "New....."


ahellquist:
This is my very first pull-request ever BTW

x42:
thanks rebased as 80d8e86

PR #110, opened 2015-04-22
Added transifex configuration

rdoursenaud:
[Transifex](http://transifex.com) is a web application allowing translators to work collaboratively.

I hereby suggest that we use it to coordinate the translation effort.

The project can be found at https://www.transifex.com/organization/ardour.

I've uploaded all the current translations.

This configuration allows using the [client](http://docs.transifex.com/developer/client) to easily push the source language files (```tx push -s```) and pull the translations (```tx pull -t```) for the project maintainer @pauldavisthefirst.

Translators can use the [web UI](https://www.transifex.com/projects/p/ardour/translate) to easily and efficiently collaborate to translate the strings in their languages without too much technical knowledge. All the heavy lifting is done for them.

Let me know what you think!

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,0 +1,17 @@
+[main]
+host = https://www.transifex.com
+
+[ardour.ardour]
+file_filter = gtk2_ardour/po/<lang>.po
+source_file = gtk2_ardour/gtk2_ardour3.pot
+source_lang = en
+
+[ardour.libardour]
+file_filter = libs/ardour/po/<lang>.po
+source_file = libs/ardour/ardour3.pot
+source_lang = en
+
+[ardour.gtkmm2ext]
+file_filter = libs/gtkmm2ext/po/<lang>.po
+source_file = libs/gtkmm2ext/gtkmm2ext3.pot
+source_lang = en


x42:
30 day free trial..  and then?

prokoudine:
Hello,

I'm one of the translators, and I'd like to say that much as I appreciate your effort to help us, this is likely only to make things worse. My explanation is going to be long, so I'll be grateful  if you read it in its entirety, even if you won't agree with me in the end.

See, I've been doing i18n for free software projects for the past 13 years, and one thing I've learned is that in terms of making end-users livid, an incorrectly translated message can beat an untranslated message hands down. And with online translation tools, there will be no end to incorrect translations.

This happens mostly because with online tools, you cannot easily test your translation. You need to do some extra work to get what you see online installed in your system, so that you could view the change you made live. And this is really important, because if you don't test your translation, you won't do a very good job.

With online tools, contributors tend go easy on themselves: not checking where the message actually appears, not checking if a context switch has to be added to source code before translation should happen (it used to be  quite an issue for Ardour until fairly recently), not checking if they should try to make their translation shorter (because it makes UI huge), and the list goes on. Basically, if anyone can translate, then anyone will. This leads to lousy translations, and lousy translations undermine the very idea of delivering i18n.

Before you argue, let me tell you that I've seen this over and over (and over) again with many projects on different platforms - Ubuntu's Rosetta, Transifex, Crowdin etc.. The idea that "many eyes improve translation" doesn't really work for big projects such as Ardour. With online tools, you can get more complete translations in terms of translated/untranslated messages ratio, but you will also lose quality.

With Transifex this is even worse, because (unlike e.g. Crowdin) it's doesn't build a log of changes, and this means that you have no idea what was changed by someone, so you need to go through the whole damn thing all at once before you can tell that a translation with contributions from arbitrary people is good to go. I had a very similar experience with my translation of Ardour recently, when I was sent updated files, and a diff couldn't be made for technical reasons. It was hell, and v4.0 was released with a few bloopers anyway.

If other translators disagree with me, then at least please make it possible for certain translations to never be uploaded to Transifex. That way, everyone will do their job the way they see fit.

pauldavisthefirst:
Based on Prokoudine's observations, I'm going to pass on this pull request.

PR #111, opened 2015-04-30
Don't install static library libsmf.a

nphilipp:
As the subject says: if libsmf is built statically and linked into libevoral.so, it doesn't need to be installed.


x42:
rebased & merged

PR #112, opened 2015-05-01
Fix mixed spaces/tabs in wscripts

SpotlightKid:
To prevent error when building with Python 3.


x42:
rebased & merged as 4.0-221-ga9bc95a.
Thank you.

PR #116, opened 2015-06-20
Add option to mackie control profile editor to remove binding (fixes #6301)

ovenwerks:
This fixes http://tracker.ardour.org/view.php?id=6301


pauldavisthefirst:
manually merged.

PR #118, opened 2015-07-10
Add in MXF file as a supported video format

Seablade:
No reason that I can see MXF shouldn't be a supported format, it add in the option of using XAVC as an interchange format from Lightworks to work on audio in post.


x42:
git rebased as 7d3ef2318338b4dc6f836dbdba34f56627633a6d

PR #119, opened 2015-07-12
Move learned midi binding reload after midi map load (fixes #6405)

ovenwerks:
Loading the midi map removes all bindings and so has to be done first.


ovenwerks:
There are a lot more ardour3 things floating around, but I am less sure they can be changed as easy.

x42:
The config-dir is `user_config_directory()` I think we should simply expand that in the string in the text,
otherwise we'll have the same issue when ardour5 comes out.

ovenwerks:
x42: I agree. I happened to do ardour --help on the command line and got ~/.ardour3/ If it could be a variable/function/whatever that would be better.

PR #120, opened 2015-07-20
Allow any one midi event to control only one thing.

ovenwerks:
- When learning a MIDI control, make sure the new learned control only controls one thing by deleting any old bindings with the same midi event.
 - Program change events are treated as toggle or "on", rather than toggle or PC number is value.
 - Added MasterPosition=0 to all midi map files. Going from a two unit setup with master on right to a single surface setup will not reset position to zero otherwise. As with other values in this file.
 - Encoder events are added for the MIDI map files. Will handle the 4 common types known:
  * Relative signed bit
  * Relative signed bit 2 (above with switched directions - Like MCP)
  * Relative 2s complement
  * Relative bin offset


pauldavisthefirst:
Manually merged via git.ardour.org

PR #122, opened 2015-08-04
Added PT8 session audio region import functionality

zamaudio:
The ptfformat library currently opens 97% of .ptf files and collects audio source/region/track information.


zamaudio:
Superceded, new PR soon to come.

PR #123, opened 2015-08-04
Added basic PT .ptf session audio import functionality

zamaudio:
Currently all .ptf files are supported, (PT8, PT9) audio regions/tracks only.  Layering needs work.


x42:
rebased as 4.1-520-gd686cb2 Thanks!

PR #128, opened 2015-08-12
Fix MCP master possition so LEDs work as well.

ovenwerks:
Global LEDs did not work if master position was not left most. This fixes that.


ovenwerks:
I will push this direct.

PR #129, opened 2015-08-16
Tweaking build/run scripts to facilitate recent MSYS2 as a build environment.

guysherman:
After working through building ardour under MSYS2 with Tim Mayberry, we came to the conclusion that MSYS2 is on the bleeding edge. I then went back and rebuilt most of the dependencies from source. There were still a couple of issues:
  *  Python 3 seems to put quotes around the version number, which breaks the build
  *  The dll's don't end up next to the exe, so I added a script called ardev-win, which puts all the libs folders onto the path after running ardev_common_waf.sh.

I'll be writing a blog post which describe my process for building ardour under the latest msys2.


x42:
rebased and pushed as 4f387f3  thanks!

PR #130, opened 2015-08-25
Sort entries in the keybinding editor

napcode:
This is my first pull request for ardour. Please let me know if there is anything wrong. Thanks.

### DIFF
--- a/gtk2_ardour/keyeditor.cc
+++ b/gtk2_ardour/keyeditor.cc
@@ -58,6 +58,7 @@ KeyEditor::KeyEditor ()
 	last_keyval = 0;
 
 	model = TreeStore::create(columns);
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
Thanks for your contribution. Unfortunately, it comes at a bad time. Next week I will be merging a major development branch ("tabbed") into the master branch, and this contains a major reworking of keybindings. Applying your patch will break that process. I will, however, read your patch and be sure to incorporate it after the merge if it still seems appropriate.

Thanks again, and thanks for using the correct mechanism for patches.

napcode:
Ok. I'll check the tabbed branch. My next step was adding filtering/search for the keybinding editor. Is that going to be added with "tabbed" as well? Would it make sense to develop that on top of master?

PR #131, opened 2015-08-27
Fixes a double mutex unlock for Linux VSTs

napcode:
Fixes a double mutex unlock / undefined behavior / crash for linux VSTs that could occur upon opening the VST GUI.  Line 470 unlocks the mutex correctly.


x42:
Nice find. Thanks a lot.

PR #134, opened 2015-08-31
Extensions

mojofunk:
I've had this cleanup/refactoring branch hanging around for a while relating to file extensions. It is fairly straight forward mostly just name changes, a few new utility functions and simplifications. A review or some input before I commit is appreciated.


pauldavisthefirst:
Is this PR still worth considering? Happy to be in favor of rationalization, consolidation and consistency ...

mojofunk:
I think there a few commits that are worth applying and I'd move the basename_nosuffix function into pbd/file_utils.h but I'm not sure about changing the names.

The intention was to more closely match the terminology used in what is now in C++17 http://en.cppreference.com/w/cpp/experimental/fs and use the term "extension" rather than "suffix", but we could go also go with more QT style naming like http://doc.qt.io/qt-5/qfileinfo-members.html

I'd be fine either way, but it is obviously not a priority. I'll just close this and push the straight forward refactoring/changes after the next release.

PR #136, opened 2015-09-08
Replaced broken code in recent_sessions.cc

guysherman:
The original broken code was intended to stop using ifstream
and ofstream to open files, but it used a very old MSVCism to
open streams from FILE*.

In the case for reading, this patch replaces that code by using
standard c file manipulation to read the data into a stringstream,
and then it uses the existing stream-based code to read from there.

Similarly the ofstream in the code for writing the recent files list
is replaced by a stringstream, and then a c-string is extracted
from it and written using standard c file manipulation.

Also tweaked a couple of typos from rgareus :)


x42:
rebased, separated in 2 commits and  merged

PR #137, opened 2015-09-15
Remove dependence on ardour_ui.h from many source files

mojofunk:
Here is the ardour-ui branch rebased again onto master again.

This series of commits is a branch I've had around for a while. It is a bit of refactoring so that source files don't have to include the ardour_ui.h header to access the UIConfiguration instance and for things like setting tooltips.

It results in less coupling to the ARDOUR_UI class and means faster rebuild times as ~60 less files need to be rebuilt when ardour_ui.h(or header included via ardour_ui.h) is changed.


x42:
rebased/merged

PR #138, opened 2015-09-16
Use the same default SCHED_FIFO priority as jack in ALSA backend

mojofunk:
Here is a patch for bug #6585


pauldavisthefirst:
rejecting this on the grounds that max - 20 is better default than min + 9. jack1 needs modifying to use the same approach.

PR #139, opened 2015-09-24
Fix color of mixer strip gain display.

chaot4:
Should also play nice with light themes.


pauldavisthefirst:
fixed via an alternate approach in deabc0f

PR #140, opened 2015-09-25
remove bogus check from PortManager::port_name_prefix_is_unique()

nphilipp:
The `_backend` member isn't even used in `PortManager::port_name_prefix_is_unique()`, and it breaks building here because gcc in Fedora 23 (with C++11 semantics FWIW) doesn't cast that into `bool`.

### DIFF
--- a/libs/ardour/port_manager.cc
+++ b/libs/ardour/port_manager.cc
@@ -201,10 +201,6 @@ PortManager::n_physical_inputs () const
 bool
 PortManager::port_name_prefix_is_unique (const string& first_part_of_port_name) const
 {
-	if (!_backend) {
-		return boost::shared_ptr<Port>();
-	}
-
 	boost::shared_ptr<const Ports> pr = ports.reader();
 	const string::size_type len = first_part_of_port_name.length();
 	


x42:
Which gcc version that? Which version of boost?

Ardour compiles just fine with gcc-5.2.1 (20150903) and c++11. But we've seen similar issues with clang in the past.

What is the actual error message from the compiler?

`boost::shared_ptr<AudioBackend> _backend`.  Checking if a boost::shared_ptr is valid is done in various places, even in the same file. Odd that this change resolves the issue.

PS. Unique prefix is currently work in progress.  See http://tracker.ardour.org/view.php?id=6571 and http://tracker.ardour.org/view.php?id=6522  it may not be currently used but it may return.

nphilipp:
Sorry for the late reply, I was away for a week. Here are the versions:

    gcc-5.1.1-4.fc23.x86_64
    boost-1.58.0-8.fc23.x86_64

And this is the error I get:

    ../libs/ardour/port_manager.cc: In member function ‘bool ARDOUR::PortManager::port_name_prefix_is_unique(const string&) const’:
    ../libs/ardour/port_manager.cc:205:34: error: cannot convert ‘boost::shared_ptr<ARDOUR::Port>’ to ‘bool’ in return
       return boost::shared_ptr<Port>();
                                      ^

To me, the error looks as if it doesn't like returning a `boost::shared_ptr` from a method that should return a `bool`.

pauldavisthefirst:
I am going to remove the function entirely. It was only briefly in use.

nphilipp:
Why did you close this, x42? The method isn't removed yet and the code needs a patch to build properly (at least here with my toolchain).

pauldavisthefirst:
method has been removed.

PR #141, opened 2015-09-27
Display specific error messages when failing to start AudioBackend

mojofunk:
This branch implements a facility for backends to return specific error codes in AudioBackend::_start function that will be displayed in the GUI. For instance if a sample rate or device configuration is not supported.

It also makes a change so that the Audio Setup dialog will not be hidden if the Backend fails to start so that a different configuration can be chosen.

Display a warning when closing the Audio Setup dialog when the engine is not running and a Session is loaded, that certain functions are disabled.

There is still some further work to be done to catch all possible errors in push_state_to_backend etc but this should be a good start.


pauldavisthefirst:
I'm not so comfortable with the explicit window parenting going on in some of these patches. The rest looks good.

mojofunk:
After some discussion on irc I've pushed this pull request to master. If there is an unforeseen issue with window parenting those two individual commits can be reverted.

PR #144, opened 2015-09-30
Proposed fix for bug #6615

mojofunk:
I included a commit to remove the vestigial frame from TimeAxisViewItem as it was not being used anymore.


chaot4:
Nice!

I think the frame should even disappear a little earlier. Maybe <5 pixels?

mojofunk:
Applied to ardour master as of d6943a34. The width to hide the frame can be tweaked later, this at least fixes the solid black regions when zoomed out.

PR #146, opened 2015-10-03
Click: Normalized then dropped by 6dB

zamaudio:
Applied gain to both emph click and default click using sox.
This results in louder click sound by default and user is
able to boost click to 0dBFS using the fader on the pref pane.


zamaudio:
Seems it's been committed but this PR is still active... ? Closing...

PR #148, opened 2015-10-07
Fix "Select All * Edit Point" when used as shortcut. -fixes #6626

chaot4:


x42:
rebased and merged. Thanks!

PR #150, opened 2015-10-09
Workaround Bug #6602, limited to dragging a midi event and pressing the delete key

tartina:
This is a workaround for #6602 till a better solution is found, and a proposed patch to make Editor::commit_reversible_command more robust.

### DIFF
--- a/gtk2_ardour/editor.cc
+++ b/gtk2_ardour/editor.cc
@@ -3556,9 +3556,9 @@ Editor::commit_reversible_command ()
 			cerr << "Please call begin_reversible_command() before commit_reversible_command()." << endl;
 		} else {
 			before.pop_back();
+			_session->commit_reversible_command ();
 		}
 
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
#6338 is already marked resolved. If you have continuing issues with it, please report them on the bug.

tartina:
That was a workaround for #6338, but it could be useful for similar issues. #6338 is resolved (slicing while dragging). The commit comment was about #6338, maybe it's a little bit misleading.

mojofunk:
I have pushed the part of this PR that is a fix for bug #6602 to master as e2b72419

 If both of the issues in the PR are now addressed, please close.

Thanks.

tartina:
I only have one doubt: in Editor::commit_reversible_command is it safe to call _session->commit_reversible_command() if before vector is empty?
But the 2 issues are now fixed, so I think this pr can be closed.

PR #151, opened 2015-10-09
Fix midi region navigation

chaot4:
For the first commit it has to be discussed if there are no side effects that I missed. And the second commit changes the shortcuts -- so this probably should also be discussed.

Those two commits fix #6555.

### DIFF
--- a/gtk2_ardour/editor_canvas_events.cc
+++ b/gtk2_ardour/editor_canvas_events.cc
@@ -64,10 +64,6 @@ using Gtkmm2ext::Keyboard;
 bool
 Editor::track_canvas_scroll (GdkEventScroll* ev)
 {
-	if (Keyboard::some_magic_widget_has_focus()) {
-		return false;
-	}
-
 	framepos_t xdelta;
 	int direction = ev->direction;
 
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
first part seems suspect.

second part is fine, but i'm not totally clear on the reason/purpose ...

chaot4:
From my understanding an event that should be handled by a magic widget can only be passed to track_canvas_scroll when it was let through by that magic widget. Testing this with with the different magic widgets supported my understanding. If my understanding is wrong then -- of course -- I have to rework the first commit.

The purpose of the second commit is to also enable the convenient shift+scroll navigation in the canvas when being in draw mode, a note is selected and one is hovering over the corresponding midi region. Of course it is possible to switch to grab mode, then ctrl+shift and then back to draw mode. But I don't see the purpose of not enabling horizontal scrolling via shift+scroll when ctrl+shift+scroll is still unused and the current functionality of shift+scroll can simply be transfered to that shortcut. Also regarding consistency, as ctrl+scroll is mapped to zooming in this case.

Does that clarify things?

pauldavisthefirst:
magic widgets are intended for keyboard focus only, not general events. so the first part still seems odd.

the second part makes sense. does it work for you if i commit just this change?

pauldavisthefirst:
cherry-picked as 5f7ae1294ed

PR #152, opened 2015-10-11
Disable scrolling in the dropdown menus of midi region heads.

chaot4:
Instead pass it to the canvas as proposed by x42 in #6143 (first comment).


pauldavisthefirst:
I plan to get rid of these dropdowns entirely but for now this is a reasonable temporary fix.

Manually merged.

PR #153, opened 2015-10-12
When a session or track template already exists, ask if it should be overwritten.

chaot4:
These commits fix #6587. They also introduce a dialog overwrite_file_dialog (in utils.cc) for such purposes which I then also introduce at other places in the code.


pauldavisthefirst:
manually rebased and merged.

PR #154, opened 2015-10-14
Ctrl+shift functionality (toggle all) for plugin leds.

chaot4:
The Fader led has special behavior and is only toggled when clicked
explicitly. When the Fader led is clicked with ctrl+shift then it's toggled and all the other plugin leds are set accordingly.


pauldavisthefirst:
manually rebased and merged.

PR #155, opened 2015-10-15
Fix midi menu consistency

chaot4:
That is:
* Introduce "Transform..." to the menu of MIDI regions.
* Introduce "Transpose..." to the context menu of MIDI notes.
* (Fix a tiny memory leak on the way.)


chaot4:
The last commit can also be taken out if you want to keep the unused functions. I just thought I better clean up to avoid confusion in the future.

pauldavisthefirst:
manually rebased and merged.

PR #157, opened 2015-10-22
Reduce time to load and unload a Session

mojofunk:
A series of patches focused on reducing the time to load and unload a session, more details in the commit log.

This is the second pull request for this branch, there was a problem with the changes in the last patch of the previous PR related to removing the static GhostRegion::CatchDeletion signal but it should now be correct.


pauldavisthefirst:
manually rebased and merged.

PR #158, opened 2015-10-26
Add an AppData file.

nphilipp:
This file gets used by GNOME Software to present an application to users
so they can install it from a software repository.


x42:
What do we gain from this? It's not relevant for binaries from ardour.org, and given past history we won't be able to keep the screenshots  and other meta-data up to date and translatable upstream.

Note: While gtk-ardour is GPL2, the combined license (of all lib involved in ardour) is GPL-3+ see http://metadata.ftp-master.debian.org/changelogs/main/a/ardour/ardour_4.4~dfsg-1_copyright

pauldavisthefirst:
The only GPL3 code is in the ALSA backend, which Fedora packages separately.

pauldavisthefirst:
We've traditionally been willing to keep upstream packaging meta data files in the repo. The appdata file is just another example of that. It doesn't provide any benefit for our own packages, but it is just a courtesy, in essence to linux packaging.

nphilipp:
Thanks for the heads-up regarding licensing of the ALSA backend. I'm not sure if I should the license in the appdata file then -- libltc upon which ardour depends is LGPLv3+ and who knows how this affects the effective license of Ardour in a usable form.

pauldavisthefirst:
LGPL has different consequences from GPL. At least it used to 20 years ago.

nphilipp:
Yeah, I'm just not sure if a GPLv2+ source would be implicitly "upgraded" to GPLv3 when linked to a LGPLv3(+) library (not being a lawyer and so forth).

x42:
GPL2  and  LGPL3+  => GPL3   http://www.gnu.org/licenses/gpl-faq.html#AllCompatibility

nphilipp:
This still leaves the question what should be put into the appdata file ;).

I've just chatted with Richard Hughes who wrote [the appdata spec](https://secure.freedesktop.org/~hughsient/appdata/) originally, he doesn't have a preference though. I could contact the current AppStream spec maintainer [Matthis Klumpp](https://launchpad.net/~ximion), if he has a preference -- the [appstream spec](http://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_license) doesn't indicate one way or the other either.

x42:
There is currently no way to disable LTC integration, so all ardour binaries are a combination and hence the result must be licensed as GPLv3.

The license of the individual source-code part is not affected by this.
IIUC "App data" concerns the ready-to-run application that the user is about to install (not the source-code)  and hence GPL3+ is correct.

x42:
Can <caption> and <description> in an AppData file be translated?

jonathanstowe:
According to http://people.freedesktop.org/~hughsient/appdata/

```
If you ship an .xml.in file rather than an .xml file, you can use intltool to translate the data. You can also use itstool to translate the data, and we've provided a appdata.its file for convenience.
```

nphilipp:
BTW, I checked with those who care about this AppData/AppStream thing (Richard Hughes (@hughsie) and Matthias Klumpp (@ximion). According to them the metadata should track the project source license, not what combining it with all its dependencies results in. That makes things a lot less work-prone ;).

ximion:
I've added a few remarks on the metadata file and translation process.
As I see, the issues are already being fixed :-) Hope you find this helpful.
In the end, you may want to validate the generated metadata file using `appstreamcli validate <file>.xml` or `appstream-util validate <file>`.
The appstreamcli utility also has a nice feature to validate the whole combination of .desktop file and MetaInfo file, just install your app into $DESTDIR and then run `appstreamcli validate-tree $DESTDIR` (that's used in the KDE CI to ensure data doesn't break). Both tools are available on Debian, Ubuntu and Fedora (appstream / appstream-util package).

nphilipp:
I've incorporated most changes suggested by @ximion, except adding a Unix timestamp to the release tag which would be laborious to get right if building from somewhere else than a git repo (e.g. a tarball).

ximion:
:+1:

PR #161, opened 2015-11-04
Show info dialog when opening chat via Help -> Chat.

chaot4:
This should clarify for users that in IRC you post your question and
wait, and not leave after a few minutes.

NOTE: Please review the text that is shown in the dialog and correct it when necessary (which is probably the case).


pauldavisthefirst:
whitespace issues. i can fix them, but you might want to consider the source of them ...

chaot4:
Would be kind if you can fix them. Thanks.

PR #164, opened 2015-11-10
Fix a memory leak in Stateful class and some formatting fixes

mojofunk:
I can push this if it looks good


pauldavisthefirst:
It looks good, please push.

PR #165, opened 2015-11-12
Add support for portaudio callback API but keep blocking as default

mojofunk:
Enabling the callback API is a compile time option but keep blocking API as the default until further and wider testing.

I can push this when it is an appropriate time for this change.


mojofunk:
This commit has already been pushed to ardour master

PR #166, opened 2015-11-12
Some refactoring and documentation in pbd/demangle

mojofunk:
I can push this if it looks good.


mojofunk:
Applied in ardour master as of 7f3118c8

PR #167, opened 2015-11-12
Fix broken load-session utility and change it to load and save a session

mojofunk:
Rename run-session-tests.sh script to load-save-session.sh and make it operate on
only a single session and add a separate load-save-session-collection.sh script

Add --massif option to load-save-session.sh script

Rename some poorly named variables and add some documentation to the
load-save-session.sh script

Not sure how long this utility has been broken but fixing it and modifying it to also save the session was functionality I needed and is useful for finding changes to the session file as a result of the save(it found bug 6670).

I can push this if it looks good.


mojofunk:
Applied in ardour master as 44f46d2b

PR #168, opened 2015-11-12
Fix bug #6506, crash when moving tempo markers

mojofunk:
Seems to fix it. Not totally familiar with all the code paths so a review would be appreciated.


mojofunk:
This has been committed to master as 9694dfc8d2

PR #169, opened 2015-11-13
Add support to signal the transport state by OSC

johannes-mueller:
Implement transport_speed in the same style like transport_frame. Up to
now transport_speed and record_enabled are implemented.

punch_in, punch_out and transport_locked are missing. Actually I don't know how to test transport_locked. I'd have to investigate what it exactly is first.

Please take a look to see whether you'd pull this. If yes would you want me to implement the missing queries first?

### DIFF
--- a/libs/surfaces/osc/osc.cc
+++ b/libs/surfaces/osc/osc.cc
@@ -336,7 +336,10 @@ OSC::register_callbacks()
 		REGISTER_CALLBACK (serv, "/ardour/transport_stop", "", transport_stop);
 		REGISTER_CALLBACK (serv, "/ardour/transport_play", "", transport_play);
 		REGISTER_CALLBACK (serv, "/ardour/transport_frame", "", transport_frame);
+		REGISTER_CALLBACK (serv, "/ardour/transport_speed", "", transport_speed);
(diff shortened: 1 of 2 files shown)


x42:
the callbacks need an
```
if (!session) {
    return;
}
```
clause.  Otherwise this looks good.

libs/surfaces/osc/osc.cc
```diff
@@ -788,6 +790,33 @@ OSC::transport_frame (lo_message msg)
 	lo_message_free (reply);
 }
 
+void
+OSC::transport_speed (lo_message msg)
+{
```
x42:
```
if (!session) {
    return;
}

File: libs/surfaces/osc/osc.cc [hunk 2]
```diff
@@ -788,6 +790,33 @@ OSC::transport_frame (lo_message msg)
...
+	lo_message_add_double (reply, ts);
+
+	lo_send_message (lo_message_get_source (msg), "/ardour/transport_speed", reply);
+
+	lo_message_free (reply);
+}
+
+void
+OSC::record_enabled (lo_message msg)
+{
```
x42: ditto.  session may be NULL

x42:
rebased, merged and session-guards added in  4.4-398 ba02c17

Thanks for your contribution.

PR #170, opened 2015-11-14
Added support to set gain automation by OSC.

johannes-mueller:
Added support for OSC message

"/ardour/routes/gain_automation" rid state

where rid is the remote route id and state is a character representing
the desired gain automation state:

'm': Manual (Off)
'p': Play
'w': Write
't': Touch

### DIFF
--- a/libs/surfaces/osc/osc.cc
+++ b/libs/surfaces/osc/osc.cc
@@ -362,6 +362,8 @@ OSC::register_callbacks()
 		REGISTER_CALLBACK (serv, "/ardour/routes/send/gainabs", "iif", route_set_send_gain_abs);
 		REGISTER_CALLBACK (serv, "/ardour/routes/send/gaindB", "iif", route_set_send_gain_dB);
 
+		REGISTER_CALLBACK (serv, "/ardour/routes/gain_automation", "is", route_set_gain_automation);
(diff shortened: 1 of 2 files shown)


x42:
ideally there'd be an API to control all automatables - not just fader-gain.

johannes-mueller:
I will look into this after new year.

x42:
having a 2nd look.. the faders are special-cased in routes anyway.

the only common part with generic plugin controls would be the switch {} statement
for `boost::shared_ptr<AutomationControl>`

what do you think of

```
static void set_automation_state (boost::shared_ptr<AutomationControl> c, char state)
{
switch (state) ....
}

OSC::route_set_gain_automation (int rid, char state)
{
  if (!session) { return -1;}
  boost::shared_ptr<Route> r = session->route_by_remote_id (rid);
  if (!r || !r->gain_control ()) { return -1; }
  set_automation_state (r->gain_control (), state);
}
```

johannes-mueller:
Sounds sensible to me.

Unfortunately I won't have time to hack on this before January. I will leave the PR open though. Maybe other people come up with some more clever ideas.

johannes-mueller:
New PR with same functionality #225 ...

PR #172, opened 2015-11-23
make copies of global lists before extending

nphilipp:
Not doing so can make source or object files appear multiple times in
the list of files to be compiled or linked, e.g. when doing './waf build
install', subsequently leading to linker errors.


x42:
was merged/rebased as 4.4-277  e25ddc3

PR #174, opened 2015-11-25
Add more menu entries to the 'Help' menu.

chaot4:
* The functions were already implemented -- I just added the links in
ardour.menu.in, corrected spelling and the link to the tracker.
* The tracker link was corrected as I considered it confusing when one
is not already logged in.

### DIFF
--- a/gtk2_ardour/ardour.menus.in
+++ b/gtk2_ardour/ardour.menus.in
@@ -544,6 +544,11 @@
       <menuitem action='Chat'/>
       <menuitem action='Manual'/>
       <menuitem action='Reference'/>
+      <menuitem action='Forums'/>
+      <menuitem action='Howto_Report'/>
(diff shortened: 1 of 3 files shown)


pauldavisthefirst:
2 requested changes:

* use S_("Context|TranslationTerm") to disambiguate Manual.
* use foo-bar rather than Foo_Bar for action names (I know we've been very inconsistent in the past, but we're trying to be better).

chaot4:
You meant "to disambiguate Reference" (that's what I did in the commit)? Because "Manual" is already diambiguated. I also fixed the action names.

pauldavisthefirst:
Ah, OK, now I see the point of your comment: *extra* clarity for translators over and above what S_(A|B) provides. thanks.

pauldavisthefirst:
manually merged via git.ardour.org

PR #175, opened 2015-11-26
Fix "Crop Region to Range".

chaot4:
Before only regions that contained the start of the range were cropped,
now all are cropped, i.e. also those that begin inside the range and exceed its end.


pauldavisthefirst:
manually merged.

PR #176, opened 2015-11-26
Fix fill track

chaot4:
It then behaves like Multi-Duplicate until the end marker. The end
marker is excluded as this would often lead to size 1 regions. For convenience I added a helper function to Playlist.


pauldavisthefirst:
manually merged.

PR #177, opened 2015-11-29
No longer shorten note by one tick

ngeiswei:
# Overview

This PR contains
1. removes shorting by one tick notes inserted in the midi region
2. minor indentation fix in the NodeBase class

# Motivation

I don't think shortening the notes in the sequencer is right, for the following reasons
1. It is inconsistent with other behaviors such as resizing a note, and naming a note length in the midi list editor. See the image attached, 
ardour
the first 3 notes and the last ones have been inserted before the fix. The 2 in the middle have been inserted after the fix, and have their lengths correctly named.
2. Even though the note off and note on can have same timestamps, I believe Ardour can still strive to send them is the right order (I didn't check what it's doing though). Then assuming they are sent in the right order, it's up to the client (the synth) to implement the correct behavior.
3. If that really remains an issue then I suggest to add an option to silently shorten notes off preceding the same notes on. In any case,  we do want to add eventually on-the-fly control over note lengths, this can be useful to shorten or lengthen notes (convenient for some portamento effect for instance). But I don't think it should be represented explicitly in the note off bbt. What if for instance the note is moved and no longer overlaps with the same following note on? So this should be more considered as sanity post-processing.

### DIFF
--- a/gtk2_ardour/midi_region_view.cc
+++ b/gtk2_ardour/midi_region_view.cc
@@ -1702,9 +1702,9 @@ MidiRegionView::update_sustained (Note* ev, bool update_ghost_regions)
 
 	if (note->length() > 0) {
 		const framepos_t note_end_frames = min (source_beats_to_region_frames (note->end_time()), _region->length());
-		ev->set_x1 (trackview.editor().sample_to_pixel (note_end_frames));
+		ev->set_x1 (std::max(1., trackview.editor().sample_to_pixel (note_end_frames)) - 1);
 	} else {
-		ev->set_x1 (trackview.editor().sample_to_pixel (_region->length()));
+		ev->set_x1 (std::max(1., trackview.editor().sample_to_pixel (_region->length())) - 1);
 	}
 
 	ev->set_y1 (y0 + std::max(1., floor(midi_stream_view()->note_height()) - 1));
--- a/gtk2_ardour/midi_region_view.cc
+++ b/gtk2_ardour/midi_region_view.cc
@@ -554,12 +554,6 @@ MidiRegionView::button_release (GdkEventButton* ev)
 					group->canvas_to_item (event_x, event_y);
 
 					Evoral::Beats beats = get_grid_beats(editor.pixel_to_sample(event_x));
-
-					/* Shorten the length by 1 tick so that we can add a new note at the next
-					   grid snap without it overlapping this one.
-					*/
-					beats -= Evoral::Beats::tick();
-
 					create_note_at (editor.pixel_to_sample (event_x), event_y, beats, true);
 				} else {
 					clear_selection ();
@@ -570,14 +564,7 @@ MidiRegionView::button_release (GdkEventButton* ev)
 		case MouseDraw:
 			{
 				Evoral::Beats beats = get_grid_beats(editor.pixel_to_sample(event_x));
-
-				/* Shorten the length by 1 tick so that we can add a new note at the next
-				   grid snap without it overlapping this one.
-				*/
-				beats -= Evoral::Beats::tick();
-
 				create_note_at (editor.pixel_to_sample (event_x), event_y, beats, true);
-
 				break;
 			}
 		default:
--- a/gtk2_ardour/note.h
+++ b/gtk2_ardour/note.h
@@ -51,8 +51,8 @@ public:
 	void set_x1 (ArdourCanvas::Coord);
 	void set_y1 (ArdourCanvas::Coord);
 
-        void set_outline_what (ArdourCanvas::Rectangle::What);
-        void set_outline_all ();
+	void set_outline_what (ArdourCanvas::Rectangle::What);
+	void set_outline_all ();
 
 	void set_outline_color (uint32_t);
 	void set_fill_color (uint32_t);
--- a/gtk2_ardour/note_base.cc
+++ b/gtk2_ardour/note_base.cc
@@ -238,14 +238,14 @@ NoteBase::set_mouse_fractions (GdkEvent* ev)
 	_mouse_y_fraction = yf;
 
 	if (notify) {
-                if (big_enough_to_trim()) {
-                        _region.note_mouse_position (_mouse_x_fraction, _mouse_y_fraction, set_cursor);
-                } else {
-                        /* pretend the mouse is in the middle, because this is not big enough
-                           to trim right now.
-                        */
-                        _region.note_mouse_position (0.5, 0.5, set_cursor);
-                }
+		if (big_enough_to_trim()) {
+			_region.note_mouse_position (_mouse_x_fraction, _mouse_y_fraction, set_cursor);
+		} else {
+			/* pretend the mouse is in the middle, because this is not big enough
+			   to trim right now.
+			*/
+			_region.note_mouse_position (0.5, 0.5, set_cursor);
+		}
 	}
 }
 
--- a/gtk2_ardour/note_base.h
+++ b/gtk2_ardour/note_base.h
@@ -59,7 +59,7 @@ class NoteBase : public sigc::trackable
 	virtual ~NoteBase ();
 
 	void set_item (ArdourCanvas::Item *);
-        ArdourCanvas::Item* item() const { return _item; }
+	ArdourCanvas::Item* item() const { return _item; }
 
 	virtual void show() = 0;
 	virtual void hide() = 0;
@@ -94,8 +94,8 @@ class NoteBase : public sigc::trackable
 	virtual ArdourCanvas::Coord x1 () const = 0;
 	virtual ArdourCanvas::Coord y1 () const = 0;
 
-        float mouse_x_fraction() const { return _mouse_x_fraction; }
-        float mouse_y_fraction() const { return _mouse_y_fraction; }
+	float mouse_x_fraction() const { return _mouse_x_fraction; }
+	float mouse_y_fraction() const { return _mouse_y_fraction; }
 
 	const boost::shared_ptr<NoteType> note() const { return _note; }
 	MidiRegionView& region_view() const { return _region; }
@@ -128,8 +128,8 @@ class NoteBase : public sigc::trackable
 	/// hue circle divided into 16 equal-looking parts, courtesy Thorsten Wilms
 	static const uint32_t midi_channel_colors[16];
 
-        bool mouse_near_ends () const;
-        virtual bool big_enough_to_trim () const;
+	bool mouse_near_ends () const;
+	virtual bool big_enough_to_trim () const;
 
 protected:
 	enum State { None, Pressed, Dragging };
@@ -143,10 +143,10 @@ protected:
 	bool                              _own_note;
 	bool                              _selected;
 	bool                              _valid;
-        float                             _mouse_x_fraction;
-        float                             _mouse_y_fraction;
+	float                             _mouse_x_fraction;
+	float                             _mouse_y_fraction;
 
-        void set_mouse_fractions (GdkEvent*);
+	void set_mouse_fractions (GdkEvent*);
 
 private:
 	bool event_handler (GdkEvent *);
--- a/libs/ardour/enums.cc
+++ b/libs/ardour/enums.cc
@@ -308,12 +308,12 @@ setup_enum_writer ()
 	REGISTER (_LayerModel);
 
 	REGISTER_ENUM (InsertMergeReject);
-        REGISTER_ENUM (InsertMergeRelax);
-        REGISTER_ENUM (InsertMergeReplace);
-        REGISTER_ENUM (InsertMergeTruncateExisting);
-        REGISTER_ENUM (InsertMergeTruncateAddition);
-        REGISTER_ENUM (InsertMergeExtend);
-        REGISTER (_InsertMergePolicy);
+	REGISTER_ENUM (InsertMergeRelax);
+	REGISTER_ENUM (InsertMergeReplace);
+	REGISTER_ENUM (InsertMergeTruncateExisting);
+	REGISTER_ENUM (InsertMergeTruncateAddition);
+	REGISTER_ENUM (InsertMergeExtend);
+	REGISTER (_InsertMergePolicy);
 
 	REGISTER_ENUM (AfterFaderListen);
 	REGISTER_ENUM (PreFaderListen);
--- a/libs/ardour/midi_model.cc
+++ b/libs/ardour/midi_model.cc
@@ -1665,9 +1665,9 @@ MidiModel::resolve_overlaps_unlocked (const NotePtr note, void* arg)
 
 		if ((sb > sa) && (eb <= ea)) {
 			overlap = OverlapInternal;
-		} else if ((eb >= sa) && (eb <= ea)) {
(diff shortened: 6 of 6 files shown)


x42:
If you have the patch applied, can you check if you can still enter continuous notes with "Session > Properties > Misc > never allow overlapping notes" enabled ?

There are a couple of off-by-ones regarding sample-alignment in Ardour, and this tick-subtraction acts as workaround until those are fixed.

Play the midi back and print the result with `jack_midi_dump -r`.  For all cases (except loop-wrap) the note-on/off spacing is exact (thanks to another off-by-one somewhere).  I just tested this with step-entry.

ngeiswei:
@x42 I cannot enter contiguous notes after enabling "Session > Properties > Misc > never allow overlapping notes" indeed. I'm wondering how other DAWs handle that problem...

I tried to use jack_midi_dump -r, but it doesn't output anything. I saw in QJackCtrl that a midi input port "-r" was created, I connected Ardour to it, but still no output. I must be using things wrongly, studying...

ngeiswei:
I cannot find documentation about jack_midi_dump anyway. No manpage, no --help, no wiki, nothing, not even in the jackd1 source code... Keep searching...

ngeiswei:
Here, they say something about Cubase 5.0

http://www.soundonsound.com/sos/mar01/articles/cubasenotes.asp

"Cubase 5.0 normally avoids this possibility by forcing a short gap between Note Off and Note On commands on the same MIDI channel".

It's kinda up to interpretation but the way I interpret it is that you'd allow a note off to have the same time as the next note on, as that is what the user expects, but at play time you'd send the note off slightly earlier, of an amount determined by a user option.

ngeiswei:
I tried a few DAWs, well I could only get the precise start-end time info in EnergyTX, and it confirms what I think in terms of what the user expects. The snapped note off has the same time as the next note on (as my PR corrects).

So I guess to move forward with that PR I would need to update the behavior of "Session > Properties > Misc > never allow overlapping notes" (and probably others) under the assumption that a note off at the same time of a note on is not considered as overlapping.

ngeiswei:
On top of it, to support well for old synthesizers, I guess I should add an option
```
Session > Properties > Misc > shorten note off of N-ticks
```
But I'd rather do that in an other PR and after actually encountering the problem on some hardware (unless someone has experienced such a problem already can test if such subsequent patch would work).

ngeiswei:
@x42 I've fixed the overlapping resolution. Let me know if there are other potential related issues I should take care of.

x42:
thanks, I'll have a look. 

quick re about jack_midi_dump.. (midi-monitor):  Is there a synth on the midi track? If so that'll  eat the midi-events and output audio.  Create a Midi-track w/o synth or remove the synth.

x42:
I'd like a second opinion about this from @drobilla or @pauldavisthefirst  regarding the consequences of changing the overlap definition in the midi_model.

`resolve_overlaps_unlocked` is only used by `add_note_unlocked`.  but the latter is used in various places.

one example:  `Sequence<Time>::append` calls it.  The documentation for this in Sequence.cpp:802 reads "timestamp MUST be >= the latest event".  The change you propose to ">" would no longer trigger `OverlapStart`  which may or may not be an issue when reading .mid files.

ngeiswei:
I'm not seeing the whole picture, so not completely sure, but after reviewing a bit `Sequence<Time>::append` I don't think it would affect it, the constraint you mention seems to correspond to

```
	assert(_notes.empty() || ev.time() >= (*_notes.rbegin())->time());
```

which seems  unrelated to the definition of `OverlapStart`.

ngeiswei:
I've added one more commit, to better display the end note. It should display the note border at end_time-1 since end time correspond to a note off, only end_time-1 still corresponds to a note on. Well, I tried that first but it didn't work, so instead I'm displaying pixel(end_time)-1 and I like the result.

See before the fix
before_note_length_fix

After the fix 
after_note_length_fix

7890:
@ngelswel, if you're in the mood to hunt off-by-one things this could be of interest:
http://trellis.ch/lowres/a3/ardour_position_and_length.pdf
cheers

ngeiswei:
I'll definitely look into this, as well as the jack_midi_dump for that issue.

pauldavisthefirst:
In spite of the apparently thoroughly written document written by tom (7890) i would strongly advise you to ignore most of its contents. A substantial amount of time was wasted on IRC because of his failure to comprehend a few basic concepts. 

There ARE issues with off-by-one throughout Ardour, both for audio and for MIDI, but that document is a chimera. Reading it once is probably worthwhile just to get a sense of the fact that it isn't trivial to define some things, but reading it more than once or paying attention to details just leads down a rabbit hole of misdirection.

ngeiswei:
Thanks @pauldavisthefirst !

7890:
@pauldavisthefirst, the reason this was put to a document is to prevent wobbly IRC discussions, where things are volatile. If you find anything wrong in that document, it's very easy to pont it out and i'd like you to point it out. It doesn't surprise me much that you advise to ignore these issues - nothing has changed in that respect, which i see as part of the problem.

pauldavisthefirst:
I have not advised anyone to ignore these issues. We wasted enough time on IRC with your complete failure to grasp some basic ideas. My advice was that off-by-one errors exist and need to be fixed, but that your document reflects your (frankly) surprising inability to understand the nature of the problems and where the solutions will be found.

7890:
I'll ignore your inability to get concrete until you do :)

ngeiswei:
I have run jack_midi_dump on the same midi region as the one showed in the screenshot, and here is the output:

```
0: 90 3f 40 note on (channel 0): pitch 63, velocity 64
77: 80 3f 40 note off (channel 0): pitch 63, velocity 64
77: 90 3f 40 note on (channel 0): pitch 63, velocity 64
154: 80 3f 40 note off (channel 0): pitch 63, velocity 64
154: 90 3f 40 note on (channel 0): pitch 63, velocity 64
230: 80 3f 40 note off (channel 0): pitch 63, velocity 64
230: 90 3f 40 note on (channel 0): pitch 63, velocity 64
307: 80 3f 40 note off (channel 0): pitch 63, velocity 64
307: 90 40 40 note on (channel 0): pitch 64, velocity 64
307: 90 3d 40 note on (channel 0): pitch 61, velocity 64
384: 80 40 40 note off (channel 0): pitch 64, velocity 64
384: 80 3d 40 note off (channel 0): pitch 61, velocity 64
384: 90 3d 40 note on (channel 0): pitch 61, velocity 64
384: 90 42 40 note on (channel 0): pitch 66, velocity 64
460: 80 3d 40 note off (channel 0): pitch 61, velocity 64
460: 80 42 40 note off (channel 0): pitch 66, velocity 64
460: 90 3d 40 note on (channel 0): pitch 61, velocity 64
460: 90 41 40 note on (channel 0): pitch 65, velocity 64
25: 80 3d 40 note off (channel 0): pitch 61, velocity 64
25: 80 41 40 note off (channel 0): pitch 65, velocity 64
25: 90 3d 40 note on (channel 0): pitch 61, velocity 64
25: 90 40 40 note on (channel 0): pitch 64, velocity 64
102: 80 3d 40 note off (channel 0): pitch 61, velocity 64
102: 80 40 40 note off (channel 0): pitch 64, velocity 64
256: 90 43 40 note on (channel 0): pitch 67, velocity 64
256: 90 40 40 note on (channel 0): pitch 64, velocity 64
256: 90 3c 40 note on (channel 0): pitch 60, velocity 64
256: 90 3d 40 note on (channel 0): pitch 61, velocity 64
135: 80 3c 0 note off (channel 0): pitch 60, velocity 0
135: 80 3d 0 note off (channel 0): pitch 61, velocity 0
135: 80 40 0 note off (channel 0): pitch 64, velocity 0
135: 80 43 0 note off (channel 0): pitch 67, velocity 0
0: b0 40 0 control change (channel 0): controller 64, value 0
0: b0 7b 0 control change (channel 0): controller 123, value 0
0: b1 40 0 control change (channel 1): controller 64, value 0
0: b1 7b 0 control change (channel 1): controller 123, value 0
0: b2 40 0 control change (channel 2): controller 64, value 0
0: b2 7b 0 control change (channel 2): controller 123, value 0
0: b3 40 0 control change (channel 3): controller 64, value 0
0: b3 7b 0 control change (channel 3): controller 123, value 0
0: b4 40 0 control change (channel 4): controller 64, value 0
0: b4 7b 0 control change (channel 4): controller 123, value 0
0: b5 40 0 control change (channel 5): controller 64, value 0
0: b5 7b 0 control change (channel 5): controller 123, value 0
0: b6 40 0 control change (channel 6): controller 64, value 0
0: b6 7b 0 control change (channel 6): controller 123, value 0
0: b7 40 0 control change (channel 7): controller 64, value 0
0: b7 7b 0 control change (channel 7): controller 123, value 0
0: b8 40 0 control change (channel 8): controller 64, value 0
0: b8 7b 0 control change (channel 8): controller 123, value 0
0: b9 40 0 control change (channel 9): controller 64, value 0
0: b9 7b 0 control change (channel 9): controller 123, value 0
0: ba 40 0 control change (channel 10): controller 64, value 0
0: ba 7b 0 control change (channel 10): controller 123, value 0
0: bb 40 0 control change (channel 11): controller 64, value 0
0: bb 7b 0 control change (channel 11): controller 123, value 0
0: bc 40 0 control change (channel 12): controller 64, value 0
0: bc 7b 0 control change (channel 12): controller 123, value 0
0: bd 40 0 control change (channel 13): controller 64, value 0
0: bd 7b 0 control change (channel 13): controller 123, value 0
0: be 40 0 control change (channel 14): controller 64, value 0
0: be 7b 0 control change (channel 14): controller 123, value 0
0: bf 40 0 control change (channel 15): controller 64, value 0
0: bf 7b 0 control change (channel 15): controller 123, value 0
```

I suppose it does show that everything is sent in the right order. Obviously I can check while playing it with an instrument that it does play normally.

Let me know if I can do anything further to help.

x42:
looks like you missed the `-r` option (ie relative timestamps) to previous event.

at 120 bpm, 48K  a step-entry sequence of whole notes with current master gives nicely spaced events at exactly 24000 samples:

```
....
    +0: be 40 00 control change (channel 14): controller  64, value   0
    +0: be 7b 00 control change (channel 14): controller 123, value   0
    +0: bf 40 00 control change (channel 15): controller  64, value   0
    +0: bf 7b 00 control change (channel 15): controller 123, value   0
+59840: 90 3c 50 note on  (channel  0): pitch  60, velocity  80
+24000: 80 3c 40 note off (channel  0): pitch  60, velocity  64
    +0: 90 3e 50 note on  (channel  0): pitch  62, velocity  80
+24000: 80 3e 40 note off (channel  0): pitch  62, velocity  64
    +0: 90 40 50 note on  (channel  0): pitch  64, velocity  80
+24000: 80 40 40 note off (channel  0): pitch  64, velocity  64
    +0: 90 41 50 note on  (channel  0): pitch  65, velocity  80
+24000: 80 41 00 note off (channel  0): pitch  65, velocity   0
+38721: b0 40 00 control change (channel  0): controller  64, value   0
    +0: b0 7b 00 control change (channel  0): controller 123, value   0
    +0: b1 40 00 control change (channel  1): controller  64, value   0
    +0: b1 7b 00 control change (channel  1): controller 123, value   0
....
```

x42:
In short: while there's an off-by-one in the GUI, the engine currently does the right thing.

ngeiswei:
my version of `jack_midi_dump` doesn't seem to register the `-r` option, it just renames the `midi-monitor` port by `-r`. I'm using jack 1, that might be the reason, I'm gonna try to make it do the right thing.

@x42 so you've tried without the patch and you get the right behavior, hmm, I guess the engine is somehow adding this extra tick somewhere down the line.

So first step I've got to make this -r option work to see how it behaves with the patch.

ngeiswei:
@x42 That is not what I observe

On the master branch I get
```
+326144: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+23988: 80 3c 40 note off (channel  0): pitch  60, velocity  64
   +12: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+23988: 80 3c 40 note off (channel  0): pitch  60, velocity  64
```

On the branch associated with that PR I get
```
+433152: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+24000: 80 3c 40 note off (channel  0): pitch  60, velocity  64
    +0: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+24000: 80 3c 40 note off (channel  0): pitch  60, velocity  64
```

Note that you cannot reproduce that off-by-one bug if you have obtained the whole notes by resizing them, only be inserting with the chosen length unit.

x42:
The first is normal for two identical notes.  There is an intentional gap.  play "C", "D" .. and the gap of 12 samples will vanish.

ngeiswei:
Weird, I'm getting the +12 over 2 different notes as well, even starting with a new project from scratch.

```
+231936: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+23988: 80 3c 40 note off (channel  0): pitch  60, velocity  64
   +12: 90 3e 40 note on  (channel  0): pitch  62, velocity  64
+23988: 80 3e 40 note off (channel  0): pitch  62, velocity  64
+35468: b0 40 00 control change (channel  0): controller  64, value   0
```

I'm gonna recompile my branch and test 2 consecutive different notes.

x42:
Mmh I don't. (Ardour 4.4-513 6554f0f)

midi_dump

One oddity is if the region-end coincides with the note-off (the note is shortened by 1 sample), but i just extended the region in this case. the "Panic" which is sent when looping is also in step sync.

ngeiswei:
2 consecutive different notes on this PR's branch gives

```
+600064: 90 3c 40 note on  (channel  0): pitch  60, velocity  64
+24000: 80 3c 40 note off (channel  0): pitch  60, velocity  64
    +0: 90 3e 40 note on  (channel  0): pitch  62, velocity  64
+24000: 80 3e 40 note off (channel  0): pitch  62, velocity  64
```

which seems normal, however I confirm that it does not add the +12 when re-triggering the same note.

I'll try again with the master (which is https://github.com/Ardour/ardour/commit/6554f0f as well on my system)...

ngeiswei:
I confirm, I'm getting +12 whether or not the notes are the same... Although I've noticed something odd, on Ardour's About, the build is indeed https://github.com/Ardour/ardour/commit/6554f0f but the version is still 4.4.344.

ardour_2_different_notes_6554f0f


Any idea? For what is worse, this build from the same project folder, just switching the branches using git. Maybe I should try from a fresh folder.

x42:
Did you draw the notes using the mouse?

I used step entry, and the list editor also shows them all as with Length: "whole".

ngeiswei:
Hey @x42 I didn't know about the step entry editor, it's awesome!

So yes, the problem occurs only when drawing the notes with the mouse (inserting, not resizing), not with the step editor.

I think this PR is ready, but let me know if you need more information or testing.

x42:
Rebased as Ardour 4.4-592- 93ea63f
Thanks a lot for your time and effort

PR #178, opened 2015-11-29
Make capitalization consistent and other minor spelling corrections.

chaot4:
It doesn't make everything perfectly consistent but it should at least be quite a bit more consistent than before. ;)


x42:
rebased as 7a846b7

PR #179, opened 2015-11-30
Hide MIDI ghost note when over existing note.

chaot4:
Currently a ghost note is always drawn -- independent from if the mouse cursor is over a note or not. This commit changes this behavior such that a ghost note is _not_ drawn, when the cursor is over a note. Then only the note information of the not under the cursor is displayed.

This PR fixes 6625.


chaot4:
I dug further into ghost note drawing and found other issues -- meanwhile I also changes this code. So I'll be submitting a larger PR soon also solving this issue.

PR #180, opened 2015-12-02
Fix several MIDI ghost note issues

chaot4:
See the commit messages for the list of fixes issues. Unfortunately one minor issue remains to be fixed but I commented it accordingly.


x42:
rebased as 4.6-7-gef9713d.

PR #182, opened 2015-12-03
Fix "Crop Region to Range" -- second attempt.

chaot4:
Now it should work in all cases. Also made the variable names a little clearer.


chaot4:
Rebased and merged by paul.

PR #183, opened 2015-12-03
Make the duplicate routes dialog look a little nicer.

chaot4:
More space between the count_spinner and its label; less space around the count_box.

Yes, this is totally subjective. Feel free to simply close this PR withour comment if you disagree.


x42:
rebased as 4.4-424 fd5dfd6

PR #187, opened 2015-12-07
NetBSD: Stop hardcoding Python binary name

krytarowski:
Caught with pkgsrc on NetBSD


x42:
rebased as Ardour 4.4-448 2da6e45

PR #188, opened 2015-12-09
Remove two useless separator elements from editor ruler popups.

chaot4:
See the commit message. Nothing more to say really.


x42:
rebased as 4.4-483 d0cdf96

PR #192, opened 2015-12-12
define __ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES=0 to avoid

david0:
This fixes a name collision with sigc++ 2.6.4 on OS X.
See also https://bugzilla.gnome.org/show_bug.cgi?id=759315

### DIFF
--- a/libs/appleutility/CoreAudio105/CABufferList.h
+++ b/libs/appleutility/CoreAudio105/CABufferList.h
@@ -83,7 +83,7 @@ protected:
 		mName(name),
 		mBufferMemory(NULL)
 	{
-		check(numBuffers > 0 /*&& channelsPerBuffer > 0*/);
+		__Check(numBuffers > 0 /*&& channelsPerBuffer > 0*/);
 		mNumberBuffers = numBuffers;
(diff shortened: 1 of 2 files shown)


x42:
__ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES is not available in the MacOSX10.5.sdk. 
We still provide PPC builds and 32bit Carbon-GUI Plugin support, even 12 years after its EOL, some AU used in Studios around the world are still Carbon GUIs, e.g  Miroslav Philharmonik...

`#undef check` is the preferred stopgap solution for the time being.

david0:
I thought 10.5 is not relevant as http://ardour.org/requirements.html says 
> Operating System	OS X 10.6 (Snow Leopard) to OS X 10.11 (El Capitan).

If 10.5 is still a target, than this PR can be closed because its not a acceptable solution.

david0:
Closing since this does not work for 10.5. Situation will be improved since libsigc++ renamed the `check`-function.

x42:
That's great news! Kudos to the sigc++ devs.

PR #194, opened 2016-01-07
Changes/Fixes to mouse scroll wheel behaviour in the Editor

mojofunk:
It would be good to get some feedback on these changes as there are some small changes in behaviour and perhaps reasons why these changes may not be appropriate but mostly it is is just some refactoring and fixes.

I'm not sure about adding an option for always using mouse position as zoom focus when scrolling but as Mixbus and Ardour did have different behaviours some people may want to always use the zoom focus setting when scrolling with the mouse wheel.


mojofunk:
This branch is now merged into Ardour:master

PR #195, opened 2016-01-10
Created updated mapping for M-Audio Axiom Air 25

guysherman:
The existing mapping for M-Audio Axiom 25 used a different set of CC values for transport controls. It also mapped the rotary encoders, which IMHO is less convenient now that MIDI Learn works in conjunction with mappings.

There is one slight issue with the OOB experience of this keyboard, but the mapping file contains instructions on how to slightly reconfigure the keyboard to work really nicely.


pauldavisthefirst:
both commits cherry-picked into master. thanks!

PR #196, opened 2016-01-12
Fix man page encoding

nphilipp:
French and Russian man pages were still encoded in ISO-8859-1 and KOI8-R, respectively, instead of UTF-8.


pauldavisthefirst:
Not 100% sure about these changes. Need to consult/read around more before pulling.

nphilipp:
I think it's good to only use UTF-8 encoding, converting to other encodings is straight-forward and doesn't involve guesswork (like if the Russian man page is in ISO-8859-5 or KOI8-R).

nphilipp:
@prokoudine, was there a special reason to originally use KOI8-R encoding for the Russian version of the manpage, other than that it was prevalent at the time?

prokoudine:
Prevalent at the time.

pauldavisthefirst:
manually rebased and merged to master. thanks!

PR #197, opened 2016-01-13
Change the tap tempo estimator to least-squares regression

potocpav:
The previous estimator had a very bad performance. The bpm value did not converge with a sufficient number of taps*. This change is to use a well-known unbiased estimator, least-squares linear regression. It is now possible to tap tempo correctly to a fraction of a bpm.

*) the error limit is equal to 58% of the error af a single tap, assuming normally i.i.d. errors.

### DIFF
--- a/gtk2_ardour/tempo_dialog.cc
+++ b/gtk2_ardour/tempo_dialog.cc
@@ -267,39 +267,38 @@ TempoDialog::pulse_change ()
 bool
 TempoDialog::tap_tempo_button_press (GdkEventButton *ev)
 {
-	gint64 now;
-	now = g_get_monotonic_time (); // microseconds
+	double t;
 
+	// Linear least-squares regression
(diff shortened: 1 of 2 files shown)


potocpav:
Fixed a potential precision issue: time is 64-bit, double mantiss is 52-bit. 12 bits are lost when converting time to double, corresponding to roughly 1/256 of a second.

mojofunk:
I just tried this change and it definitely seems to give an improvement in accuracy.

chaot4:
For what it's worth -- I reviewed the code two weeks ago and it seems correct (just forgot to write a comment).

mojofunk:
I just rebased and pushed this change to ardour master as revision 8d408d000d and it will be in a nightly build >= 4.6.382

I also added you to authors in the About dialog, thanks a lot.

This PR can now be closed.

PR #200, opened 2016-02-01
Proposed fix for bug 6558, ambiguous external files with same name 

mojofunk:
Proposed fix for bug 6558. It is a bit unclear what the intention is behind the "origin" member and whether it is only meant to be used for external files etc but if this fix is considered correct I can commit it.

This branch includes whitespace fixes, the actual fix is a relatively small part.

### DIFF
--- a/libs/ardour/audiofilesource.cc
+++ b/libs/ardour/audiofilesource.cc
@@ -141,6 +141,11 @@ AudioFileSource::AudioFileSource (Session& s, const XMLNode& node, bool must_exi
 		throw failed_constructor ();
 	}
 
+	if (Glib::path_is_absolute(_origin) &&
+	    Glib::file_test(_origin, Glib::FILE_TEST_EXISTS)) {
+		_path = _origin;
+	}
+
 	if (init (_path, must_exist)) {
 		throw failed_constructor ();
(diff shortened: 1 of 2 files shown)


mojofunk:
With this change the external directory that contains the file is still added to the session audio-search-path even though it isn't required any more to find the file. So that could also be changed if it isn't required for some other functionality.

pauldavisthefirst:
timbyr: "origin" is used primarily as a reference for time-stretched data. when a double-stretch occurs, rather than re-stretch the already stretched data (now in its own source), we compute the real stretch factor relative to zero stretch, go back to the "origin" source file, and generate the stretch from there.

pauldavisthefirst:
... hence, this fix is likely not correct, but might be .... can't think about the details right now.

mojofunk:
I've added some notes in bug report for you to have a look at when you get time, better to continue discussion there: http://tracker.ardour.org/view.php?id=6558

I'll leave this pull request open in the meanwhile until it is determined whether it is an appropriate fix

mojofunk:
A change equivalent to the last commit in this series has been committed to master branch as dbf86a49 so closing this PR

PR #201, opened 2016-02-02
Fix bug with PT5 and add preliminary PTX support

zamaudio:
- Previously, the PT5 wav files would appear jumbled in the regions, they did not match up correctly, now they do.
- Previously, PTX support was non-functional, now PT10 and PT11 works with 3 point audio on files/regions/tracks.


x42:
rebases as Ardour 4.6-336 fbbde01
Cheers!

PR #202, opened 2016-02-11
Fix 6609: Right-click in active track name box crashes

elgoun:
When right-clicking in the text entry, popup menu grabs focus. Consequently, the "focus out" handler is called, destroys the text entry and replaces it by the label name of the track.
When menu pops up, it tries to access to a widget no longer available.


mojofunk:
Thanks, this PR was rebased and merged  to ardour master as revision 1fbe4253aa

please close

PR #203, opened 2016-02-11
Fix 6677: Post-export script reinterprets timestamp format placeholde…

elgoun:
…r giving incorrect filename

Due to localtime and its statically allocated buffer, time_struct variable is
set at construct time but its value changes over time due to subsequent calls to localtime
in ardour process. Replacing localtime by localtime_r fix the problem.

This also fix 6713: Name of Audio (timestamp) does not match with written Filename in CD-Cue file


mojofunk:
Thanks, this PR was rebased and merged to ardour master as revision a3dd27c41b

please close

PR #205, opened 2016-02-18
 Fix 6772: crash when clicked "plugin analysis" on any linux-vst plugin

elgoun:
In VSTPlugin::connect_and_run, make _plugin->processReplacing called
 with the right buffer.

 See notes on Mantis before merge.


x42:
VST does not support in-place processing. 

Connecting outputs directly to buffers is not correct (they may be the same as input buffers).  We need to provide scratch buffers and copy back later.

PR #207, opened 2016-02-21
use matching floating point types

nphilipp:
I needed this change to compile Ardour with the GCC 6 prerelease which we have in Fedora Rawhide, without it, it would trip over the arguments to `std::max` having different types (`double` and `float` respectively).

### DIFF
--- a/libs/ardour/parameter_descriptor.cc
+++ b/libs/ardour/parameter_descriptor.cc
@@ -177,9 +177,9 @@ ParameterDescriptor::update_steps()
 			step      = step      / logf(30.0f);
 			largestep = largestep / logf(30.0f);
 		} else if (integer_step) {
-			smallstep = 1.0;
-			step      = std::max(1.0, rint(step));
-			largestep = std::max(1.0, rint(largestep));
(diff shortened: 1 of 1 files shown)


x42:
odd, `rint()` does returns a double. (see `man 3 rint`)
This patch currently breaks compilation with clang and gcc4, gcc5 and MSVC.

`rintf()` would return a float.  I can only assume  that gcc6 tries to be smart.  Since the parameter here is a float gcc6 turns `rint()` into `rintf()`.

The proper solution will be to use   `std::max (1.f, rintf (step))`

x42:
updated to use rintf() in Ardour 4.7-16-g30497db - please test

nphilipp:
I assume that the man page is for the C library version of `rint()`. If cppreference.com can be trusted, `rint()` in C++11 or later [returns the type of its argument](http://en.cppreference.com/w/cpp/numeric/math/rint), so `float` for `float` or `double` for `double`.

The changes you committed made it build again with both gcc 5.3.1 and (pre-) 6. Thanks!

PR #208, opened 2016-02-22
don't trip over missing `nm` tool

nphilipp:
Fix for this Fedora bug: [ardour4 startup script needs nm command from binutils](https://bugzilla.redhat.com/show_bug.cgi?id=1289349)


pauldavisthefirst:
rebased to master and merged.

PR #211, opened 2016-02-26
Fixes leaking mixerstrip tooltips

napcode:
These tooltip strings have been leaking.


pauldavisthefirst:
manually re-created. thanks.

PR #212, opened 2016-02-26
Bugfix/invalid read on new session

napcode:
This fixes an invalid access (via Timer -> ProcessorEntry::Control::slider_adjusted () ->  AutomationControl::get_value) to a now longer existing session. 
The monitor section failed to instantiate upon opening the 2nd session due to an exception that occurred when registering the monitor actions (action already registered). |monitor_actions| has static storage-class and must be initialized only once.

### DIFF
--- a/gtk2_ardour/mixer_ui.cc
+++ b/gtk2_ardour/mixer_ui.cc
@@ -432,6 +432,8 @@ Mixer_UI::add_strips (RouteList& routes)
 		}
 
 	} catch (const std::exception& e) {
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
Not really acceptable. the action changes are heading in the right direction, but proctoggle shouldn't be assigned twice. the assert(false) is not acceptable. It is entirely possible for this exception to be thrown and for things to carry on. We do not crash the program in favor of not leaking ...

pauldavisthefirst:
rebased and merged. thanks.

PR #213, opened 2016-02-29
Keyeditor improvements

napcode:
This PR:
- fixes a minor typo in one of the action groups
- adds sorting to the keybinding editor for both columns
- fixes a bug that prevents removing newly created bindings (to reproduce: add a binding, remove it, 
restart ardour, binding is still there but can now be deleted properly)
- adds a dialog for colliding keybindings

TODO:
A few things I'd like to do in the near future
- the indentation in bindings.* is messy. I'd like to create a NOP commit / PR to fix this
- allow alternate keybindings (as separate column, probably just 1) and remove the *-alt actions
- filtering in keyeditor
- find a way to display the multiple [show, attach, detach, etc.] entries in keyeditor properly

Please comment if there are any thoughts about the TODO list :)


x42:
rebased as Ardour 4.7-291-g290d9e5

PR #215, opened 2016-03-04
Improved pop up menu display in editor regions list

yonideshi:
- Proposed fix for regions list drag and drop
After right-click on the regions list pop up menu appears in the middle of drag and drop operation and executing "remove unused" regions can't be accomplished with a mouse.

- Some whitespace corrections

### DIFF
--- a/gtk2_ardour/editor_regions.cc
+++ b/gtk2_ardour/editor_regions.cc
@@ -189,6 +189,7 @@ EditorRegions::EditorRegions (Editor* e)
 	_display.get_selection()->set_mode (SELECTION_MULTIPLE);
 	_display.add_object_drag (_columns.region.index(), "regions");
 	_display.set_drag_column (_columns.name.index());
+	_display.set_drag_button_mask(Gdk::BUTTON1_MASK);
 
 	/* setup DnD handling */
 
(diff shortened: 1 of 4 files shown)


x42:
Which OS/Window-system is that that allows to initiate a DnD operation with right-click?
Which version of gtk? I cannot reproduce the described issue.

Also note that Ardour is by default built with a patched version of gtk which does not pass button-1 events though to the GtkEntry embedded in the treeview. see http://nightly.ardour.org/list.php#build_deps

PS. We try avoid touching files in `libs/*` and `gtk2_ardour/` in the same git commit.

yonideshi:
You are correct. I can't reproduce it with official Ardour 4.7 release. Please close.

x42:
I have not been able to re-produce this right-click/DnD on debian/testing with vanilla gtk 2.24.29-1 either (xorg, mate and xorg, openbox).

x42:
If this does affect distro builds, this patch is still interesting.

We generally try to get patches upstream and avoid quirks. Though in some cases e.g. the aforementioned GtkEntry + DnD, it wasn't possible. Some parts of Ardour are very _special_ cases.

yonideshi:
I have Ubuntu 15.10 with gtk 2.24.28-1. After today system update the problem has gone, so it was probably some bug not related to Ardour and just waste of my and your time...

yonideshi:
Update: Problem probably appears only in Ubuntu, and is related to new overlay scrollbars which seems to be buggy.

PR #216, opened 2016-03-06
Formatting

napcode:
Hi,

The indentation in bindings.* was pretty messy. In order to solve this automatically I created a clang-format style configuration which should be pretty close to the current coding style in the code base. Nevertheless, there are some minor differences:
- Putting the return type on separate lines for function definitions also puts all return types on inline definition on separate lines. This is in contrast to the current style but clang-format can't handle it better. 

This PR:
- creates a clang-format style definition (usable with clang-format version >= 3.6)
- reformats bindings.* using this definition
- removes some uncharacteristic comments 
- changes no code functionality

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,25 @@
+﻿---
+
+Language: Cpp
+AlignAfterOpenBracket: 'true'
+AlignTrailingComments: 'true'
+AlwaysBreakTemplateDeclarations: 'true'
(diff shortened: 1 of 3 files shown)


pauldavisthefirst:
All components of this PR have been applied manually using "different" approaches. Thanks for noticing how messed up the bindings.* code was.

PR #218, opened 2016-03-08
fix __cpuid() on x86

nphilipp:
Here's the corresponding Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1312551

TL;DR: Ardour's __cpuid() implementation doesn't work on x86 (32bit).

The previous version used memory operands that gcc (probably dependent
on optimization flags and/or version) could address relative to the
stack pointer, but pushing %ebx onto the stack changed it. Here, the
address of the regs array is put into %esi and the individual members
are written into directly.

Patch kindly provided by @fweimer.

### DIFF
--- a/libs/pbd/fpu.cc
+++ b/libs/pbd/fpu.cc
@@ -48,32 +48,25 @@ FPU* FPU::_instance (0);
 static void
 __cpuid(int regs[4], int cpuid_leaf)
 {
-        int eax, ebx, ecx, edx;
         asm volatile (
 #if defined(__i386__)
 	        "pushl %%ebx;\n\t"
 #endif
-	        "movl %4, %%eax;\n\t"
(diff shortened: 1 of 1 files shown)


fweimer:
A completely different solution would just use the `<cpuid.h>` header provided by recent GCC versions.

x42:
Thanks! I'll queue this up for testing on all the build-hosts and targets.

Ising recent gcc is not an option.  Some derivatives of Ardour are compiled with MSVC (with WinXP compatibility).  Official  Ardour releases are compiled with gcc-4.2 to be backwards compatible with older GNU/Linux and OSX systems.

x42:
It survived the smoke test :) merged/rebased as Ardour 4.7-317-g5a41b8

PR #219, opened 2016-03-11
Fix #5884, #4928, #5697: errors when using timestretch tool

potocpav:
Details are in the bug tracker for #5884, particularly the last comment:
http://tracker.ardour.org/view.php?id=5884
The other two bugs that are also probably resolved:
http://tracker.ardour.org/view.php?id=4928
http://tracker.ardour.org/view.php?id=5697

As mentioned in #5884, I am not sure if this is the best way to fix the issue. Maybe it's better to change `Region::use_sources` to take an additional argument, master_sources? I am not, however, familiar enough with C++ & the codebase to do that.

### DIFF
--- a/libs/ardour/region.cc
+++ b/libs/ardour/region.cc
@@ -270,6 +270,7 @@ Region::Region (boost::shared_ptr<const Region> other)
 	_hidden = false;
 
 	use_sources (other->_sources);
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
manually rebased against master and then merged. thanks for this fix (which is definitely the correct way to fix the issue).

PR #220, opened 2016-03-19
Renaming issue in regions list

yonideshi:
Small correction for renaming regions in regions list. Problem affects regions with more than 1 audio channel. When user tries to rename it but name was not changed (abandoned edit) then "[2]" is added to the name every time. It results in region name like this "Audio 1.1 [2] [2] [2] [2] [2] [2]" after a few tries. Also a real name is not reflected in region list.


x42:
good find. rebased and merged as Ardour 4.7-478 35e4c99
Cheers!

PR #221, opened 2016-03-24
Non single key keycodes

yonideshi:
Key combinations like Ctrl-F5, Alt-Insert were not loaded. They worked only for current session, but disappeared after reload  despite the fact there were correct entries in ardour.keys.


pauldavisthefirst:
manually rebased against master and pushed. thanks for this fix!

PR #222, opened 2016-03-28
Don't crash when keyeditor is empty

napcode:
The recent changes to the bindings capitalization caused a crash in the keyeditor when an outdated ardour.keys file rendered the editor empty.


pauldavisthefirst:
manually rebased against master, and pushed. thanks for this fix!

PR #223, opened 2016-04-01
bounced region name issue

yonideshi:
Bounced regions were named after track. Previous version of Ardour added "bounce" to the name. It seems that it was overlooked. Sometimes it is difficult to find bounced region when there are tens or hundred of regions on the list. Methods new_audio_source_path and new_midi_source_path correctly enumerate and "legalize" files paths. Probably this suffix could be added a region name only.

### DIFF
--- a/libs/ardour/session.cc
+++ b/libs/ardour/session.cc
@@ -5723,14 +5723,12 @@ Session::write_one_track (Track& track, framepos_t start, framepos_t end,
 		goto out;
 	}
 
-	legal_playlist_name = legalize_for_path (playlist->name());
-
 	for (uint32_t chan_n = 0; chan_n < diskstream_channels.n(track.data_type()); ++chan_n) {
 
-		string base_name = string_compose ("%1-%2-bounce", playlist->name(), chan_n);
+		string base_name = string_compose ("%1-bounce", playlist->name());
 		string path = ((track.data_type() == DataType::AUDIO)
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
This commit will break for multichannel tracks, because you've removed the channel number from the name. I like the general idea. Care to fix that particular problem?

yonideshi:
1. "base_name" was unused variable, so channel info was not considered in this part of code
2. "new_audio_source_path" method calls "legalize_for_path" internally and then rebuilds file name correctly using channel number information
3. Summarizing: "new_audio_source_path" method previously took "/path/file_name" rebuild it with channel info and took care of collisions. With this commit it will take "/path/file_name-bounce" and will follow as before. 

Thanks.

luzpaz:
Is this ticket still relevant ?

pauldavisthefirst:
hard to say. It's not an issue I've seen raised over the lat 6 years ....

luzpaz:
@yonideshi are you planning to follow-up ?

x42:
Since Ardour 6.3  (specifically f564c924932e and 80a7d32e07e7) there is a prompt for new region-names to mitigate this issue.
Rather than unconditionally adding a "bounce" suffix the user is prompted to specify a name.

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #225, opened 2016-04-09
Support to set gain automation by OSC

johannes-mueller:
This is a retry of #170. I felt like hacking on miniLAC and reimplemented it according to the ideas of @x42.

So greetings from miniLAC :)
  

Added support for OSC message

"/ardour/routes/gain_automation" rid state

where rid is the remote route id and state is a character representing
the desired gain automation state:

'm': Manual (Off)
'p': Play
'w': Write
't': Touch

### DIFF
--- a/libs/surfaces/osc/osc.cc
+++ b/libs/surfaces/osc/osc.cc
@@ -487,6 +487,8 @@ OSC::register_callbacks()
 		REGISTER_CALLBACK (serv, "/ardour/routes/send/gainabs", "iif", route_set_send_gain_abs);
 		REGISTER_CALLBACK (serv, "/ardour/routes/send/gaindB", "iif", route_set_send_gain_dB);
 
+                REGISTER_CALLBACK (serv, "/ardour/routes/gain_automation", "is", route_set_gain_automation);
+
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
Sorry :(

Changes occuring in the VCA2 branch suggest the way forward here. I think we would want a more generic message:

/ardour/route/automation_state "gain" "w"

thus allowing for

/ardour/route/automation_state RID "solo" "m"
/ardour/route/automation_state RID "mute" "t"

and so on. 

and for plugins:

/ardour/route/plugin_automation_state RID plugin_ID parameter_ID "m"

The vca2 branch has made all the controls for all automatable state much more explicit and they now own their state, rather than referencing properties of their owner (a Route).

I'll wait to see if x42 has opinions too. I expect/hope to merge the vca2 branch with master sometime in the next couple of weeks.

johannes-mueller:
Never mind. I will take at VCA2.

johannes-mueller:
As by now Ardour comes with Lua scripting, this PR is obsolete. The functionality that I had in mind is now achievable in a more elegant way by Lua scripting.

PR #226, opened 2016-04-16
ptformat: Make PT import more resilient to bad user choices and displ…

zamaudio:
…ay messages

Previously, libptformat would attempt to parse all kinds of files,
now the library stops parsing when the version number and session rate
is outside valid ranges, returning an error code to the caller.

If there is a valid PT file detected, but some audio files are missing,
Ardour now pops up an error message to inform the user that some files
may be missing from the import.  A success message is displayed otherwise.

Signed-off-by: Damien Zammit <damien@zamaudio.com>


x42:
merged as Ardour 4.7-751-g277893b

PR #227, opened 2016-04-23
compilable under BSD

devnexen:
This branch is to make it possible to compile under BSD (tested under Free and OpenBSD), avoiding linuxisms. Hope it finds its way :)

### DIFF
--- a/gtk2_ardour/video_timeline.cc
+++ b/gtk2_ardour/video_timeline.cc
@@ -33,10 +33,6 @@
 #include "video_timeline.h"
 #include "video_tool_paths.h"
 
-#ifdef PLATFORM_BSD
-#include <sys/wait.h>
-#endif
-
 #include <gtkmm2ext/utils.h>
 #include <pthread.h>
 #include <curl/curl.h>
--- a/gtk2_ardour/video_timeline.cc
+++ b/gtk2_ardour/video_timeline.cc
@@ -33,6 +33,10 @@
 #include "video_timeline.h"
 #include "video_tool_paths.h"
 
+#ifdef PLATFORM_BSD
+#include <sys/wait.h>
+#endif
+
 #include <gtkmm2ext/utils.h>
 #include <pthread.h>
 #include <curl/curl.h>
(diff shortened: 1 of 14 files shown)


x42:
Odd that `obj.linkflags = bld.env['LDFLAGS']` is needed. waf adds that automatically.

I just tried   `LDFLAGS="-L/testing" ./waf configure`
it's printed in the config log and also used when linking (check with ./waf -v)

libs/ardour/wscript
```diff
@@ -373,7 +373,7 @@ def build(bld):
                         'SNDFILE','SAMPLERATE','LRDF','AUDIOUNITS', 'GIOMM',
                         'OSX','BOOST','CURL','TAGLIB','VAMPSDK','VAMPHOSTSDK','RUBBERBAND']
     obj.use          = ['libpbd','libmidipp','libevoral',
-                        'libaudiographer',
+                        'libaudiographer', 'libexecinfo',
```
x42: `libexecinfo` should be probably be added as uselib "EXECINFO" or otherwise special cased for BSD.

devnexen:
true especially EXECINFO was defined I forgot to remove the unneeded one. Thanks for the feedback.

x42:
looking at the combined diff ( https://patch-diff.githubusercontent.com/raw/Ardour/ardour/pull/227.diff )
there are still a few `obj.linkflags = bld.env['LDFLAGS']`  additions. Are those needed?

The ones to initialize it are fine, IMHO (allows to consistently use `+=` elsewhere) any thoughts, @pauldavisthefirst ?

gtk2_ardour/video_timeline.cc
```diff
@@ -33,6 +33,10 @@
 #include "video_timeline.h"
 #include "video_tool_paths.h"
 
+#ifdef PLATFORM_BSD
+#include <sys/wait.h>
+#endif
+
```
x42: Why is this needed here?  `WNOHANG` ?  if so it might be better in `libs/pbd/pbd/system_exec.h`
where the `wait()` wrapper is defined.

PR #228, opened 2016-04-26
Build Ardour for FreeBSD

beerml:
Tested in FreeBSD v10.2, v10.2 and v11current with clang v3.4, v3.8.
If gcc is installed on the system you have to use CC=clang CXX=clang++ in front
of ./waf configure


beerml:
Seems, I messed up this pull request quite a bit.
I close it for now and try again when I figured out how to rebase correctly ;-(

PR #229, opened 2016-05-26
Update fr.po

MrHeadwar:
Updated french translation


pauldavisthefirst:
If you're willing to revisit the French translation, we are nearing a string freeze for release 5.0. All 3 .po files need some work.

x42:
marged as part of #249

PR #232, opened 2016-06-23
Enable build for FreeBSD.

beerml:
Use
./waf configure --no-lxvst
for configuration!

### DIFF
--- a/gtk2_ardour/ardour_ui.cc
+++ b/gtk2_ardour/ardour_ui.cc
@@ -33,6 +33,11 @@
 #include <sys/resource.h>
 #endif
 
+#ifdef __FreeBSD__
+#include <sys/types.h>
+#include <sys/sysctl.h>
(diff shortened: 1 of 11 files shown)


beerml:
This works for FreeBSD 10.x and FreeBSD 11Current with the clang toolchain (including LV2 support).
With gcc I can compile but not link.
If both toolchains are installed, use CC=clang CXX=clang++ ./waf configure --no-lxvst
All dependencies besides the LV2 stack can be resolved via FreeBSDs package manager (pkg).

x42:
Rebased and slightly tweaked as 86763cba and 5.0-pre0-335-gb027d9d

Changing the `--also-include` etc stored options in the wscript is really not good practice.  Those are for a user to set.  The rest looks good.  commited with some minor refactoring (trailing whitespace, etc)

PR #233, opened 2016-07-05
Fix ldd(1) parsing on NetBSD

krytarowski:
* NetBSD

```
$ ldd /bin/cat
/bin/cat:
        -lc.12 => /lib/libc.so.12
```

 * Linux

```
$ ldd /usr/bin/cat
        linux-vdso.so.1 =>  (0x0000726abb373000)
        libc.so.6 => /lib64/libc.so.6 (0x0000726abafa7000)
        /lib64/ld-linux-x86-64.so.2 (0x0000726abb374000)
```

### DIFF
--- a/gtk2_ardour/ardour.sh.in
+++ b/gtk2_ardour/ardour.sh.in
@@ -23,7 +23,7 @@ fi
 
 ## Glib atomic test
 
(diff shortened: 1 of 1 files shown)


x42:
rebased as 5.4-17 f9c83a0

PR #234, opened 2016-07-05
Detect free space on NetBSD

krytarowski:


x42:
Rebased as Ardour 5.4-18 5b1b3a7

PR #235, opened 2016-07-05
Handle NetBSD behavior of handling of _POSIX_SOURCE or _XOPEN_SOURCE

krytarowski:
On NetBSD once these symbols are defined, we are disabling additional
features rather than enabling hidden ones like on Linux. To overcome it
define explicitly _NETBSD_SOURCE in waf files.


krytarowski:
I will redo it for next version.

PR #236, opened 2016-07-05
NetBSD ships with <sys/endian.h> not <endian.h>

krytarowski:


x42:
Rebased as Ardour 5.4-19 f69776e

PR #237, opened 2016-07-05
NetBSD ships with pollts(2) similar to Linux specific ppoll()

krytarowski:


x42:
Rebased as Ardour 5.4-20 1ac7dad

PR #238, opened 2016-07-05
NetBSD uses statvfs for the functionality of statfs on Linux

krytarowski:


x42:
Rebased as Ardour 5.4-21 9d3077b

PR #239, opened 2016-07-05
NetBSD ships with backtrace(3) in libexecinfo

krytarowski:


x42:
Rebased as Ardour 5.4-22 e426f07

Thanks a lot! Finally NetBSD support should all be upstream.

krytarowski:
Thank you!

PR #240, opened 2016-07-06
Don't add audio outs to non-audio routes with strict I/O

FrnchFrgg:
If a route has strict I/O then the main delivery follows the output of
the last processor, but libardour ensures it has at least as many
outputs as inputs in the master strip. A good consequence is that mono
tracks get their expected panner.

An akward side-effect is that MIDI-only routes (e.g. midi tracks or busses
without a synth) get two audio channels that have no use (and indeed no
panner is added because there is nothing to pan).

Skip the completion of audio outs if there was no audio out to begin
with.


x42:
Rebased as Ardour 5.0-pre0-471-g6911702 -- Thanks!

PR #245, opened 2016-07-11
Make MIDI monitor a pass-through for audio and midi

FrnchFrgg:
MIDI monitor only accepted midi data and output that same data. That was
logical for a MIDI plugin, but a consequence is that automatic pin
configuration makes MIDI monitors opaque to audio data, which means
drag'n'dropping a MIDI monitor for debugging purposes can suddenly cut
audio, or even change the channel count if strict I/O is enabled.

Improve the MIDI monitor so that it passes through all incoming data
unchanged.


x42:
rebased and merged as 30240b3

PR #248, opened 2016-07-15
Better choice for ports on external send creation

FrnchFrgg:
Ardour tried to make an educated guess at the initial number of outputs
for a new send. It used the channel configuration of the master bus, if
it existed, else the channel configuration of the route itself.

That guess is good in most cases, but in the case of a track/bus without
audio channels, creating a send with audio doesn't make sense. In that
case, also use the route outputs as a base for the send configuration.


x42:
nice!.  rebased & merged as 786cb74

PR #251, opened 2016-07-17
Simplify the logic in HiAndLowPass filter

FrnchFrgg:
And beef up comments for readers using the filter as an example.

### DIFF
--- a/scripts/HiAndLowPass.lua
+++ b/scripts/HiAndLowPass.lua
@@ -4,7 +4,7 @@ ardour {
 	category    = "Filter",
 	license     = "GPLv2",
 	author      = "Ardour Team",
-	description = [[Example Ardour Lua DSP Plugin]]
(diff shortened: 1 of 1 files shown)


scripts/HiAndLowPass.lua
```diff
@@ -160,61 +166,60 @@ function dsp_run (ins, outs, n_samples)
...
-			-- copy output of high-pass into "processing memory"
+			-- prepare scratch memory (from high pass output)
 			ARDOUR.DSP.copy_vector (mem:to_float (off), outs[c]:offset (off), siz)
 
-			if cur[4] > 0 then
-				-- clear output, Low-pass mixes interpolated data into output,
-				-- in which case we just keep the output
-				ARDOUR.DSP.memset (outs[c]:offset (off), 0, siz)
+			-- run at least |lo| biquads...
+			for k = 1,ho do
```
x42: 1,lo is needed here

x42:
fixed and rebased as d527fdf

PR #252, opened 2016-07-17
Vca2

cooltehno:
Here are my-dark.color files for Ardour v.4 with screenshots.
New theme "clear-gray" is added. General screenshot with new names: "caineville" & "cubasish".
Also general screenshot for four themes with a "clear-gray" added.


pauldavisthefirst:
There are lots of problems with this PR. 

First, the vca2 branch is no longer valid. Second, the changes should be made directly to the files in gtk2_ardour/themes, not in a new folder. I'm very happy to incorporate your theme work into Ardour, but changes from now on need to take place directly on the "canonical" theme files, not additional copies of them.

Finally, the actually history here ("Add 1", "Delete 1", "Add 1", "Delete 1") is not appropriate for merging into a long-lived shared repository. 

So, what shall we do?

cooltehno:
Good day, Paul!

Thanks for detailed explanation. I try to use GitHub as a designer. Sorry for stupidness)
My goal was to make a source for developer's work.

1) About vca2 - I understood. (suppose, because it was merged to master). Now, if I want to make PR - I'll use the master branch (or other, that you'll order).

2) I can't make a new kind of theme file (I mean "name-ardour.colors" instead of "my-dark.colors"), because I use Ardour 4. I'm not sure that simple renaming of these files could be good. Therefore I've decided to make a new folder, that somebody (programmer) could take any file and rename it correctly.
   If I want to make a new kind of file, should I install Ardour5, make color changes , save with "new-ardour.colors" name and after make a PR?

3) That labyrinthine history is the result of my way to make new folders. In Git Hub site I didn't find any knob to make a folder, but in google I found the way "to make a new file in a new folder" (just to wright a foder name in path). So I've made few new empty files called "1" in new folders and after upload my theme files and deleted "1" - files. Monkey's way, but I don't know other)

P.S.
https://community.ardour.org/node/13784 - here I've noticed, that you use a screenshot with old names (ableton-like, cubase-like), therefore I uploaded  a screenshot with new names (cainville and cubasish) with my PR.
Using GitHub I make only problems, sorry for that. So I could use my usual way to share - googledrive))

zapashcanon:
> Finally, the actually history here ("Add 1", "Delete 1", "Add 1", "Delete 1") is not appropriate for merging into a long-lived shared repository.

GitHub now allows to squash commits and merge.

x42:
superseded by #258

PR #253, opened 2016-07-18
Optimize a-HiAndLowPass

FrnchFrgg:
- Better formula for the lpf smoothness parameter. It is computed only
  on init and rate change anyway...
- Only run as many Biquads as needed to save computing power, esp. when
  using low steepness and parameters are not changing.


x42:
rebased/merged as 1dcb54b

PR #260, opened 2016-08-04
workaround changes in glibmm 2.49.x

nphilipp:
Glib::RefPtr defines the operator bool() as explicit which breaks
comparisons like "some_ref_ptr == 0" or "... != 0".

https://bugzilla.gnome.org/show_bug.cgi?id=769502


pauldavisthefirst:
Where does this break?

nphilipp:
It breaks when building ardour against glibmm >= 2.49.1.

PR #261, opened 2016-08-13
AppData release tags need a date to be valid

nphilipp:
Or a timestamp, but that's deprecated. Create it from the last commit,
store it in/retrieve it from `revision.cc` and substitute it in the
template.

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

### DIFF
--- a/wscript
+++ b/wscript
@@ -165,10 +165,12 @@ def fetch_tarball_revision_date():
     with open('libs/ardour/revision.cc', 'rb') as f:
         content = f.readlines()
         remove_punctuation_map = dict((ord(char), None) for char in '";')
-        raw_line_tokens = content[1].decode('utf-8').strip().split(' ')
 
-        rev = raw_line_tokens[7].translate(remove_punctuation_map)
-        date = raw_line_tokens[12].translate(remove_punctuation_map)
+        raw_line_tokens = content[2].decode('utf-8').strip().split(' ')
+        rev = raw_line_tokens[4].translate(remove_punctuation_map)
+
+        raw_line_tokens = content[3].decode('utf-8').strip().split(' ')
+        date = raw_line_tokens[4].translate(remove_punctuation_map)
 
         return rev, date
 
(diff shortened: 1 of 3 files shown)


wscript
```diff
@@ -271,12 +281,16 @@ def create_stored_revision():
...
         #
         text =  '#include "ardour/revision.h"\n'
-        text += 'namespace ARDOUR { const char* revision = \"%s\"; }\n' % rev
-        print('Writing revision info to libs/ardour/revision.cc using ' + rev)
+        text += (
+            'namespace ARDOUR { '
+                'const char* revision = \"%s\"; '
+                'const char* date = \"%s\"; }\n'
+        ) % (rev, rev_date)
+        print('Writing revision info to libs/ardour/revision.cc using ' + rev + ', ' + rev_date)
```
x42: This will break bundle-scripts, 'revision' is expected in the 2nd line:

https://github.com/Ardour/ardour/blob/master/tools/define_versions.sh#L13

nphilipp:
The second commit should fix `define_versions.sh` parsing `revision.cc`.

pauldavisthefirst:
Why not just put a newline after the revision?

nphilipp:
Sorry for the late reply.

> Why not just put a newline after the revision?

`revision.cc` looks like this when I apply the patches to current master:

    #include "ardour/revision.h"
    namespace ARDOUR {
        const char* revision = "5.5-74-ge94d12c";
        const char* date = "2016-12-09";
    }

Do you mean something different?

nphilipp:
@pauldavisthefirst ping, what did you mean with "put a newline after the revision"?

tartina:
I think Paul meant something like this:
```cpp
#include "ardour/revision.h"
namespace ARDOUR { const char* revision = "5.9";
const char* date = "2017-05-14"; }
```
and patching `define_versions.sh` would not be needed.

nphilipp:
@tartina, I adapted the patch you sent me a while ago and applied it to my branch. I've done some tests with `tools/define_versions.sh` and it failed if the additional data is on a third line, so I moved it back to the second line in the file.

@pauldavisthefirst, @x42 what do you think?

NB: I'm on vacation for a week, so I might not immediately reply.

nphilipp:
Ping?

nphilipp:
Hi @pauldavisthefirst, @x42 I'd like this change to get into 6.0. Is there anything I need to rework for that to happen?

x42:
6.0 is still ages away, development on it has only just begun.

anyway, it looks fine to me now (w/o the newline)
```
#include "ardour/revision.h"
namespace ARDOUR { const char* revision = "XXX"; const char* date = "XXX"; }
```
should be parsed correctly by `tools/define_versions.sh`

nphilipp:
Ping? :)

tartina:
Any news on this?

pauldavisthefirst:
manually rebased/merged

PR #262, opened 2016-08-16
allow linking unbundled version of qm-dsp

nphilipp:
In the course, rename the bundled, built version libqm-dsp so it matches
packaged versions.

### DIFF
--- a/libs/qm-dsp/wscript
+++ b/libs/qm-dsp/wscript
@@ -23,10 +23,17 @@ def options(opt):
     autowaf.set_options(opt)
 
 def configure(conf):
-    conf.load('compiler_cxx')
-    autowaf.configure(conf)
+    if conf.is_defined('USE_EXTERNAL_LIBS'):
+        conf.check_cxx(header_name="base/Pitch.h", mandatory=True)
+        conf.check_cxx(lib="qm-dsp", uselib_store="QMDSP", mandatory=True)
+    else:
+        conf.load('compiler_cxx')
+        autowaf.configure(conf)
 
 def build(bld):
+    if bld.is_defined('USE_EXTERNAL_LIBS'):
+        return
+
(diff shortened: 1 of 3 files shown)


x42:
Did you test Onset Detection (Rhythm Ferret) and Transient Analysis (Prefs > Audio > Region Analysis ;; ctrl left/right arrow jump)?

Last I investigated in updating our copy of qm-dsp/qm-vamp-plugins, there were subtle differences.

nphilipp:
How subtle, what do I need to look out for?

nphilipp:
Just noticed that it fails to use the (external) static library for linking, don't merge right now (not that you would).

nphilipp:
Okay, seems to link alright now, both to the external and bundled libraries.

I've tried out both Percussive and Note Onset with both and found no discernible differences, other than that the canvas apparently doesn't like so many small regions -- it seems to miss drawing the separation sometimes, clicking into the region will show that both bundled and unbundled have ferreted out the same results. Ardour is groaning under the load though, everything is slow as molasses with that many split regions (shouldn't have done it to a ~4:30mins song, 4 times (once per each Percussive/Note Onset and bundled/unbundled).

nphilipp:
@x42 Oh yeah, if you can tell me how to test Transient Analysis (and what to look out for), that'd be great.

nphilipp:
@x42 I've tested Transient Analysis with a slightly shorter piece (a little more than 30 seconds), ctrl-clicked forward and backward and it looked as if using the bundled or unbundled version doesn't make a difference here either.

nphilipp:
...and retested Onsets with the shorter piece (and transient analysis turned off). I can't spot a difference.

x42:
Rebased as Ardour 5.4-16  0aaded5

PR #263, opened 2016-08-17
Create AKAI_MIDIMix_Normal_Mode.map

eighthkeepa:
My midi map for AKAI Midimix. This one has no volume knobs bindings.

The only buttons that will be lit upon press - mute buttons.


x42:
Rebased as 5.4-14 c68a103

Thanks a lot, and sorry for the delay merging this.

PR #264, opened 2016-08-17
Create AKAI_MIDIMix_EQ_Mode.map

eighthkeepa:
My midi map for Akai MIDIMix. This one has volume knobs bindings included, so you can control equalizers of each track.

The only buttons that will be lit upon press - "mute" buttons.


x42:
Rebased onto master as  Ardour 5.4-15 ca90e56

PR #265, opened 2016-09-03
Expose record enable/disable to Lua bindings

johannes-mueller:
Is there a reason, why rec enabling and disabling is not exposed to the Lua bindings?

Recenabling in the Lua bindings is quite useful because one can setup a Lua EditorAction script that prepares everything (including recenable) for the recording.

**Update:** It seems that there are other Lua bindings missing. Other missing ones  I would find useful

* `Locations::remove()` in order to clear the markers when canceling a recording to start from scratch
* `Region::last_position()` and `Region::last_length()` to see how exactly a region has been moved or trimmed


x42:
merged as 5.3-52 8bef762

There are many bindings missing. First one we need to establish that it's safe to call the given method (rt-safeness, thread/context)  from Lua and ideally we also write an example script to exercise them and add relevant doxygen doc..  - we're already behind with the latter. It's a slow process.

PR #266, opened 2016-09-05
Make some functions which are needed for native VST support like snpr…

beerml:
…intf() visible in FreeBSD.

### DIFF
--- a/libs/fst/wscript
+++ b/libs/fst/wscript
@@ -88,4 +88,8 @@ def build(bld):
         'PACKAGE="' + I18N_PACKAGE + str(bld.env['MAJOR']) + '"',
         'LIBARDOUR="' + bld.env['lwrcase_dirname'] + '"',
         ]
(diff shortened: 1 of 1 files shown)


beerml:
This was tested with the DISTRHO Mini-Series plugins built natively for FreeBSD. 
Configure command which works now for FreeBSD is e.g.:
CC=clang38 CXX=clang++38 ./waf configure --optimize --ptformat --also-include=/usr/local/include --also-libdir=/usr/local/lib

x42:
Rebased as Ardour 5.4-24 8d8ca6a

Curious, what is `sys.platform` on freebsd? 
Would some more specific  `sys.platform == 'freebsd'`  or `sys.platform.startswith('freebsd`)`work as well?

beerml:
"print sys.platform"  returns freebsd10.
So yes, sys.platform.startswith('freebsd') should be fine.
Thanks for merging!

PR #267, opened 2016-09-06
Update clear-gray-ardour.colors

cooltehno:
correction_060916
This commit changes item "gtk_bases" from light (color22) to darker (color88). The knob of "Enable/Disable MIDI input" in Edit List had the same colour and was not visible. This commit makes it clear visible, and it has no much influence on a general design.


pauldavisthefirst:
Very glad to see you've mastered the use of git pull requests!

cooltehno:
Thanks, Paul! My git experience is the result of cooperation with your good assistants in IRC. Thanks for them and you for 5.3 release. Сongratulations!

PR #269, opened 2016-09-08
Lua export chapter marks

johannes-mueller:
I am experimenting with the Lua interface trying to create useful examples.

This is a script that exports  location markers, that are not xruns,  as mp4 chapter marks. This requires `Location::name()` to be exposed to Lua.


x42:
Rebased as 5.3-69-gcf0d06c. I've also change the file-path line to use `ARDOUR.LuaAPI.build_filename()` for portability (ie. backslash separators on windows).
Cheers!

PR #270, opened 2016-09-10
Example script to prepare the session to record

johannes-mueller:
Another example script to tryout the new possibilities of the Lua interface. 

This is to get the session ready to record a podcast. All the relevant settings ensured by just one action.

Personally I use it in combination with [wiimote recording control](https://github.com/linux-podcasting-hacks/wiimote-recording-control) of the Linux podcasting hacks.

It requires `Locations::remove()` to be exposed to Lua, as we want to clear the location markers before recording.


x42:
Nice.  rebased as e54ae02 and ba0eb858 (Ardour 5.3-64) Thanks!

PR #271, opened 2016-09-10
add request_play_loop() & get_play_loop() to luabindings

7890:


x42:
rebased as 98d9887. Thanks!

PR #274, opened 2016-09-15
add method ARDOUR::LuaAPI::get_plugin_insert_param()

7890:


libs/ardour/lua_api.cc
```diff
@@ -176,6 +176,18 @@ ARDOUR::LuaAPI::set_plugin_insert_param (boost::shared_ptr<PluginInsert> pi, uin
 	return true;
 }
 
+float
+ARDOUR::LuaAPI::get_plugin_insert_param (boost::shared_ptr<PluginInsert> pi, uint32_t which)
+{
+	boost::shared_ptr<Plugin> plugin = pi->plugin();
+	if (!plugin) { return NULL; }
```
x42: return NULL; //  is not a float.

libs/ardour/lua_api.cc
```diff
@@ -176,6 +176,18 @@ ARDOUR::LuaAPI::set_plugin_insert_param (boost::shared_ptr<PluginInsert> pi, uin
...
+
+	bool ok=false;
+	uint32_t controlid = plugin->nth_parameter (which, ok);
+	if (!ok) { return NULL; }
```
x42: ditto.

LuaAPI::get_plugin_insert_param (boost::shared_ptr<PluginInsert> pi, uint32_t which, bool& ok)

and   .addRefFunction() in luabindings.cc is probably the cleanest solution to this.

pauldavisthefirst:
should really be get_plugin_param(). PluginInserts are internal things that a LUA user doesn't really need to name.

7890:
it probably should however that's another topic. the method is adapting to current naming rules.

7890:
Paul, i've added get_processor_param(). The set/get combo is now:

.addFunction ("set_processor_param", ARDOUR::LuaAPI::set_processor_param)
.addFunction ("set_plugin_insert_param", ARDOUR::LuaAPI::set_plugin_insert_param)
.addFunction ("get_processor_param", ARDOUR::LuaAPI::get_processor_param)
.addFunction ("get_plugin_insert_param", ARDOUR::LuaAPI::get_plugin_insert_param)

x42:
get_processor_param() was rebased as 5.3-122-ga1fae47

Lookup by name was dismissed, names are not suitable as identifiers.

7890:
_by_name() does not intend to make a query by using a unique identifier. how could this misunderstanding be eliminated by using better naming?

7890:
i'd like to add that using an index that can change is also not a very good identifier. it's an index in the domain of that one plugin. an index 0 can point to different plugins at different points in time. this makes it possibly even a more dynamic "identifier" than what a plugin parameter name will be.
knowing that, _get_name() was never intended to overcome the absence of a robust uuid or similar. it's equally good or bad as using by_index().

7890:
using this github pull request facility ought to make it easy for the main repo and the contributor, but now it looks a bit like a semi-transparent mess in several ways.

pauldavisthefirst:
how is this a semi-transparent mess?

7890:
it's not obvious what was actually commited and commit messages got lost.

x42:
Everything from this pull request was reverted. 2b8979647b7989b127e0004c2e75cc999d42166d undoes everything from faf44386460b6cdda0aa249eb138e8fd82700194

7890:
is that the whole story?

x42:
The original pull request was not rebased onto master, the PR itself included a "merge from ardour/master"  so I squashed it.  Anyway git history is pretty clear.

7890:
if that's the whole story, let's leave it at that. good to see anyways that some method(s) are now available to use.

x42:
PS for future reference:

> should really be get_plugin_param(). PluginInserts are internal things that a LUA user doesn't really need to name.

not quite.  Since Lua exposes the internal libardour API,  plugins are correctly referenced via http://manual.ardour.org/lua-scripting/class_reference/#ARDOUR:PluginInsert  or Processor (one does not usually use the plugin(s) inside the PI directly.

PR #275, opened 2016-09-27
more Russian .po file updates

psemiletov:
Some updates and fixes to the Russian .po file.


prokoudine:
Hello Peter,

Thanks for submitting the update. However there are a few issues with the proposed changes. The most obvious one is adding back the context like "Window|". It should only be used in the original message to separate same message that should be translated differently in the context.

Secondly, I'm afraid, I cannot accept the change of the translation for "New..." as it diverges from all other GTK+ apps, and snapshot-related menu entries should rather use verbs. The rest looks really good so far.

x42:
Can you rebase this pull request to only include actual translation changes?

pauldavisthefirst:
Incorrect merge/rebase action; no followup on translation issues. Closed for now.

PR #282, opened 2016-10-29
Making buildable under OpenBSD.

devnexen:


pauldavisthefirst:
this generally looks good, but might be nicer to treat openBSD as its own case for the final two diff chunks, rather than a "mingw and NOT openBSD"

devnexen:
@beerml true that.

pauldavisthefirst:
manually rebased and merged.

PR #285, opened 2016-11-20
update german translation

edogawa23:


x42:
Rebased as Ardour 5.4-409 979d5d1. Cheers!

PR #290, opened 2016-11-27
fixed compilation on osx, minimal version was wrong

powerpaul17:


x42:
XCode usually 1 version backwards compatible. e.g. on 10.11, there's still the 10.10 SDK.

I suppose the actual issue is the common handling of [ 'mavericks', 'yosemite', 'el_capitan' ] in this case. Then again current wscript works for me as-is on Yosemite (10.10.5) and I know there are some compiling it on El Capitan without changes.

What version of OSX or macOS and XCode are you using?

powerpaul17:
I am on "Sierra" (10.12) and I had problems with some unavailable header files, unfortunately I can't remember now which ones exactly. I just noticed, that in the other cases (lion, etc) "-mmacosx-version-min" always corresponded to the actual minimal version of the case.

powerpaul17:
Maybe it requires adding an additional case for "Sierra" in which the minimal version is 10.9

pauldavisthefirst:
that would be better, in my opinion.

powerpaul17:
ok, I adapted the respective portions of the wscript.

pauldavisthefirst:
manually rebased and merged.

PR #292, opened 2016-11-29
Append channel number as in Automatable::describe_parameter

ngeiswei:
Fixing the fact that midi controller names coming from patch files don't have channel num appended to them.

I'm not totally sure it is the right way to do it but it works for me.

### DIFF
--- a/libs/ardour/instrument_info.cc
+++ b/libs/ardour/instrument_info.cc
@@ -151,7 +151,8 @@ InstrumentInfo::get_controller_name (Evoral::Parameter param) const
 		return "";
 	}
 
-	return control_names->control(param.id())->name();
+	std::string name(control_names->control(param.id())->name());
+	return string_compose(name + " [%1]", int(param.channel()) + 1);
 }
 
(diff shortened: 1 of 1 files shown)


x42:
Any opinion about using   `+ ((Config->get_first_midi_bank_is_zero()) ? 0 : 1)` instead of `+ 1`?

(I know the pref says "banks", but midi channels likewise have that "0..15"  vs "1..16" issue)

ngeiswei:
I would probably let `+ 1` until someone requests to have the option for `0..15` vs `1..16` rather than conflating this option with banks. According to my experience `1..16` is much more widespread for the user (and `0..15` from the developer). Think for instance of channel 10 which is often reserved for drums, I don't remember ever seeing in any user documentation the mention of channel 9 instead.

ngeiswei:
Any update on this? I've another PR coming but I'd rather have this one taken care of first.

pauldavisthefirst:
manually reimplemented to fit in with other changes in instrument_info.cc

PR #293, opened 2016-11-30
Proposed fix managing shared playlists (see #7150)

elgoun:
Actually, when duplicating a track with "share playlist", the current
playlist is owned by the new created track(orig-track-id). The sharing
mecanism is made by diskstreams pointing on the same(shared) playlist.
Since playlist now owned by the new track, selecting another playlist in
the original track "forgets" the playlist for this track.You can't
select the shared playlist anymore from the original track.

This commit adds a way to keep trace of shared playlist between tracks.


pauldavisthefirst:
manually rebased and merged.

PR #301, opened 2016-12-01
one typo

trebmuh:


x42:
See the comment above:
```
 XXX translators: use a string here that will be at least as long
  as the longest automation label (see ::automation_state_changed()
```

The current cases that need to fit  in engl.  are "Play", "Write", "Touch", "Manual". 
The lower case "Y" in Play has a descender, but  "Manual" is the longest.  so "Mgnual" is correct.

trebmuh:
OK, thanks for the explanation. So, because the automation control for the generic plugin UI are translated in French as "Manuel", "Lire", Écrire" and "Reprendre", I think I understand that the French translation of this string shouldn't be "Manuel" but "Rfprendre"?

PR #304, opened 2016-12-02
Fix crash when track creation failed draging region

elgoun:
Add more check when creating a new track from a region drag into editor's empty area.

Fix #6496


mojofunk:
I tested, rebased and pushed this commit to master branch. This PR can now be closed, thanks.

PR #305, opened 2016-12-02
Making one string translatable

trebmuh:
It should allow this dialog window to be translatable : 

ardour-to_be_translatable

I hope this is the correct way to do so. Please double-check what I've done before merging.


x42:
Merged as 5.5-4 90fcdcf

trebmuh:
Thanks.

PR #309, opened 2016-12-08
ifdef out uri_map when lv2_support is disabled

wargarblgarbl:
Currently when compiling Ardour on FreeBSD11 with the option --no-lv2, ardour will still attempt to use uri_map.h and lv2_plugin.h in a few places, even though in theory the option should disable it. This is a fix for this issue.


wargarblgarbl:
Going to squash the changes over to just one commit and resubmit.

pauldavisthefirst:
manually rebased and merged.

PR #310, opened 2016-12-18
Tracker Editor

ngeiswei:
## Overview

A rudimentary, read-only for now Tracker Editor, reminiscent of the tracker music editors of the 90s such as ProTracker for Amiga.

## Motivation

1. Bring the piano-roll and tracker style worlds together.
2. Leverage Ardour's professionalism for Tracker music editing.
3. Add new features that are missing in existing trackers.

## Status

1. The interface sucks
2. No editing yet

## Features

1. Arbitrary resolution (lines per beat)
2. Negative delay (have an event start a little bit before the corresponding line)
3. Display automation

## Remaining to be done

0. Write documentation.
1. Update tracker automation when editing track view automation.
2. Support audio tracks and trim automation.
3. Support multiple tracks and regions.
4. Support editing.
5. Create a dedicated beautiful widget instead of using Gtk::TreeView.
6. Remember what notes are on which tracks

### Piano-Roll view without the Tracker Editor
ardour-piano-roll-interface

### Tracker Editor on the visible midi region, 4 lines per beat, events that cannot be displayed are rendered as `***`
ardour-tracker-interface-4-lines-per-beat

### Switch to 8 lines per beat, now most events can be displayed
ardour-tracker-interface-8-lines-per-beat

### Show existing automations as well, bright values are actual values, grayed values are interpolations
ardour-tracker-interface-8-lines-per-beat-existing-automations

### Note velocities and delays are displayed as well
ardour-tracker-interface-8-lines-per-beat-vol-delay


ngeiswei:
I forgot to mention, this PR contains also some indentation fixes (whitespace -> tab) on some other files. If you want I can disentangle the commits in question and create a new PR just for that.

ngeiswei:
Oh, I forgot to access it right-click on the region then `<Region name>->MIDI->Tracker Editor (read only)...`

Of course once it supports audio tracks and displaying multiple regions and tracks at once, the access will have to be moved somewhere else.

x42:
Do you think it's realistic to stabilize it to go along with the minor release cycle (roughly every 2 months: next: end Jan/beginning Feb'17)?

Would it make sense to keep most of it in a branch until there are editing capabilities and interface "sucks less"?

All of the NO-OP, whitespace changes could go upstream easily. As for the rest: If it helps with maintenance, we could merge it and  use a configure option to `#ifdef` hide the entry-points  (not the whole code) until it's ready for release.

I haven't yet read though it all, yet, but it certainly seems like a nice addition.

Conceptually it seems odd that Ardour allows an arbitrary number of voices (up to 127 notes) and no time-grid for events.  While a tracker-interface is usually limited in both.  How do you plan to resolve this for editing capabilities?

ngeiswei:
In terms of stability, i.e. not crashing Ardour, I think it is already pretty good, at least I didn't manage to crash it, but I might have missed some cases of course.

In terms of adding editing + improving the interface, I don't think it would be ready by Jan/Feb, maybe, perhaps just editing. Regarding the interface, a simple trick that could improve it a lot would be to have 2 header rows, one over sets of columns to the separate between midi notes, midi automations, processor automations and main automations, one under it as we have now. Unfortunately Gtk::TreeView doesn't allow this. I just learned on IRC that you are moving away from Gtk, maybe this move will make improving the interface easier.

@x42 says:
"Conceptually it seems odd that Ardour allows an arbitrary number of voices (up to 127 notes) and no time-grid for events. While a tracker-interface is usually limited in both. How do you plan to resolve this for editing capabilities?"

1. The arbitrary number of voices is solved by intelligently dispatching notes depending on their timing and content, see https://github.com/Ardour/ardour/pull/310/files#diff-ca90a004714fec2b1ddae72fefe3b86fR49 there's a catch though, once editing will be supported I'll need a way to store across sessions on which column notes are. For now it is just automatically assigned.
2. The time-grid interface is solved by using the `Delay` column, which can be both positive and negative. The delay is automatically calculated depending on the grid granularity (number of lines per beats).

Regarding merging. The more I think about it the more I think I should do this properly:
1. Take all the NO-OP commits and create a new PR with it for upstream/master.
2. Take all the Tracker Editor commits, either create a new PR for a feature branch or keep it on my account branch till at least editing is supported.

The reasons I wanted to make this PR now though are
1. Not accumulate too many commits (there are already over 100)
2. Give a taste of what it is, possibly gather interest amongst enthusiasts that may want to help

So if you think it's better on your end to merge that PR, possibly disabling that code with a `#ifdef` till it becomes more mature, I'm happy with that of course. On the other hand, I do fully understand that you may feel ambivalent about it as it isn't necessarily considered a professional functionality (although I claim this is purely cultural, cause making music with a tracker interface is f. efficient :-), so I'm well ready to milk as much as possible my PRs and wait for when the time will be right.

ngeiswei:
I'm closing. I plan to

1. take the non tracker editor bit out of it, create a new PR with these
2. simplify the git history for the tracker editor part, since no body else is working on it ATM
3. support editing (will be ready after a few weeks to a few months)

x42:
Thanks, that makes sense.

I still would not mind merging (2) as well. if needed it's as simple as `#ifdef` out the entry points in `ardour.menus.in` for releases  and meanwhile collect feedback using nightly builds.

PR #311, opened 2016-12-19
0007159: Check that regions are part of same layer during group selec…

PatrickStephansen:
…tions.

There is probably a better way of achieving this behaviour since this
is my first contribution.

See the Mantis issue with ID 7159 for more detail.


PatrickStephansen:
This has been done much more thoroughly by someone else in v6. The preferences menu contains the option "Regions in active edit groups are edited together:" which can be set to "only if they have identical length, position and layer" to give identical behaviour to what I hard coded here. This PR isn't relevant anymore.

PR #313, opened 2016-12-22
More 8 whitespace to tab indentation fixes

ngeiswei:
I had forgotten these.


x42:
Merged as 5.5-277 6431c4a

For future reference: we generally try to not touch files in `libs/` and `gtk2_ardour/` in a single commit. (if there's inter-dependence commit `libs/*` first, then change the UI.  It's not a big deal with these NO-OP whitespace changes, though.  But some derived projects only use the libs and we'd like to avoid merge conflicts.

ngeiswei:
Got it. Thanks.

PR #314, opened 2016-12-27
Add install rule

strk:


x42:
closing this. `Makefile` was removed.

I'm sorry but it just doesn't cut it to compile a complex piece of software like Ardour by typing `make` and hoping for the best. There are intricate details and one should pay attention to configuration options.

Not to mention `make uninstall` which may not completely remove old versions and leave e.g. old backends in place, causing issues on the next install...

The README points to http://ardour.org/development.html which has built-instructions and https://nightly.ardour.org/list.php has complete build-logs for all variants which can serve as walk-through.

PR #315, opened 2016-12-29

cooltehno:
blue_b_m_corrections_291216
1. The frame of the selected track header is changed from Red to Violet color. It's more  harmoniously with general design. (gtk_bright_indicator - color 23)
2. The fill of the selected track header is changed to Violet color. It was very similar to an unselected midi track header. Also more  harmoniously with general design. (gtk_track_header_selected - color 23)
3. The frame of the selected midi region is changed from Brown to Blue color. This is more noticeable & harmoniously with a general design. (selected time axis frame - color 85)
4. The GTK text is changed to darker – more readable. (gtk_texts – color 27)
5. The text of the general menu, names of tracks is changed to darker – more readable. (gtk_foreground: color 93: from #51578A to #2F3880)
6. The fill of the selected BUS-track header is changed to lighter. It’s more close to “light-theme-conception” and makes more readable the name track text. (gtk_audio_bus – color 4)
7. The transparency of the midi region in “E”-mouse mode is little reduced. This parameter has the same value to all cooltehno’s themes now. (editable region – alpha:0.14)

### DIFF
--- a/gtk2_ardour/themes/blueberry_milk-ardour.colors
+++ b/gtk2_ardour/themes/blueberry_milk-ardour.colors
@@ -97,7 +97,7 @@
     <Color name="color 90" value="0x222020ff"/>
     <Color name="color 91" value="0x8ef823ff"/>
     <Color name="color 92" value="0xcfdee4ff"/>
-    <Color name="color 93" value="0x51578aff"/>
+    <Color name="color 93" value="0x2F3880ff"/>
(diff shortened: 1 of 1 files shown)


cooltehno:
Big thanks, Paul, as usual :) Special thanks for the light area:
light_area_thanks
A little problem I've noticed, and made a bugtracker post:
http://tracker.ardour.org/view.php?id=7197

PR #316, opened 2016-12-29
Update caineville-ardour.colors

cooltehno:
caineville_corrections_291216
1. The punch rectangle is changed from Yellow to Orange color. (transport punch rect - color 9)
2. The transparency of the midi region in “E”-mouse mode is little reduced. This parameter has the same value to all cooltehno’s themes now. (editable region – alpha:0.14)


pauldavisthefirst:
manually rebased and merged.

PR #319, opened 2017-01-07
Formula fix for slider position (new gain math)

gbevin:
Was reading through the source code and looked into reusing slider_position_to_gain and gain_to_slider_position for another open source project. I noticed the comment "/* XXX Marcus writes: this doesn't seem right to me. but i don't have a better answer ... */" and wrote the proper inverse formula instead of the one that's in the sources. Maybe this helps.

### DIFF
--- a/libs/ardour/ardour/utils.h
+++ b/libs/ardour/ardour/utils.h
@@ -144,9 +144,8 @@ slider_position_to_gain (double pos)
 
 	return g;
 #else
-	/* XXX Marcus writes: this doesn't seem right to me. but i don't have a better answer ... */
 	if (pos == 0.0) return 0;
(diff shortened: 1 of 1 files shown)


x42:
It'd be interesting to profile this, you're basically trading 3 sqrt() calls with exp() and log() which are more expensive.

Some food for thought: http://lists.ardour.org/pipermail/ardour-dev-ardour.org/2016-May/014037.html

gbevin:
It's definitely more expensive, that's for sure, but it's also correct while the other one supposedly isn't (per in-code comment, and my testing). For my purpose in the other app, the extra CPU hit for this is not a problem, it only gets called when a user interacts manually with the GUI.

x42:
Merged as part of the rework in e91b80eb85

PR #320, opened 2017-01-20
Update metronome tooltip on gain change

dsheeler:


x42:
Rebased as Ardour 5.5-555 e315b25eb - Thank you!

PR #321, opened 2017-01-21
Add note name to number converter

ngeiswei:
Useful for the midi list editor (in a forthcoming PR).

### DIFF
--- a/libs/ardour/ardour/parameter_descriptor.h
+++ b/libs/ardour/ardour/parameter_descriptor.h
@@ -45,7 +45,8 @@ struct LIBARDOUR_API ParameterDescriptor : public Evoral::ParameterDescriptor
 
 	static std::string midi_note_name (uint8_t);
 
-	/** Dual of midi_note_name, convert a note name intoa  midi note number. */
+	/** Dual of midi_note_name, convert a note name into its midi note number. */
+	static std::map<std::string, uint8_t> build_midi_name2num();
 	static uint8_t midi_note_num (const std::string& name);
 
 	ParameterDescriptor(const Evoral::Parameter& parameter);
--- a/libs/ardour/ardour/parameter_descriptor.h
+++ b/libs/ardour/ardour/parameter_descriptor.h
@@ -20,6 +20,8 @@
 #ifndef __ardour_parameter_descriptor_h__
 #define __ardour_parameter_descriptor_h__
 
+#include <boost/algorithm/string/compare.hpp>
+
 #include "ardour/types.h"
 #include "ardour/variant.h"
 
@@ -46,7 +48,8 @@ struct LIBARDOUR_API ParameterDescriptor : public Evoral::ParameterDescriptor
 	static std::string midi_note_name (uint8_t, bool translate=true);
 
 	/** Dual of midi_note_name, convert a note name into its midi note number. */
-	static std::map<std::string, uint8_t> build_midi_name2num();
+	typedef std::map<std::string, uint8_t, boost::algorithm::is_iequal> NameNumMap;
+	static NameNumMap build_midi_name2num();
 	static uint8_t midi_note_num (const std::string& name);
 
 	ParameterDescriptor(const Evoral::Parameter& parameter);
--- a/libs/ardour/ardour/parameter_descriptor.h
+++ b/libs/ardour/ardour/parameter_descriptor.h
@@ -20,7 +20,7 @@
 #ifndef __ardour_parameter_descriptor_h__
 #define __ardour_parameter_descriptor_h__
 
-#include <boost/algorithm/string/compare.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 
 #include "ardour/types.h"
 #include "ardour/variant.h"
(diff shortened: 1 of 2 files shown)


x42:
What is the eventual use-case for this function? Does it allow user-entry?

Note that ParameterDescriptor::midi_note_name() is translatable. I suppose you should compare against the gettext string and UTF-8  e.g.  Do, Do ♯, Ré, Ré ♯ ... for French.

ngeiswei:
The use case is #322 (can edit the note name in the midi list editor, not just the note number).

As for the translation, I wonder who wants that, but if you think it's important I can have it support it as well.

x42:
German users want "H" instead of "B" and some French composers and musicians asked for solfege last year: mainly for the verbose cursor in the editor and plugin controls.

I don't know if it make sense for  a tracker interface on a grid, but display and entry should be consistent.

If the list editor displays "Do" don't expect a user to enter "C". Then there's Japanese ハ, too (though the Ardour translation has not caught up there, yet)

Maybe the exiting method needs `midi_note_name(bool translate)` and prefill the entry?

ngeiswei:
Fair enough. I think you are right about offering optional midi note translation (i.e. when the language setting is different than English). And yes we probably want pre-calculate a bijective map:
```
note name <-> note number
```
rather than hardcode an English note name parser.

I'll study how automatic translation works in Ardour and get back.

trebmuh:
> As for the translation, I wonder who wants that, but if you think it's important I can have it support it as well.

I do. And basically, pretty much most of the musician from the latin world would are not using the A, B, C ... notation.

ngeiswei:
Don't merge just yet, it needs more testing...

ngeiswei:
It's annoying I can't test it for other languages than English. I can change the OS language, all my programs seem to be translated but ardour isn't. I think it should work of course but I can't test...

pauldavisthefirst:
you probably need to use ./waf configure with the --nls option

ngeiswei:
Does not work either. I should note that it works on the pre-compiled version of Ardour provided by Ubuntu Studio 16.10 (although the sharp symbols are displayed as ?? - for the French language). Maybe I'm missing some library, although looking at the output of `./waf configure --nls` doesn't reveal anything obvious.

ngeiswei:
Oh, I had the good idea to install instead of running `ardev` and it works! Well the i18n works, my code actually has some problems, so don't merge just yet...

ngeiswei:
Actually, I'm observing the same problem on the master, being that instead of having "Sol" being display (the French translation of "G"), I see "Note|Sol". If I switch back to English, then it properly displays "G" with the "Note|" prepended to it.

ngeiswei:
I'm no i18n expert but I suspect it is a translation error. See for instance
https://github.com/Ardour/ardour/blob/master/libs/ardour/po/fr.po#L1679
https://github.com/Ardour/ardour/blob/master/libs/ardour/po/de.po#L1697
you may observe that the context `Note|` is still present in the French translation but missing in the German one.

ngeiswei:
Just to be clear, don't merge just yet, there's another issue beside i18n.

ngeiswei:
AFAIK, you can merge. The i18n is a different issue, IMO.

#322 has been tested as well can be merged right after, AFAIK.

ngeiswei:
Any news on this? I'm gonna need it soon.

x42:
I'll need to double check `boost::algorithm::ilexicographical_compare` availability, but this looks fine otherwise.

pauldavisthefirst:
can't see a reason why, given that we know the note name is a latin alpha, strcasecmp() isn't adequate.

what am i missing?

ngeiswei:
Because if `translate` is set to true (the default) in `ParameterDescriptor::midi_note_name (const uint8_t b, bool translate)` then the keys in `name2num` are translated, so they could be anything, not just latin.

ngeiswei:
And of course `boost::algorithm::ilexicographical_compare` is case insensitive. So to sum up, the reasons are

1. Supports i18n
2. Is case insensitive

ngeiswei:
Anything preventing this PR to be merged?

x42:
Why does the map have to be lexically sorted at all?

ngeiswei:
It doesn't have to be *lexically* sorted, but it has to be sorted in a way that is case insensitive, which `boost::algorithm::ilexicographical_compare` offers. The reason is has to be sorted at all, lexically or not, is so the note name can be mapped to the note number, see `ParameterDescriptor::build_midi_name2num` and `ParameterDescriptor::midi_note_num`.

x42:
I must still be missing something. I don't see why sorting is needed. Both note numbers and note names are unique and the map is isomorph.   `std::map<std::string, uint8_t>`  should suffice.

ngeiswei:
Case-insensitive allows the user to enter "f4" and be recognized as "F4". For having used it, I can tell that it is really convenient.

ngeiswei:
Now if you don't want to depend on `boost::algorithm::ilexicographical_compare` I can certainly rewrite the code to do without it, it's more elegant that way.

ngeiswei:
Well, I can always reimplement `is_ilexicographical_compare` :-) ;-)

x42:
So `midi_note_num (const std::string& name)` should convert the string to lower-case before lookup, shouldn't it?

just `boost::to_lower()` when building the map and `name2num.find(boost::to_lower(name))`

ngeiswei:
If `is_ilexicographical_compare` is used there is no need to convert the string to lower-case before look-up, the look-up itself will be case-insensitive.

Do you want me to recode it as to not rely on `boost::algorithm::ilexicographical_compare`?

ngeiswei:
Anyway, you did convince me, using `boost::to_lower()` when building the map and performing the look-up is simpler than using `is_ilexicographical_compare`, certainly in terms of LOC, so I'll go with that.

ngeiswei:
Closing, will squash into a single commit.

PR #322, opened 2017-01-21
Support note name editing in midi list editor

ngeiswei:
Need to merge #321 before this one.


x42:
Rebased as Ardour-5.8-744 8242f4f63

PR #323, opened 2017-01-23
Meter point toggle button -> popup menu

dsheeler:


x42:
Looks good, except the `MeterPointChangeTarget` should not be in libardour (it's GUI only, not part of the libardour API), and instead in `gain_meter.h`.

Should I rebase and fix it, or can you do that?

dsheeler:
I'll fix the MeterPointChangeTarget header situation.  Can I fix the merge-commit situation in this pull-request?  Do I need to create a new one?

PR #324, opened 2017-01-23
Meter point toggle button now uses a popup menu instead of toggling

dsheeler:


x42:
Rebased as Ardour 5.5-582 33c74357f  Thanks a bunch!

PR #326, opened 2017-02-17
PT Import:  MIDI support

zamaudio:
This provides MIDI import support.

However, currently we still don't know how to parse the start offset of MIDI regions,
so I have made the import dump all the midi regions at offset zero on a single new MIDI track.


x42:
Rebased as Ardour 5.6-53 4f4400b76. Cheers!

PR #327, opened 2017-02-26
Updated Swedish language

Umeaboy:
Signed-off-by: Kristoffer Grundström <hamnisdude@gmail.com>

### DIFF
--- a/gtk2_ardour/po/sv.po
+++ b/gtk2_ardour/po/sv.po
@@ -19,283 +19,287 @@ msgstr ""
 
 #: about.cc:127
 msgid "Fons Adriaensen"
-msgstr ""
+msgstr "Fons Adriaensen"
 
 #: about.cc:128
 msgid "Brian Ahr"
-msgstr ""
+msgstr "Brian Ahr"
 
 #: about.cc:129
 msgid "John Anderson"
-msgstr ""
+msgstr "John Anderson"
 
 #: about.cc:130
 msgid "Marcus Andersson"
-msgstr ""
+msgstr "Marcus Andersson"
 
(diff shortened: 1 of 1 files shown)


Umeaboy:
Is there anyone that can merge this? Just curious.

mojofunk:
Thanks for this translation update. Yes there there are several people that can merge PR's and do so regularly. How long it may take is variable and dependent on many factors but you can expect further feedback in a within a few days.

jplloyd:
Some comments from a native speaker:

When things can translate to multiple different words, be sure that the terms you use are the correct ones, both semantically and contextually. Same thing applies to the reverse scenario, if a term you are translating can have multiple different meanings. Here "Config" is translated to (one ambiguous way of saying) "Configure", whereas in the context of the string it actually means "Configuration / Setup".

Your translations of "Range" uses the term for effective range of e.g. transceivers, like a cellphone tower, rather than a frequency/time range. The term you replaced, "omfång", is the correct one.

For "Insert", I would use "Infoga" instead of "Klistra in" (which instead is the standard translation of "Paste").

"Ambigiös" is not a word, and "utmatning" should probably be replaced by "utsignal".

The rest looks fine.

pauldavisthefirst:
Any plans to update this PR or respond to the comments? I'd love to merge this before 5.11 ....

Umeaboy:
I'll look into it very soon. Hold on. :)

Umeaboy:
I made the appropriate changes. I don't think I've missed anything. Who can merge?

pauldavisthefirst:
manually rebased as 1d0221349. thanks!

PR #328, opened 2017-03-04
Refine Beats::operator== down to half a tick

ngeiswei:
This addresses an issue that rises when changing a midi note start time or length to +/- one tick (I wanted to create an issue in the bug tracker but I'm still waiting for receiving the email to complete the process of creating my account).

To reproduce it: left-click on a midi note, open the note edit dialog and change the time or length by + or - one tick, then close, reopen on that same note, you'll notice that the tick hasn't changed. If it has changed, then you're lucky, try again one another time or length and you'll see no change.

I initially tried to fix that by replacing `<=` by `<`, so specifically
```
		return fabs(_time - b._time) <= (1.0 / PPQN);
```
by
```
		return fabs(_time - b._time) < (1.0 / PPQN);
```
on all methods concerned but it didn't work (probably due to floating point inaccuracies, still weird cause it's double, but I don't have another explanation). So instead I replaced `1.0` by `0.5` and it works. I'm pretty sure replacing by say `0.95` would work just as well, but `0.5` rounds to half a tick and is cooler in my own mind. Feel free to change the value to anything as long it fixes the issue. I see no reason why 0.5 wouldn't be fine though.

P.S: you may need to replace in file `gtk2_ardour/audio_clock.cc` the following by
```
	if (_mode == MinSec)
		tmp->set_text (" 88:88:88,888 ");
	else
		tmp->set_text (" 88:88:88,88 ");
```
by
```
	tmp->set_text (" 88:88:88,888 ");
```
in order to see what you're editing in the note edit dialog. I wanted to create a PR for that but @x42 told be the issue might be more subtle than I think so I prefer to let to someone else.

Oh, something else, the first commit fixes the problem. The second commit improves the code of the various operators by reusing others when possible. I think it's better that way but if you don't like it for any reason feel free to cherry pick only the first commit, as the second one is not essential for the fix.

### DIFF
--- a/libs/evoral/evoral/Beats.hpp
+++ b/libs/evoral/evoral/Beats.hpp
@@ -77,18 +77,18 @@ public:
 	}
 
 	inline bool operator==(const Beats& b) const {
-		/* Acceptable tolerance is 1 tick. */
-		return fabs(_time - b._time) <= (1.0 / PPQN);
+		/* Acceptable tolerance is 1/2 tick. */
+		return fabs(_time - b._time) <= (0.5 / PPQN);
 	}
 
 	inline bool operator==(double t) const {
-		/* Acceptable tolerance is 1 tick. */
-		return fabs(_time - t) <= (1.0 / PPQN);
+		/* Acceptable tolerance is 1/2 tick. */
+		return fabs(_time - t) <= (0.5 / PPQN);
 	}
 
 	inline bool operator==(int beats) const {
-		/* Acceptable tolerance is 1 tick. */
-		return fabs(_time - beats) <= (1.0 / PPQN);
+		/* Acceptable tolerance is 1/2 tick. */
+		return fabs(_time - beats) <= (0.5 / PPQN);
 	}
 
 	inline bool operator!=(const Beats& b) const {
@@ -96,8 +96,8 @@ public:
 	}
 
 	inline bool operator<(const Beats& b) const {
-		/* Acceptable tolerance is 1 tick. */
-		if (fabs(_time - b._time) <= (1.0 / PPQN)) {
+		/* Acceptable tolerance is 1/2 tick. */
+		if (fabs(_time - b._time) <= (0.5 / PPQN)) {
 			return false;  /* Effectively identical. */
 		} else {
 			return _time < b._time;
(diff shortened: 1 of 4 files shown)


mojofunk:
If this is indeed a bug in the Evoral::Beats class then I would like to see a unit test first added to libs/evoral/tests before your patch/changes are applied so we can verify that it does fix the issue in isolation and hopefully ensure that it does not get broken in the future.

ngeiswei:
Excellent suggestion, I'll add an unit test.

ngeiswei:
I've added a unit test. I tests 1000 value pairs, it doesn't need that many values (10 is enough), just to be cautious. When replacing 0.5 by 1.0 the unit test fails so it really captures the bug.

x42:
Check out the "ticktime" branch, it's work in progress to  replace the double with an integer (tick count) and all these rounding issues will be gone.

ngeiswei:
I see. So you recommend to still merge this one as a temporary bandage, or close and wait for the ticktime branch to be completed?

pauldavisthefirst:
We are going to introduce a "superclock" for 6.0 that will remove rounding issues. I think this PR should probably be closed.

ngeiswei:
Interesting, what is the "superclock" branch. For what I know the ticktime branch fixes this issue.

pauldavisthefirst:
the branch doesn't exist yet. all time will be based on a clock with resolution on the order of 2^-36, with integer factors corresponding to all sample rates and all musical time divisions.

ngeiswei:
I see, it looks like it's gonna take a while to be implemented thought.

If ticktime never gets merged, I would personally prefer if this one is merged. I'm reasonably confident that it creates no lateral issues (though I'm not 100% sure either).

x42:
Is there an effective difference compared to simply doubling PPQN? 

As for changing the operators, my gut tells me that this'd need triple checking: Some of them are specific to retain strict weak ordering when using Beats in STL.

just my 2c

ngeiswei:
Yep, doubling PPQN has the same effect!

In that case maybe I can recreate a PR doubling PPQN, keeping the unit test, but not the operator factorizations?

ngeiswei:
Based on @pauldavisthefirst 's remark in https://tracker.ardour.org/view.php?id=7947#c21088 and the fact the unit test of this PR pass without the need of that change, I believe this issue can be safely closed.

PR #329, opened 2017-03-06
Details

7890:
(see commit messages)


gtk2_ardour/time_axis_view.h
```diff
@@ -240,6 +240,7 @@ class TimeAxisView : public virtual AxisView
 	std::string            controls_base_unselected_name;
 	std::string            controls_base_selected_name;
 	Gtk::Menu*             display_menu; /* The standard LHS Track control popup-menus */
+	Gtk::MenuItem*         rename_menu_item;
```
x42: Why does the class need a reference to the MenuItem?

7890: This is analog to mixer_strip.h. It helps if someone needs to get a handle on rename_menu_item from a less local focus.

x42:
seems to be superseded by #330 -- closing

PR #330, opened 2017-03-08
Details #2

7890:
adapting to cruft removal in
c93bef0ab693021f0905420a7e004c67ab7d1fe8


x42:
Rename as context menu in the track-header makes no sense to me. The action is already directly accessible there.

7890:
Many ways lead to Rome. "Rename" is missing in the context menu of the track header.

7890:
Let's discuss this for months

7890:
does anything speak against the tooltips?

PR #334, opened 2017-05-12
Fix compiler warning of unused variable.

chaot4:
As boring as the title sounds... ;)


x42:
Rebased as Ardour 5.8-743 77a2632e3.

PR #335, opened 2017-05-12
Add note name parser to convert to midi note number

ngeiswei:
Supports i18n, is case and whitespace insensitive for more resilent parsing.


ngeiswei:
I tested with French, works as expected, also decided to be whitespace insensitive (the French translation has whitespaces in them, but it's good to have anyway).

x42:
rebased as 5.8-742 40f3efa3a. Thanks!

PR #336, opened 2017-05-13
Fix range split on midi regions (fixes #7277)

chaot4:
Please give this another thought before merging. :) I didn't find any side effects of this change (except, of course, fixing the bug) but there might be something I overlooked.

### DIFF
--- a/libs/ardour/midi_region.cc
+++ b/libs/ardour/midi_region.cc
@@ -242,9 +242,7 @@ MidiRegion::post_set (const PropertyChange& pc)
 void
 MidiRegion::set_start_beats_from_start_frames ()
 {
-	if (position_lock_style() == AudioTime) {
-		_start_beats = quarter_note() - _session.tempo_map().quarter_note_at_frame (_position - _start);
-	}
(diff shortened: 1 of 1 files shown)


nmains:
Thanks for the patch.
I've addressed this issue in the music_timeline branch in a different way.
Similarly to master, it only sets the start_beats property if locked to AudioTime.
The reason is that start_beats may be an exact musical value and resetting it to the
qnote position of a frame may introduce a rounding error.
The difference might be seen by examining the XML file after range split.
In master, this will probably look like 2.99997249845, whereas in music_timeline,
it will be 3.0.
Of course if the regions are trimmed to a start of 0, there will be no obvious difference.
Having said that, the patch fixes the bug, but i suspect that under some conditions
it might re-introduce a missing first note bug.
Probably best to test over tempo changes or ramps to see if this is the case.
It would be great if you could try music_timeline branch to confirm that this is fixed
in that branch.

thanks.

chaot4:
I just checked your music_timeline branch and the bug 7277 seems to be fixed there. Thus, I think it would be best to close this PR and instead have this bug fixed when your branch is merged.

PR #337, opened 2017-05-14
Fix double "Tape" in Record Mode drop down when adding new track.

chaot4:
This bug probably also lead to "Tape" being shown in mixbus even
though it should not.


x42:
Rebased as 5.8-738 092ce6dc0 thanks.

PR #342, opened 2017-05-31
Fix LD_LIBRARY_PATH in GNU/Linux startup script

nickolas360:
Currently, the startup script for GNU/Linux adds the current working directory to ``LD_LIBRARY_PATH`` if ``LD_LIBRARY_PATH`` is not empty or unset.

For example, if ``LD_LIBRARY_PATH`` is set to ``/lib`` when the current script is run, it will be set to ``<install-dir>/lib::/lib``, which includes the current working directory as one of the paths.

This pull request removes the extra colon added to ``LD_LIBRARY_PATH``.

### DIFF
--- a/tools/linux_packaging/ardour.sh.in
+++ b/tools/linux_packaging/ardour.sh.in
@@ -15,9 +15,8 @@ checkdebug(){
 checkdebug "$@"
 
 
-# LD_LIBRARY_PATH needs to be set here so that epa can swap between the original and the bundled version
-# (the original one will be stored in PREBUNDLE_ENV)
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
+# this needs to be set so that we can restore the environment when we want to
+# find JACK (or similar)
 export PREBUNDLE_ENV="$(env)"
 
 BIN_DIR=$(dirname $(readlink -f $0))


pauldavisthefirst:
This removes setting LD_LIBRARY_PATH at all, which seems incorrect in this context

nickolas360:
LD_LIBRARY_PATH is set correctly after the removed line, [in this line](https://github.com/Ardour/ardour/blob/46a2bfa0e61dd16a0514d81087c08ec965c3e595/tools/linux_packaging/ardour.sh.in#L44):
```
export LD_LIBRARY_PATH=$INSTALL_DIR/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
```

The [previous time](https://github.com/Ardour/ardour/blob/46a2bfa0e61dd16a0514d81087c08ec965c3e595/tools/linux_packaging/ardour.sh.in#L20) LD_LIBRARY_PATH is set, a colon is added to the beginning, but this results in two colons being added, which adds the CWD to the library paths.

nickolas360:
Oh, if you're talking about the possibility of ``LD_LIBRARY_PATH`` not being included in ``PREBUNDLE_ENV`` (if ``LD_LIBRARY_PATH`` is unset when the script is called), that shouldn't be a problem because ``EnvironmentalProtectionAgency::restore()`` [calls clear()](https://github.com/Ardour/ardour/blob/46a2bfa0e61dd16a0514d81087c08ec965c3e595/libs/pbd/epa.cc#L124).

So in that case, ``LD_LIBRARY_PATH`` would be correctly unset during ``restore()``.

nickolas360:
Sure, although then that line is the same as the simpler ``export LD_LIBRARY_PATH=$LD_LIBRARY_PATH``.

But I can't seem to find why that line is necessary at all; it plays no role in restoring environment variables, and the real library path is set later in the script.

PR #343, opened 2017-06-04
add route dialog now has "Add" and "Add and Close" buttons.  "Add" ke…

dsheeler:
…eps dialog open so user can add multiple templates.


x42:
Merged as 5.9-60-g2c41df947. Thanks!

PR #344, opened 2017-06-04
Fix a-EQ when parameter changes are very slow

ngeiswei:
If the parameters change too slowly the filter may never get updated. Indeed,
in spite of v_f0, v_g or v_bw being updated, set_params may never be called,
thus v_filter never updated.

Find attached a project showing the problem. You'll see the frequency response graph doesn't get updated for the first 2 lines, cause the rate of change is too low. Of course it's not just the graph, the filter itself isn't updated as well.

[automation-debug.zip](https://github.com/Ardour/ardour/files/1050667/automation-debug.zip)


ngeiswei:
I do feel a bit of concern that v_f0, v_g and v_bw do not end up exactly to their target values though. I believe that probably was the intention of the else-branches that this patch removes, although of course, for reasons I explain, they alone, did not solve that either.

So maybe it's not an ideal fix and more thought is required.

x42:
I think the else branch was to avoid denormals.

x42:
Rebased as Ardour 5.9-61 a07ca8fa3

PR #345, opened 2017-06-10
Playhead update

johannes-mueller:
This adresses the issue described in http://tracker.ardour.org/view.php?id=6887#c19765

Two things:
* `Session::Located` is emitted in `Session::backend_sync_callback()` to make sure there is a Located signal when `AudioEngine::transport_frame()` is up to date.
* Suppress emission of `Session::Located` in `Session::locate()` if `AudioEngine::transport_frame()` is not up to date, to prevent the playhead from jumping back and forth.

I debugged this together with @x42 on IRC and we ended up with the question, why `Session::backend_sync_callback()` is called with `TransportStopped`.

Since then I did some 16 hours of reallife recording and editing work using this branch. No problem occured, the playhead behaves as expected.

### DIFF
--- a/libs/ardour/session_time.cc
+++ b/libs/ardour/session_time.cc
@@ -183,19 +183,26 @@ int
 Session::backend_sync_callback (TransportState state, framepos_t pos)
 {
 	bool slave = synced_to_engine();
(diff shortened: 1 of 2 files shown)


x42:
Rebased and merged as Ardour 5.9-93 05c77db77. Thanks a lot!

PR #346, opened 2017-06-18
Show makeup gain in a-comp's inline rendering

johannes-mueller:
Is there a reason not to do it?

### DIFF
--- a/libs/plugins/a-comp.lv2/a-comp#stereo.ttl.in
+++ b/libs/plugins/a-comp.lv2/a-comp#stereo.ttl.in
@@ -33,7 +33,7 @@
         lv2:symbol "rel" ;
         lv2:default 80.000000 ;
         lv2:minimum 1.000000 ;
-        lv2:maximum 500.000000 ;
+        lv2:maximum 2000.000000 ;
         unit:unit unit:ms ;
     ] ,
     [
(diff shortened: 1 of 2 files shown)


x42:
Quick experiment on top of this patch to show the range -60 .. +10dBFS

https://gist.github.com/x42/030ecdcb697f20489826074832c20ce8

johannes-mueller:
I like your experiment. It's good to have the lines at 0.0 dBFS clearly visible.

x42:
@zamaudio  any opinions? reasons not to merge this?

zamaudio:
Perhaps the makeup should not affect the graph when the plugin is bypassed.  Otherwise LGTM.

johannes-mueller:
@zamaudio yeah, I missed that in the stereo version. Fixed in 265a879.

johannes-mueller:
In the meantime it's more than just the makeup gain thing. It also adresses #7120.

If the channel strip is wide enough (≥100 px) we show the graph with a gain reduction meter bar overlay. If the width is <100px we switch to a view where only a output level and a gain reduction meter bar is shown.

In this meter bar mode we take the whole height for the inline display, that we are granted, which might be a little too much, but I think as for now, we can keep it.

x42:
In narrow mode the gain-meters increases the height requirement. IIUC #7120 suggested a horizontal meter (about the same size as an inline-control)  to save vertical space.

johannes-mueller:
Right. But #7120 raises more fundamental questions about the a-plugins. I will write to the dev mailing list when I'm back from my weekend trip.

johannes-mueller:
Closed in favor of [PR 365](https://github.com/Ardour/ardour/pull/365)

PR #348, opened 2017-06-19
Update clear_gray-ardour.colors

cooltehno:
clear_gray_correction_180617
1. The loop rectangle is changed to a gray color. It adds more conformity to a gray theme design (transport loop rect - color 104).
2. The piano roll stripes are slightly changed to lighter colors. It provides more difference between looped and non-looped areas (piano roll black – color 81; piano roll white – color 48).
3. The automation track header is changed from light to dark. It gives more comfortable reading the track name (gtk_automation_track_header – color 25).
4. The audio track header is slightly changed to a lighter color.  It provides more color space for the dark automation track header (gtk_audio_track – color 87).


pauldavisthefirst:
manually rebased. thanks!

PR #350, opened 2017-06-23
Add missing `echo` in ardev_common.sh.in

nickolas360:
The error message at the top of ardev_common.sh.in is missing a call to ``echo``, so Bash tries to run the error message text as a command.


x42:
rebased & merged as Ardour 5.10-115  2106ab431.

Thanks!  What a stupid oversight that was.

PR #351, opened 2017-06-25
Addressing #7371: don't trigger session->undo() from non-GUI-thread

johannes-mueller:
When triggering Session::undo() or Session::redo() from a
non-GUI-thread (e.g. from a surface protocol) Ardour crashes if setting a
CairoWidget dirty due to a ENSURE_GUI_THREAD assertion. (see #7371)

By triggering undo by BasicUI::access_action() rather than by Session::undo()
we ensure that the GUI thread will finally call Session::undo().

So more like a workaround ... but better than crashing :)

### DIFF
--- a/libs/surfaces/control_protocol/basic_ui.cc
+++ b/libs/surfaces/control_protocol/basic_ui.cc
@@ -334,7 +334,7 @@ BasicUI::undo ()
 void
 BasicUI::redo ()
 {
-	access_action ("Editor/redor");
+	access_action ("Editor/redo");
 }
 
 void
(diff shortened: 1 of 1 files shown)


x42:
I've merged the 2 commits and rebased it.. Ardour 5.10-163 f7452c2a3
Thanks.

johannes-mueller:
Thx.

One thing that came into my mind about that: Is it likely that the action paths will change in the future? In that case undo/redo from surfaces will silently stop working.

pauldavisthefirst:
when and if they change, it will be a major effort, not a random, trivial
change. the entire codebase will need auditing for the use of actions.
back-compatibility with existing user-created mappings will be a
particularly thorny area.

PR #352, opened 2017-06-28
correctly display button menu entries

7890:
'_' in a label seems to be interpreted as "next char is mnemonic" thus route names will be garbled if their name contains a '_'.
add missing replace_all() for r->name().

### DIFF
--- a/gtk2_ardour/patch_change_dialog.cc
+++ b/gtk2_ardour/patch_change_dialog.cc
@@ -34,6 +34,7 @@
 #include "gui_thread.h"
 
 #include "pbd/i18n.h"
+#include "pbd/replace_all.h"
 
 using namespace std;
 using namespace Gtk;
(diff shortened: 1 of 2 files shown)


x42:
Don't modify the string, use the proper API instead (create managed a MenuItem with `mnemonic=false` pass it to the MenuElem c'tor)

https://developer.gnome.org/gtkmm/2.24/classGtk_1_1MenuItem.html
https://developer.gnome.org/gtkmm/2.24/classGtk_1_1Menu__Helpers_1_1MenuElem.html

7890:
it's using the same technique as seen in other spots (to prevent said bug)

x42:
The _other spots_ are wrong too.  There is a proper API for this, so no workarounds are needed to solve this.

There are a few cases where we do want to replace underscores with spaces for display purposes,
but route-names are in general printed as-is (incl underscores).

x42:
Please see 972e01f9c for a proper solution.

7890:
is it up to me to close this PR? you've solved the issue.

PR #353, opened 2017-06-29
correctly display region name in region context menu

7890:


7890:
closed due to superfluous include

PR #354, opened 2017-06-29
correctly display region name in region context menu

7890:


7890:
PS: even if the replace "_" with "__" hack seems reasonable to escape "_", it won't do that (display in menu is "__").
you're safe to merge.

x42:
Applied in Ardour 5.10-179 79384339e
Thanks!

PR #355, opened 2017-06-30
Adhere more to code style guide

7890:
-Use tab for indentation
-Use curly braces for if / while

### DIFF
--- a/gtk2_ardour/window_manager.h
+++ b/gtk2_ardour/window_manager.h
@@ -66,7 +66,7 @@ class Manager : public ARDOUR::SessionHandlePtr
 	void set_transient_for (Gtk::Window*);
 	Gtk::Window* transient_parent() const { return current_transient_parent; }
 
-                                        private:
+  private:
 	typedef std::list<ProxyBase*> Windows;
 	Windows _windows;
 	Glib::RefPtr<Gtk::ActionGroup> window_actions;
@@ -155,7 +155,7 @@ class ProxyWithConstructor: public ProxyBase
 		}
 	}
 
-                                        private:
+  private:
 	boost::function<T*()>	creator;
 };
 


x42:
Wow, that's nice! 

It'd be great to consolidate those into 2 commits (`libs/`  `gtk2_ardour/`) don't clutter up the history and lengthen `git bisect` runs for  whitespace NO-OP changes.

7890:
i'd like the moon to be out of cheese, it would be great!

pauldavisthefirst:
if you don't want to do that, then tell us the workflow you used to
generate this commit and we'll do it ourselves, without snark.

snark on IRC is great, snark on GH's PR system is just irritating.

x42:
It cannot be merged anyway, there is no public remote repository. Github says "patch is too large and not available).

7890:
please see #357, it's now all in one commit. does that help?

x42:
Yes that does help, see discussion there.  As for the libs, a nice way to solve this would be to use *this* branch as-is.   `git reset`  (no --hard, just the index) to the last upstrem version.  `git add libs/ardour libs/pbd libs/gtkmm2ext`  and commit.

Note that 3rd party libs (vamp, qm-dsp, libltc etc  are not covered by ardour's coding-style) we try to stay close to upstream for those.

7890:
Paul, the workflow was to grep for multiple spaces and open each file, go through it and decide whether or not the spaces are valid (to align stuff) or in place where a tab was expected. I could have missed some, i'd say roughly 95% of obvious tab cases are catched. the alignment of multi-line statements needs a second round, i've touched some but not systematically (i.e. tab up to indentation of statement, then use space to align to method ()). it's a start..

7890:
x42, i've not touched any other files than in /gtk2_ardour/*.
it needs a separate round for /libs/*

x42:
My bad, for some reason I though some of those "diff to large" referenced lib/*
Anyway I'll close this #357  supersedes this PR. Thanks!

PR #356, opened 2017-06-30
fix typos

7890:


7890:
Dictionaries suggest that "Busses" is incorrect plural of "Bus". It's "Buses".
Any opinions on changing it to "Buses" globally?

gtk2_ardour/generic_pluginui.cc
```diff
@@ -565,7 +565,7 @@ GenericPluginUI::ControlUI::ControlUI (const Evoral::Parameter& p)
 	   below). be sure to include a descender.
 	*/
 
-	automate_button.set_sizing_text(_("Mgnual"));
```
x42: This typo is intentional. read the comment above.

x42:
In the pro-audio world "bus", "busses"  is most common in ardour's default locale (en_US). 
In the UK "buss", "Busses" for consoles.

gtk2_ardour/generic_pluginui.cc
7890: hm.. i don't understand why it needs a typo. could it be "the widest string you will put" instead?

7890:
yeah, so Busses, that's why i didn't touch any of these.

x42:
PS. Why 3 commits for 3 typos? In general prefer to group commits to to lib/ and gtk2_ardour/.

For no-op cases, there's no good reason to separate commits; quite the opposite: It only lengthens `git-bisect` runs and clutters up history.

7890:
it's for you so you can cherry-pick

pauldavisthefirst:
On Fri, Jun 30, 2017 at 7:10 PM, 7890 <notifications@github.com> wrote:

> *@7890* commented on this pull request.
> ------------------------------
>
> In gtk2_ardour/generic_pluginui.cc
> <https://github.com/Ardour/ardour/pull/356#discussion_r125145006>:
>
> > @@ -565,7 +565,7 @@ GenericPluginUI::ControlUI::ControlUI (const Evoral::Parameter& p)
>  	   below). be sure to include a descender.
>  	*/
>
> -	automate_button.set_sizing_text(_("Mgnual"));
>
> hm.. i don't understand why it needs a typo. could it be "the widest
> string you will put" instead?
>

the comment directly above it indicates the reason. the text is used for
widget sizing and needs to include a descender to get sizing
correct/appropriate/symmetrical.

7890:
understood now, thanks. i was ignorant of that text but it says that at least one letter should use the vertical space below the baseline to give a hint about the maximal height (of a label or so). clearly "Manual" would not do that.

trebmuh:
I've been hit by the same a few months ago :)

@pauldavisthefirst @x42 : couldn't the text be "Mpfdhl" (or something like that) instead of "Mgnual" then people aren't tempted to think it's a typo?

7890:
the tricky part could be to promote that into the translation tools (po files) so that translators know what's expected to fill in. (?)

7890:
effectively i run aspell against the en po, then stumbled upon Mgnual ;)

PR #358, opened 2017-07-02
Select newly created region after separate

7890:
current behaviour:
using a time selection on a region: from context menu "Separate":
-this will leave a time selection over the new region
-> there is no way to move or grab the region since its covered by the ts
-> user must click anywhere to unselect
-> there is no obvious case where keeping the initial selection makes sense as is after the op

new behaviour:
-the newly created region is selected after the op (ts cleared)

### DIFF
--- a/gtk2_ardour/editor_ops.cc
+++ b/gtk2_ardour/editor_ops.cc
@@ -4069,6 +4069,16 @@ Editor::bounce_range_selection (bool replace, bool enable_processing)
 
 	if (in_command) {
 		commit_reversible_command ();
+
+		if(replace)
+		{
+			select_all_selectables_using_time_selection();
+			if(!Config->get_automation_follows_regions())
+			{
+				selection->clear_points ();
+			}
+		        selection->clear_time ();
+		}
 	}
 }
 
--- a/gtk2_ardour/editor_ops.cc
+++ b/gtk2_ardour/editor_ops.cc
@@ -3190,6 +3190,10 @@ Editor::separate_region_from_selection ()
 
 		separate_regions_between (selection->time);
 		select_all_selectables_using_time_selection();
+		if(!Config->get_automation_follows_regions())
+		{
+			selection->clear_points ();
+		}
 		selection->clear_time ();
 
 	} else {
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
You don't want ::select_all_selectables() to select regions. That method
selects anything and everything.

7890:
The "Separation" is called on the context of a time selection. there can't be a track or other selection at this time. using select_all_selectables_using_time_selection(); should work in this case (does here). it basically converts the time to a region selection which is possible after the cut.

7890:
out of the new regions, only the one that was covered by time selection is of interest and will be selected. that makes "Separate" different from "Split" (which again has preferences on what should be selected).

7890:
just try it out and let me know if anything breaks, maybe i've overlooked something

7890:
Paul, "Select All in Range" from region context menu is using the same function:
```
editor.cc:	edit_items.push_back (MenuElem (_("Select All in Range"), sigc::mem_fun(*this, &Editor::select_all_selectables_using_time_selection)));
```
effectively PR just uses this same method after cutting along the time selection start and end to select the region that was covered by time selection. i can't see anything else selected other than the expected region and it's related things like automation points.

7890:
2017-07-03_ardour_separate
(before "Separate", after "Separate")

pauldavisthefirst:
that was my point: it selects thing other than a region.

7890:
it selects the same as if "Select All in Range" would have been chosen after todays "Separate". this is fine. i don't see your point. what would you expect?

7890:
related: there is the preference "move relevant automation when audio regions are moved".

i propose to rename it to "tie automation data to region".

this option would define if automation points will be automatically selected when a region is selected or not. the selection should at all times be visually represented (vs. today's implicit/invisible selection based on preference for move).

the "tie automation data to region" preference would apply for all operations on the region, namely for move, copy and paste. manually changing the automation points selection that's created based on preference would still be possible.

today copy won't consider automation points unless explicitly selected. however there should be a relation to how move is handled. having a common preference "tie automation data to region" seems to be a possible solution.

PS: having a separate preference to handle automation data for copy like there is for move will make it impossible to visually represent the selection. it will disallow to select automation points as soon as the region is selected (because we don't know yet if the region will possibly be moved or copied). 

later on "tie automation data to region" could be a per-track setting that's easily accessed via track header and/or a toggle button.

any thoughts?

7890:
Paul, please see the new commits. it's possible to get exclusive region selection now, depending on preference.

PR #359, opened 2017-07-02
Don't try to chop of the file name extension from a template dir

johannes-mueller:
Template files reside in
`.config/ardour5/templates/$(template_name)/$(template_name).template`

We run through `.config/ardour5/templates/` and find there the names of the
directories the .template-files are located in. These directory names don't
have a .template extension. So we shouldn't try to chop the non existing
extension of, because then we only modify template names with a '.' in them.

### DIFF
--- a/libs/ardour/template_utils.cc
+++ b/libs/ardour/template_utils.cc
@@ -22,7 +22,6 @@
 
 #include <glibmm.h>
 
-#include "pbd/basename.h"
 #include "pbd/file_utils.h"
 #include "pbd/stl_delete.h"
 #include "pbd/xml++.h"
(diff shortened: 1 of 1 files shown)


johannes-mueller:
This does not work for route templates as their template files are not stored in their own template specific directories.

Will reopen, when a solution is found.

x42:
Merged as part of #360

pauldavisthefirst:
One small comment on this. Not saying it should be reverted or anything
like that, but  ... in all new code, we should avoid this construction:

      Dialog d;
      ....
      int response = d.run();
      ...

this use of modal dialogs creates potential for many problems. newer GTK
versions have more or less removed it. the replacement is more complex -
you need a response handler method/function - but the design is much more
robust across platforms and contexts.

   foo () { Dialog* d = new Dialog; d->signal_response().connect (bind
(handler, d)); /* set up d */ d->show (); }
   handler (int response, Dialog* d) { /* handle response */;
delete_when_idle (d); }

PR #360, opened 2017-07-05
Template manager – a simple dialog to manage templates.

johannes-mueller:
Simple dialog to manage i.e. rename and remove templates.

So far, one can tidy up one's templates only by file management moving or renaming files and directories in `.config/ardour5/*templates`. This can be slightly cumbersome moreover as not all users might feel comfortable fiddling around with files under the `.config/ardour5/` tree while ardour is running. So it would be nice to have a simple dialog to at least rename and remove session as well as route templates.

This draft needs at least some polishing, nevertheless I already open a PR hoping for feedback and inspiring discussion.

One issue might be, that there's no signaling when the user does something to the templates. So if there's for example an `AddRouteDialog` instance active while the, user with `TemplateDialog` is renaming and removing templates, the `AddRouteDialog` instance won't notice the changes. Finally it will offer the user the same templates as before the changes the user made in the meantime which might no longer be available. No catastrophe but a line in the error logs.

I appreciate any kind of feedback.

Note that this PR includes the commit of [PR 359](https://github.com/Ardour/ardour/pull/359), which IMHO should be merged directly.

### DIFF
--- a/gtk2_ardour/ardour-sae.menus
+++ b/gtk2_ardour/ardour-sae.menus
@@ -10,6 +10,7 @@
            <menuitem action='Save'/>
            <menuitem action='Snapshot'/>
            <menuitem action='SaveTemplate'/>
+           <menuitem action='ManageTemplates'/>
            <separator/>
 
 	   <menu name='Files' action='Files'>
@@ -56,11 +57,11 @@
                <menuitem action='TransitionToRoll'/>
                <menuitem action='TransitionToReverse'/>
 	       <separator/>
-              
+
 	       <menuitem action='set-playhead'/>
 
 	       <menu action="MovePlayHeadMenu">
- 	           
+
                    <menuitem action='playhead-to-edit'/>
   	           <menuitem action='center-playhead'/>
 	           <separator/>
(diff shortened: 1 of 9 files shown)


x42:
Keep in mind that plugin-state is saved with both track and session-templates.

Seleting session-templates needs to do a `rm -rf` (see `PBD::remove_directory()`) and route-templates need to also delete a sub-folder names like the route-template (without `.template` suffix).

Renaming is trickier, because templates use an absolute path inside the state. For example save a route-template with a-fluidsynth or convo.lv2   in both cases (session-template and route-template) there's a  ` <lv2 ... template-dir="/absolute/path/for/this/template/plugin-state-dir/" .. >`

Easiest way to solve this might be to load the XML `XMLTree tree (old_template);  /* iterate search/replace */ tree.write (renamed_state_name); `

The alternative might be to patch `LV2Plugin::set_state()`, in some smart way,  but the LV2Plugin instance has no reference to template folder (and no context if it's a session-load or template ...) which is why an absolute path was used.

johannes-mueller:
Both points addressed, see commit messages.

x42:
Rebased + merged as Ardour 5.10-249 66964aaab  Thanks!

PR #362, opened 2017-07-20
Try to fix #7429

johannes-mueller:
#7429 states that the editor mixer strip is not updated when clicking on a route which is member of an already selected group.

This PR tries to fix it by calling `Editor::set_selected_mixer_strip()` after the click events on the TimeAxisView as well as on the MixerStrip of the Route.


pauldavisthefirst:
manually rebased and pushed. thanks!

PR #364, opened 2017-07-21
Increase range of a-comp#stereo release time

johannes-mueller:
Like for the mono version in f52443e17e818145d736e8f386b6a7aba6d8b5c7


x42:
Wow, you're on a PR roll :)
Rebased as Ardour 5.10-314 0109042e7 Thanks.

PR #365, opened 2017-07-21
Experimental: Plugin inline display in generic GUI

johannes-mueller:
This one finally addresses #7120 as a whole.

## Summary

* Put PluginDisplay into a publically available class, so that we can use it from outside ProcessorBox

* Make use of PluginDisplay in GenericPluginUI

* Introduce an extension_data so that a plugin can signal if it wants its display also in the genreic ui.

* As an example for a-comp: 
    * two different render functions depending on the display size (whole graph vs. only bars)
    * cache up to three cairo_surface_t instances to avoid constant reallocation displays of different sizes are show simultaneously.

## Disclaimer

This is still somewhat experimental and probably needs some conceptional discussion as well as some code refinement.

### DIFF
--- a/libs/plugins/a-comp.lv2/a-comp.c
+++ b/libs/plugins/a-comp.lv2/a-comp.c
@@ -588,6 +588,8 @@ render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
 	AComp* self = (AComp*)instance;
 	uint32_t h = MIN (w, max_h);
 
+	const float makeup_thres = self->v_thresdb + self->v_makeup;
+
 	if (!self->display || self->w != w || self->h != h) {
 		if (self->display) cairo_surface_destroy(self->display);
 		self->display = cairo_image_surface_create (CAIRO_FORMAT_ARGB32, w, h);
@@ -626,7 +628,7 @@ render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
 	}
 	if (self->v_thresdb < 0) {
 		cairo_set_source_rgba (cr, 0.5, 0.5, 0.5, 1.0);
-		const float y = -.5 + floorf (h * (self->v_thresdb / -60.f));
+		const float y = -.5 + floorf (h * (makeup_thres / -60.f));
 		cairo_set_dash(cr, dash1, 2, 2);
 		cairo_move_to (cr, 0, y);
 		cairo_line_to (cr, w, y);
@@ -658,16 +660,16 @@ render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
 	// draw signal level & reduction/gradient
 	const float top = comp_curve (self, 0);
 	cairo_pattern_t* pat = cairo_pattern_create_linear (0.0, 0.0, 0.0, h);
-	if (top > self->v_thresdb) {
+	if (top > makeup_thres) {
 		cairo_pattern_add_color_stop_rgba (pat, 0.0, 0.8, 0.1, 0.1, 0.5);
 		cairo_pattern_add_color_stop_rgba (pat, top / -60.f, 0.8, 0.1, 0.1, 0.5);
 	}
 	if (self->v_knee > 0) {
-		cairo_pattern_add_color_stop_rgba (pat, (self->v_thresdb / -60.f), 0.7, 0.7, 0.2, 0.5);
-		cairo_pattern_add_color_stop_rgba (pat, ((self->v_thresdb - self->v_knee) / -60.f), 0.5, 0.5, 0.5, 0.5);
+		cairo_pattern_add_color_stop_rgba (pat, (makeup_thres / -60.f), 0.7, 0.7, 0.2, 0.5);
+		cairo_pattern_add_color_stop_rgba (pat, ((makeup_thres - self->v_knee) / -60.f), 0.5, 0.5, 0.5, 0.5);
 	} else {
-		cairo_pattern_add_color_stop_rgba (pat, (self->v_thresdb / -60.f), 0.7, 0.7, 0.2, 0.5);
-		cairo_pattern_add_color_stop_rgba (pat, ((self->v_thresdb - .01) / -60.f), 0.5, 0.5, 0.5, 0.5);
+		cairo_pattern_add_color_stop_rgba (pat, (makeup_thres / -60.f), 0.7, 0.7, 0.2, 0.5);
+		cairo_pattern_add_color_stop_rgba (pat, ((makeup_thres - .01) / -60.f), 0.5, 0.5, 0.5, 0.5);
 	}
 	cairo_pattern_add_color_stop_rgba (pat, 1.0, 0.5, 0.5, 0.5, 0.5);
 
--- a/libs/plugins/a-comp.lv2/a-comp.c
+++ b/libs/plugins/a-comp.lv2/a-comp.c
@@ -38,6 +38,8 @@
 #define isfinite_local isfinite
 #endif
 
+#define DISP_CACHE_NUM 3
+
 typedef enum {
 	ACOMP_ATTACK = 0,
 	ACOMP_RELEASE,
@@ -88,7 +90,8 @@ typedef struct {
 #ifdef LV2_EXTENDED
 	LV2_Inline_Display_Image_Surface surf;
 	bool                     need_expose;
-	cairo_surface_t*         display;
+	cairo_surface_t** display;
+	cairo_surface_t* cached_displays[DISP_CACHE_NUM];
 	LV2_Inline_Display*      queue_draw;
 	uint32_t                 w, h;
 
@@ -130,6 +133,10 @@ instantiate(const LV2_Descriptor* descriptor,
 #ifdef LV2_EXTENDED
 	acomp->need_expose = true;
 	acomp->v_lvl_out = -70.f;
+	for (cairo_surface_t** d = acomp->cached_displays; d < acomp->cached_displays + DISP_CACHE_NUM; d++) {
+		*d = NULL;
+	}
+	acomp->display = acomp->cached_displays;
 #endif
 
 	return (LV2_Handle)acomp;
@@ -570,8 +577,10 @@ cleanup(LV2_Handle instance)
 {
 #ifdef LV2_EXTENDED
 	AComp* acomp = (AComp*)instance;
-	if (acomp->display) {
-		cairo_surface_destroy (acomp->display);
+	for (cairo_surface_t** d = acomp->cached_displays; d < acomp->cached_displays + DISP_CACHE_NUM; d++) {
+		if (*d) {
+			cairo_surface_destroy (*d);
+		}
 	}
 #endif
 
@@ -828,6 +837,42 @@ render_inline_only_bars (cairo_t* cr, const AComp* self)
 	cairo_stroke (cr);
 }
 
+static void
+aquire_display(AComp* self, uint32_t w, uint32_t h)
+{
+	cairo_surface_t** current = self->display;
+	while (1) {
+		if (!*current) {
+			break;
+		}
+
+		const uint32_t w_ = cairo_image_surface_get_width (*current);
+		const uint32_t h_ = cairo_image_surface_get_height (*current);
+		if (w == w_ && h == h_) {
+			self->w = w_;
+			self->h = h_;
+			break;
+		}
+		++current;
+		if (*current && current == self->cached_displays + DISP_CACHE_NUM) {
+			cairo_surface_destroy (*current);
+			*current = 0;
+		}
+	}
+
+	if (!*current) {
+		*current = cairo_image_surface_create (CAIRO_FORMAT_ARGB32, w, h);
+	}
+	cairo_surface_t* tmp = *current;
+	while (current != self->display) {
+		*current = *(current-1);
+		--current;
+	}
+	*self->display = tmp;
+	self->w = w;
+	self->h = h;
+}
+
 static LV2_Inline_Display_Image_Surface *
 render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
 {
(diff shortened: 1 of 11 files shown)


x42:
I like the way where this is going.

Is there a reason to make `in_gui` an interface, rather than a property (in the .ttl)? I suppose it's party my fault. The original idea was to use something like 
```
lv2:extensionData: <inline-uri#inteface>;
lv2:optionalFeature: <inline-uri#onGui>;
```
The problem is that the .ttl does not validate unless the the ontology in included in the .ttl or becomes an official spec. For some reason this keeps nagging me, even though I don't have a good solution.

As for the visuals, the mixer-strip inline reduction meter is a bit too bold IMHO, the /fat white/ border and grid could be more subtle.

johannes-mueller:
The question interface vs property I actually had in mind, when I writing "conceptional discussion". Making it an interface that just hands over a bool (in form of a pointer) is violating LV2's principle of separating static metadata from code. But I must admit that reading through the LV2 specs I did not find a way of indicating a think like that in the .ttl.

A second conceptional thing I'm not that happy with, is the fact that the plugin has to do the caching of the display objects. Naively I would think that the host should pass an ARGB32 pointer to the plugin to say "This is where I want you to render your display to." Then the host could keep caches for the different display sizes if it wants to.

Two reasons for that
* Every plugin has to implement the caching mechanism on its own.
* The host knows probably better how many different display sizes are going to be requested. When I hacked the caching for a-comp I clearly had ardour as host in mind, which can have three sizes simultaneously (editor-mixer strip, mixer, generic gui). What if tomorrow a host with five different displays comes around the corner?

The visuals need some refinment, as well as the layout of the generic gui. I also played around with a-eq and tried how that looks like in the generic ui. It has the problem, that it somehow changes its size when tweaking the parameters.

x42:
I'm not actually sure if caching is needed. malloc/free is fast compared to actually drawing.
Now that there's some code it would be interesting to profile it.

in a-comp's case (complete redraw at @25fps), yes having the host manage the canvas might be beneficial.

The initial motivation was for more static cases where the inline-display only changes when a parameter changes (think eq-curves). The host's channel-strip display is usually invalidated a lot. If the host manages the lifetime, the plugin does not know which parts is be redrawn. If the plugin manages the display it can also do a partial redraws.
The plugin may want to cache patterns (faceplates, scaled-background images), and has to manage the lifetime of those anyway. Most LV2 plugins with GUI have infrastructure for that already, easy to re-use.
The architectural design took into account that there may already be GUI code and to allow future extension to a complete lv2:RGBA32UI (like X11UI) with events for interaction (and the inline display become a special-cased GUI, just with no interactive events).

You do have a point though, for simpler plugins which don't have all that infrastructure already, moving the complexity to the host is the right thing to do.

But anyway, it's too late to change the specs now, besides that should probably be discussed on lv2-devel. You're breaking some new ground there: Multiple GUIs, different representations depending on the geometry which are visible at the same time.

x42:
PS. related to adding a boolean  `lv2:optionalFeature` meta-data, lua-dsp plugins will need a similar mechanism to optionally show their inline-display in the generic UI. There's no unified API for that, yet.

x42:
Crash when re-sizing the generic-plugin GUI:

```
(gdb) bt
#0  0x00007f88e5f81cb0 in cairo_image_surface_get_width () at /usr/lib/x86_64-linux-gnu/libcairo.so.2
#1  0x00007f889857c8ad in aquire_display (self=0x5587b9447960, w=303, h=300) at ../libs/plugins/a-comp.lv2/a-comp.c:850
#2  0x00007f889857c9ea in render_inline (instance=0x5587b9447960, w=303, max_h=300) at ../libs/plugins/a-comp.lv2/a-comp.c:885
#3  0x00007f88eb0996f2 in ARDOUR::LV2Plugin::render_inline_display(unsigned int, unsigned int) (this=0x5587be15b140, w=303, h=300) at ../libs/ardour/lv2_plugin.cc:977
#4  0x00005587b3c593ee in PluginDisplay::render_inline(_cairo*, unsigned int) (this=0x5587be1d7ff0, cr=0x5587b68b0e00, width=303) at ../gtk2_ardour/plugin_display.cc:95
#5  0x00005587b3c598d1 in PluginDisplay::on_expose_event(_GdkEventExpose*) (this=0x5587be1d7ff0, ev=0x7fffdc6844f0) at ../gtk2_ardour/plugin_display.cc:160
```

x42:
>I also played around with a-eq and tried how that looks like in the generic ui. It has the problem, that it somehow changes its size when tweaking the parameters.

That's the table-layout of the GUI itself, since 3d2bbf93efe the single-column layout can grow taking up more horiz space (somewhat improved since b42dbe9e4, but probably the knobs should become larger and define the min.width)

johannes-mueller:
The crash is odd: At first I thought it was just a stupid error of mine and trivial to fix. But then two other crashes come up with a backtraces no longer helpful to me. They both happen, as soon as the maximum number of cached surfaces has been exeeded and at least one of the cached surfaces has been destroyed in favor of a new one.

The crash happens first happens, when a-comp signals queue_draw
```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007fffbd1b4486 in run_mono (instance=0x2d53290, n_samples=1024) at ../libs/plugins/a-comp.lv2/a-comp.c:408
#2  0x00007ffff6f52433 in lilv_instance_run (instance=0x2d52f80, sample_count=1024) at /usr/local/include/lilv-0/lilv/lilv.h:1694
#3  0x00007ffff6f63036 in ARDOUR::LV2Plugin::run (this=0x3e866d0, nframes=1024, sync_work=false) at ../libs/ardour/lv2_plugin.cc:3021
#4  0x00007ffff6f60cd3 in ARDOUR::LV2Plugin::connect_and_run (this=0x3e866d0, bufs=..., start=0, end=1024, speed=0, in_map=..., out_map=..., nframes=1024, 
    offset=0) at ../libs/ardour/lv2_plugin.cc:2671
#5  0x00007ffff6ce1269 in ARDOUR::PluginInsert::connect_and_run (this=0x3e7b640, bufs=..., start=0, end=1024, speed=0, nframes=1024, offset=0, 
    with_auto=true) at ../libs/ardour/plugin_insert.cc:977
#6  0x00007ffff6ce285f in ARDOUR::PluginInsert::run (this=0x3e7b640, bufs=..., start_frame=0, end_frame=1024, speed=0, nframes=1024)
    at ../libs/ardour/plugin_insert.cc:1171
```

The second one when plugin instance is deleted:
```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007fffbd1b4486 in run_mono (instance=0x2d53290, n_samples=1024) at ../libs/plugins/a-comp.lv2/a-comp.c:408
#2  0x00007ffff6f52433 in lilv_instance_run (instance=0x2d52f80, sample_count=1024) at /usr/local/include/lilv-0/lilv/lilv.h:1694
#3  0x00007ffff6f63036 in ARDOUR::LV2Plugin::run (this=0x3e866d0, nframes=1024, sync_work=false) at ../libs/ardour/lv2_plugin.cc:3021
#4  0x00007ffff6f60cd3 in ARDOUR::LV2Plugin::connect_and_run (this=0x3e866d0, bufs=..., start=0, end=1024, speed=0, in_map=..., out_map=..., nframes=1024, 
    offset=0) at ../libs/ardour/lv2_plugin.cc:2671
#5  0x00007ffff6ce1269 in ARDOUR::PluginInsert::connect_and_run (this=0x3e7b640, bufs=..., start=0, end=1024, speed=0, nframes=1024, offset=0, 
    with_auto=true) at ../libs/ardour/plugin_insert.cc:977
#6  0x00007ffff6ce285f in ARDOUR::PluginInsert::run (this=0x3e7b640, bufs=..., start_frame=0, end_frame=1024, speed=0, nframes=1024)
    at ../libs/ardour/plugin_insert.cc:1171
```

Both crashes are clearly related to the caching of the display surfaces, as they don't occur in a commit state without the caching.

But I don't understand why the crash does not occur in my own code. If I read `PluginDisplay::render_inline()` correctly it does not keep a reference to the display pointer it received from the plugin's render function. It just copies data with memcpy() and then forgets the reference. So I thought, if I destroy cairo_surface_t instances in the plugin's code I would not invalidate pointers that the host might still want to use.

x42:
Maybe memory corruption? double free? Valgrind may have a hint.
Do you have a guesstimate if caching actually improves performance? Searching for the cached surface may be just as expensive as re-allocating ~50KB on the heap.

libs/plugins/a-comp.lv2/a-comp.c
```diff
@@ -665,15 +738,167 @@ render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
...
+	const float reduced_0dB = self->v_thresdb * (1.f - 1.f/self->v_ratio);
+	const float rt = x1 + wd * (60.f+reduced_0dB) / 70.f;
+	cairo_set_source_rgba (cr, 0.95, 0.0, 0.0, 1.0);
+	cairo_move_to (cr, rt, y1);
+	cairo_line_to (cr, rt, y1+ht);
+	cairo_stroke (cr);
+}
+
+static void
+aquire_display(AComp* self, uint32_t w, uint32_t h)
```
x42: acquire is spelt acquire :)

libs/plugins/a-comp.lv2/a-comp.c
```diff
@@ -665,15 +738,167 @@ render_inline (LV2_Handle instance, uint32_t w, uint32_t max_h)
...
+		const uint32_t w_ = cairo_image_surface_get_width (*current);
+		const uint32_t h_ = cairo_image_surface_get_height (*current);
+		if (w == w_ && h == h_) {
+			break;
+		}
+		if (*current && current == self->display + DISP_CACHE_NUM) {
+			cairo_surface_destroy (*current);
+			*current = NULL;
+		}
+		++current;
```
x42: This may increment current, past the end. there should be a `continue;` after the `*current = NULL;` in the if-clause above.

x42: Actually if you reach `current == self->display + DISP_CACHE_NUM`, you're already after the end.  The pointer is valid from   self->display + 0 ...   self->display + (DISP_CACHE_NUM - 1).

It may be cleaner to rewrite this just with a for loop  `for (idx = 0; idx < DISP_CACHE_SIZE; ++idx)` and use `self->display[idx]`

x42:
Initialization uses `calloc()` so `acomp->display` is already zeroed. here's a quick fix. Now the question is: how to best profile it?

```diff
diff --git a/libs/plugins/a-comp.lv2/a-comp.c b/libs/plugins/a-comp.lv2/a-comp.c
index ccade7418..119ddccbc 100644
--- a/libs/plugins/a-comp.lv2/a-comp.c
+++ b/libs/plugins/a-comp.lv2/a-comp.c
@@ -132,9 +132,6 @@ instantiate(const LV2_Descriptor* descriptor,
 #ifdef LV2_EXTENDED
        acomp->need_expose = true;
        acomp->v_lvl_out = -70.f;
-       for (cairo_surface_t** d = acomp->display; d < acomp->display + DISP_CACHE_NUM; d++) {
-               *d = NULL;
-       }
 #endif
 
        return (LV2_Handle)acomp;
@@ -575,9 +572,9 @@ cleanup(LV2_Handle instance)
 {
 #ifdef LV2_EXTENDED
        AComp* acomp = (AComp*)instance;
-       for (cairo_surface_t** d = acomp->display; d < acomp->display + DISP_CACHE_NUM; d++) {
-               if (*d) {
-                       cairo_surface_destroy (*d);
+       for (uint32_t i = 0; i < DISP_CACHE_NUM; ++i) {
+               if (acomp->display[i]) {
+                       cairo_surface_destroy (acomp->display[i]);
                }
        }
 #endif
@@ -853,8 +850,10 @@ aquire_display(AComp* self, uint32_t w, uint32_t h)
                        break;
                }
                if (*current && current == self->display + DISP_CACHE_NUM) {
+                       --current;
                        cairo_surface_destroy (*current);
                        *current = NULL;
+                       break;
                }
                ++current;
        }

```

x42:
cairo_image_surface_create() takes the same time as calloc() and malloc+memset.

profiling cairo_image_surface_create(); cairo_surface_destroy() with a CAIRO_FORMAT_ARGB32 surface:  0.4usec for 70x40, 0.5 usec for 102x40, ~2 usec for 128x128,  8usec for 256x256 and 87usec for 800x600  on a i7-5600U, 2.6GHz (average of 1M iterations).

rendering a-comp takes

1.663 ms (300 x 300px)
0.096 ms (102 x 40px)
0.091 ms (70 x 40px)

johannes-mueller:
I played around with valgrind a little bit. I must admit, that I am not that familiar with profiling, but I love to learn :)

If I interpret everything correctly the relevant functions use cpu cycles like this:
* `render_inline_full():` ~11M cycles/call
* `render_inline_only_bars()`: ~200k cycles/call
* `aquire_display()`: ~17k cycles/call

Comparing the cycles needed by `render_inline()` with and without caching it looks like this:
* without caching: 5166k cycles/call
* with caching: 4919 cycles/call

In both cases it was 4 calls using `render_inline_full()` and 5 calls `render_inline_only_bars()` So caching empty pixmaps to save calloc calls saves about 5% cpu cycles.

So according to my interpretation, caching makes only sense, when caching whole pixmaps, like the grid and the curve and then only render the level pattern when run() triggers queue_draw().

x42:
Counting CPU cycles (callgrind, cachegrind) is tricky, rendering here involves libcairo and libpixman. I'm not certain if syscalls inside a library are taken into account.  malloc()/free() may block (syscall needs time in the OS, but no CPU cycles)  gtk2ardour/arcall uses `--collect-systime=yes` for that reason (kcachegrind is a good visualization tool).

I've patched the code and used  `gettimeofday()` for basic measurements, though.

Still, your conclusion is sound. 
meters.lv2 does that for example: fixed background for the faceplate, only render a single line (meter-needle) on top.  Rendering the background (curves, shades, labels) is expensive.

johannes-mueller:
Renewed version available in #370.

PR #366, opened 2017-07-22
Update zh.po for v5.10 Chinese Language

YQ-YSY:
Update zh.po for v5.10 Chinese Language


x42:
Rebased as Ardour 5.10-351 f1ce28b6d

PR #368, opened 2017-07-22
Russian po updates, typo in automation_time_axis.cc fix

psemiletov:
Russian po updates, typo in automation_time_axis.cc fix
Also I've notice that the "range" term from the source is translated differently in many places (as "Область", "Диапазон", "Выделение") - I think there is a need to translate "range" constantly with one word.

### DIFF
--- a/gtk2_ardour/automation_time_axis.cc
+++ b/gtk2_ardour/automation_time_axis.cc
@@ -174,7 +174,7 @@ AutomationTimeAxisView::AutomationTimeAxisView (
 	   as the longest automation label (see ::automation_state_changed()
 	   below). be sure to include a descender.
 	*/
-	auto_dropdown.set_sizing_text(_("Mgnual"));
+	auto_dropdown.set_sizing_text(_("Manual"));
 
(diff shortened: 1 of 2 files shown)


gtk2_ardour/automation_time_axis.cc
```diff
@@ -174,7 +174,7 @@ AutomationTimeAxisView::AutomationTimeAxisView (
 	   as the longest automation label (see ::automation_state_changed()
 	   below). be sure to include a descender.
 	*/
-	auto_dropdown.set_sizing_text(_("Mgnual"));
```
x42: This _typo_ is intentional. see the comment above. The string needs to include a descender, at least in English.

"Play" is tallest (due to the descender), "Manual" widest.

psemiletov: Thanks, I see.

x42:
rebased + changed (reverted the "Mgnual") as Ardour 5.10-391 fe5f00c367

x42:
Oops I forgot to change the commit msg (exclude the automation-change). Sorry.

psemiletov:
Thank you!

PR #369, opened 2017-07-23
Fix #6280 – region (first_frame()==0) selectable with SnapRegionBoundary

johannes-mueller:
Alternative to #367. 

Issue #6280 states that when selecting ranges using SnapToRegionBoundary it's
not possible to select regions with `first_frame() == 0`. This is because
`Playlist::find_next_region()` does not consider region boundaries == pos but
only > pos. Thus it never considers pos == 0 to be a region boundary.

This solution tries to be as little invasive as possible without changing the
semantics of `PlayList::find_next_region()`. Therefore position 0 is added to the
region boundary cache if there's a region starting at position 0 in any track.

### DIFF
--- a/gtk2_ardour/editor_ops.cc
+++ b/gtk2_ardour/editor_ops.cc
@@ -721,9 +721,12 @@ Editor::build_region_boundary_cache ()
 		return;
 	}
 
+	bool maybe_first_frame = false;
+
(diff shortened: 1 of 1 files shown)


x42:
Rebased as 5.10-373 9bfe404b4

PR #370, opened 2017-07-23
Experimental: Plugin inline display in generic GUI (renewal)

johannes-mueller:
This is a renewed and rebased version of #365.

* Caching of allocated pixmap buffers is dropped as discussion in #365 has shown that caching empty pixmap buffers to save mere memory allocation does not increase performance to be worth it.

* Plugins now signal their wish to draw the display in the generic gui more in the spirit of LV2 by a `lv2:optionalFeature` in the .ttl file instead of in code by `extension_data()`.

### DIFF
--- a/gtk2_ardour/generic_pluginui.cc
+++ b/gtk2_ardour/generic_pluginui.cc
@@ -49,6 +49,7 @@
 #include "widgets/tooltips.h"
 
 #include "plugin_ui.h"
+#include "plugin_display.h"
 #include "gui_thread.h"
 #include "automation_controller.h"
 #include "gain_meter.h"
@@ -526,6 +527,12 @@ GenericPluginUI::automatic_layout (const std::vector<ControlUI*>& control_uis)
 		output_table->show_all ();
 	} else {
 		delete output_table;
+	}
+
+	if (plugin->has_inline_display ()) {
+		PluginDisplay* pd = manage (new PluginDisplay (plugin, 300));
+		pd->set_name("inside gui");
+		hpacker.pack_end (*pd, true, true);
 	}
 	show_all();
 
--- a/gtk2_ardour/generic_pluginui.cc
+++ b/gtk2_ardour/generic_pluginui.cc
@@ -529,9 +529,8 @@ GenericPluginUI::automatic_layout (const std::vector<ControlUI*>& control_uis)
 		delete output_table;
 	}
 
-	if (plugin->has_inline_display ()) {
+	if (plugin->has_inline_display () && plugin->inline_display_in_gui ()) {
 		PluginDisplay* pd = manage (new PluginDisplay (plugin, 300));
-		pd->set_name("inside gui");
 		hpacker.pack_end (*pd, true, true);
 	}
 	show_all();
@@ -551,6 +550,11 @@ GenericPluginUI::custom_layout (const std::vector<ControlUI*>& control_uis)
 		layout->attach (*cui, cui->x0, cui->x1, cui->y0, cui->y1, FILL, SHRINK, 2, 2);
 	}
 	hpacker.pack_start (*layout, true, true);
+
+	if (plugin->has_inline_display () && plugin->inline_display_in_gui ()) {
+		PluginDisplay* pd = manage (new PluginDisplay (plugin, 300));
+		hpacker.pack_end (*pd, true, true);
+	}
 }
 
 GenericPluginUI::ControlUI::ControlUI (const Evoral::Parameter& p)
--- a/gtk2_ardour/plugin_display.cc
+++ b/gtk2_ardour/plugin_display.cc
@@ -1,0 +1,187 @@
+/*
+    Copyright (C) 2017 Paul Davis
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; if not, write to the Free Software
+    Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
+
+*/
+
+#include <gtkmm/container.h>
+
+#include "gtkmm2ext/colors.h"
+#include "gtkmm2ext/gtk_ui.h"
+#include "gtkmm2ext/gui_thread.h"
+#include "gtkmm2ext/utils.h"
+
+#include "ui_config.h"
+
+#include "plugin_display.h"
+
+
+
+PluginDisplay::PluginDisplay (boost::shared_ptr<ARDOUR::Plugin> p, uint32_t max_height)
+	: _plug (p)
+	, _surf (0)
+	, _max_height (max_height)
+	, _cur_height (1)
+	, _scroll (false)
+{
+	add_events (Gdk::BUTTON_PRESS_MASK|Gdk::BUTTON_RELEASE_MASK);
+	_plug->DropReferences.connect (_death_connection, invalidator (*this), boost::bind (&PluginDisplay::plugin_going_away, this), gui_context());
+	_plug->QueueDraw.connect (_qdraw_connection, invalidator (*this),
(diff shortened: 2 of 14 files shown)


x42:
Looks good. Are you planning to take this further in some non-compatible way or is this ready to be merged?

johannes-mueller:
From my side the further work will be more like polishing. I've been using this branch now for like a week of mixing work and it turns out to be quite useful. For the next steps user feedback would be helpful.

So in short: Yes, I think it can be merged.

The following things IMO will need some work or at least be thought of.

* The exact way indicating stuff in the inline display might need a bit tweaking. During practical usage it turns out that the real life gain reductions sometimes are so low, that the visual feedback is a little bit too weak. The green bar now indicates the output level - make up gain because that's what I thought is most useful. Yet I'd like to know other users opinion on that.

* The widget layout in the generic gui needs polishing. So far I did just another `pack_end()` to squeeze the display in.

* Now that the compression graph is drawn in a far bigger size as before, we could think of writing actual figures to the axes.

* The decision which kind of display to draw is now quite a simplistic: if (width >= 200) { big(); } else { small(); ). Maybe there's some better alternative.

* The display caching should be the job of the plugin contrarily to what I thought in the beginning. For a-comp it would be probably worth while caching the grid.

All these could be done in a compatible way and also would benefit from feedback of beta testers or even real users.

The only thing that makes me hesitate slightly is if a8957b3 is ok

x42:
Until this becomes official lv2 extension (if ever) the URIs are as good as any.

I considered in-lining the  URI definitions in the .ttl (in case someone validates the meta-data):
ie. add most of https://github.com/x42/lv2/commit/ffd4ffd1ebe4b0bf11bb13e106418e4c1016f9ef#diff-3265973a67d5eecfdfb7b7483355943b  to the plugins'  .ttl   Compare to https://github.com/x42/midifilter.lv2/commit/7e699976df0c0933178c5d1b42e04b48a19573d6  but it's really a non-issue.

Anyway I'll merge it before Ardour 5.11.

You do raise good points: The rendering (weak colors) and the size/alignment/packing in the generic-plugin UI can be improved.  But GUI details like this are always incremental w/testing and practical use and take time. It's not a regression and no showstopper.

The max. inline display width is 250% * 112px (ui-scale , see rc_option_editor.cc FontScalingOptions and processor_box.cc for the 112px).  so >= 200 may or may not be enough.

johannes-mueller:
ACK

x42:
Rebased and merged as 5.10-416 6663a2675

magnetophon:
@johannes-mueller @x42 

Thanks guys, great news!
I'll test right away!

magnetophon:
Looks great, works great: many thanks!

Two questions about the inline display:
- Would it be possible to see input levels rather than output levels?
- It seems like the graph is only displayed when the channel is very wide. (I tested with a 16 input channel which displayed the graph when in wide mode, but not in narrow mode, and with a stereo channel)
  While this suits me personly, as I'd like to see just the bar(s) inline, I was wondering if it was intentional.

x42:
As for the latter: The graph is displayed in the generic Plugin UI

johannes-mueller:
As for the former: We can discuss what's more useful.

What are your points why you would prefer the input level?

magnetophon:
@johannes-mueller To me it makes sense to look at input level, cause that is before the compressor in terms of cause and effect.

johannes-mueller:
@magnetophon @zamaudio 

Here some of my experiments with visualization of a-compressor and the in-pipeline a-expander plugins. If you have some 15 minutes to watch and to tell me what you think, I would appreciate. 

http://johannes-mueller.org/a-compressor-demo.mp4

magnetophon:
Hi @johannes-mueller.

Thanks for making this!  :)
I love your inline displays!
I'd use the red line to indicate the maximum output level in the last x sec.

My suggestions for the plugin gui would be: 
- put the input meter before the output meter.
- make the level meter in the graph react at the attack and release speed.
- the dots in the graph are unneeded imho.

magnetophon:
@johannes-mueller Any news on this?

johannes-mueller:
I'd like to wait for @zamaudio's comment on [#7439](http://tracker.ardour.org/view.php?id=7439). If he's ok with changing attack and release behavior of a-comp, I'd like to include that.

zamaudio:
Hi @johannes-mueller I am happy with the change to the attack/release, lets put it into zam-plugins too. 👍

lucianodato:
This is awesome!

johannes-mueller:
Thanks the hint to Reiss and Massberg. AFAICS a-comp only implements equation 14 of the paper, but not equation 15, which fixes the drawback "the attack envelope is now also impressed upon the release envelope."

After a first glimpse to the paper I am not sure if the R&M approach (including eq. 15) or the one I implemented for the video is the better one. But I would be happy to implement theirs. 

However I am against a switch because making details like those switchable confuses most users. Before adding a switch like that we should add a switch for peak/rms detection IMHO.

x42:
@johannes-mueller  Would this change be backwards compatible?
ie. If you have a session with some a-comp, does it still sound the same? If not, that might be a reason to postpone this until 6.0 or just make a new plugin variant (uri change).

johannes-mueller:
@x42 Yes, it would affect existing sessions, which is a valid reason to post-pone it.

@lucianodato It turns out, that originally a-comp implements a peak detector that R&M call "decoupling smooth", whereas my suggestion is what they call "branching smooth". The branching variant has slightly more distortion according to their paper. However in the long run I would prefer a combination of the two that maybe splits the release time into a part to reduce distortion and the main part for the release. But that will need some more experimenting.

**So my sugesstion:**
For 5.11 we leave a-comp as it is as for this (already merged PR). Everything beyond that I will push into a branch, to make it accessible for anyone to experiment and contribute. Then we'll take the time to discuss, test and stabilize and finally merge it before 6.0

magnetophon:
I agree with @lucianodato that more flexibility is better in general. Ardour IMHO should not focus on beginners if it reduces options for advanced users.

Let's face it, learning audio production takes time, and the only way to have a chance of making something good sounding as a beginner is to use preset  instruments or samples.
As soon as you are working with recordings, you will need to train your ears.
Learning what a few knobs and switches do is easy in comparison. 

I agree with @x42 that backwards compatibility is important too, so a new uri would be a good idea.

Regarding FB/FF: I made an [N-channel compressor](https://github.com/magnetophon/faustCompressors/blob/master/RMS_FBFFcompressor.dsp) with a fade between FB and FF, and it also has a switch between "log domain return to threshold" and "linear domain return to zero" detection.
There is a peak and an RMS version, and if wanted I can make that a switch or a knob too.
I'musing uncoupled attack and release, and if it weren't for backward compatibility, I'd say that's all we need, since you can just increase the release and get the same sound as the other one.

magnetophon:
I did some testing with rev bd40b9132cbac2d2b79ba0ef480bd41d837f8f71, and it seems the release is already decoupled from the attack: if I set attack to 100ms, I get a lot less GR with minimum release VS release = 100 ms.

johannes-mueller:
IMHO the differences between branched and decoupled peak detection are too small to justify a switch.

I don't want to go too much into a philosophical debate on that, but IMHO the a-plugins should satisfy the majority of use cases in an stable, reliable, uncomplicated way. This does not reduce options for advanced users, as one can choose plugins with more options for more special use cases. Introducing more and more options increases complexity and makes maintainability harder. Ardour should focus on being a DAW, not a plugin suite. So elaborated effects with lots of options should be left to dedicated plugin development. Just my personal opinion.

About the envelope testing: the state in 5.11 does not have the branched peak detection. Of course you get more GR with 100ms attack than with 100ms release. Attack slows down increase of GR, release slows down decrease of GR. So rule of thumb is more attack less GR, more release more GR.

You can do the test with something like a kick drum beat. Compare the decrease of GR with 0.1ms attack to the one with 100 ms attack (both with no or little release). Then the GR goes away faster with attack than without it.

lucianodato:
@johannes-mueller fair enough. What you have done with inline gui is fantastic! Keep up the great work!

magnetophon:
@johannes-mueller the way I understood it so far is that with the current implementation, the actual release is the minimum of the setting of the release and the attack knob.
My testing indicates that understanding is incorrect.

johannes-mueller:
No, release is not the minimum of attack and release but the *sum* of attack and release. So put the attack to 100 ms, than there's no chance to get a fast release.

johannes-mueller:
As promised I set up a PR in my fork to continue the development there. Feel free to test comment and discuss over here: https://github.com/johannes-mueller/ardour/pull/1

PR #371, opened 2017-07-30
More Russian translation updated and fixes

psemiletov:
Just more Russian translation updated and fixes. When most of messages will be translated and estabilished well, then I'll be able to write a Russian manual for Ardour using the translated terms. Currently it is not possibly due to the differencies in translated terms, and the use of non-common words in the translation. For example, no one say "Циклическое воспроизведение" but all people just say shortly "петля" (i.e. "loop").


x42:
I think you should coordinate with @prokoudine who is the maintainer of the Russian translation.

x42:
rebased and applied as 5.10-398 31a9f24ef. Thanks!

PR #373, opened 2017-08-10
alsautil: show playback only devices

smk-embedded:
Hi guys!
I did my first mixing project with ardour and ran into this issue. I wanted to
mix and master on my laptop with a single usb dac. I was puzzled why the
device did not show up, but was working fine with qjackctl. I had a look at
the code and thought this little contribution might help others, as I hope
my use case is not so "beyond .. scope".
I finished the one mixing project without any problems, crashes or
whatsoever, so I would consider the patch slightly tested on linux with a
usb dac.
Regards,
Stefan

### DIFF
--- a/libs/backends/jack/jack_utils.cc
+++ b/libs/backends/jack/jack_utils.cc
@@ -273,7 +273,7 @@ void
 ARDOUR::get_jack_alsa_device_names (device_map_t& devices)
 {
 #ifdef HAVE_ALSA
-	get_alsa_audio_device_names(devices);
+	get_alsa_audio_device_names(devices, HalfDuplexOut);
 #else
 	/* silence a compiler unused variable warning */
 	(void) devices;


x42:
They should show up and work with Ardour's ALSA backend as-is.
Ardour's JACK backend asks for FullDuplex devices only.

x42:
That code cannot be removed, it is needed by the ALSA backend to specifically check for half-duplex devices (like your single DAC).

You could check `ARDOUR::get_jack_alsa_device_names()` and perhaps change  line 276

`get_alsa_audio_device_names(devices)`

to

`get_alsa_audio_device_names(devices, (AlsaDuplex)0)`

smk-embedded:
Spot on, man. That works perfectly. I now tested both alsa and jack backend and all devices show up as expected. I updated the patch.

x42:
Which version of jack do you use? (jackd --version)

I just tested this,  Ardour still launches jackd with the `-d alsa -d <device>`  and asks jack to connect full-duplex instead of  `-d alsa -P <playback only device>`.

The only half-duplex device here is a HDMI display (which only works at 4096fpp), so not the best test-case.

jackd starts correctly, only warns "Cannot open PCM device alsa_pcm for capture. Falling back to playback-only mode". It ostensibly works, Ardour plays, but it's silent.

Ardour/ALSA does work correctly (with sound), as does jackd with explicit `-P <device>` launched manually or from qjackctl. (perhaps a jack2 bug here 1.9.11 -- debian's 1.9.10+20150825git1ed50c92~dfsg-5)

x42:
Rebased as Ardour 5.11-97 587a2dee5

PR #376, opened 2017-08-20
Save and Edit template descriptions

johannes-mueller:
This makes the recently introduced template descriptions savable in "Save As Template" and editable in the template manager.

Template descriptions are now stored in the top-level of the template XMLTree rather than in `<Metadata/>`. This avoids that template descriptions spill into tags in exported audio files.

When saving a track or session as template, the user is prompted to give a template description with the session description resp. the track comment as a default.

Trailing whitespace of the descriptions is silently stripped before saving.


x42:
Rebased and merged as 5.11-73 258fa81d9

PR #377, opened 2017-08-21
prh: Small fix to /cue/send/enable feedback string

phedders:


x42:
Rebased and merged as Ardour 5.11-78 18060e597

PR #381, opened 2017-09-20
Fix Beats::normalize

ngeiswei:
Unfortunately I don't how to demonstrate the bug other than by using the WIP pattern editor (aka tracker interface), but I think the issue may seem obvious by reviewing the code.

Basically the problem comes from

https://github.com/Ardour/ardour/blob/master/libs/evoral/evoral/Beats.hpp#L64

that also invert ticks based on _beats, leading to wrong Beats construction, such as
```
Beats(0, -100)
```
ending up having negative `_ticks`.

Hmm, maybe a unit test would be welcome, wouldn't it?

### DIFF
--- a/libs/evoral/evoral/Beats.hpp
+++ b/libs/evoral/evoral/Beats.hpp
@@ -38,30 +38,19 @@ public:
 
 	Beats() : _beats(0), _ticks(0) {}
 
-	/** Normalize so ticks is within PPQN. */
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
i think i fixed this via a different change in the nutempo branch. but i'll still be sure to look at this again after i merge nutempo and master.

ngeiswei:
Cool! I'll obviously re-test it once nutempo has been merged to the master.

PR #382, opened 2017-09-24
Rename undo message "Create Note" to "add note"

ngeiswei:
To be more consistent with the other "add note" message in
MidiRegionView::create_note_at.

Generally speaking it's weird that some (a minority) of opname in undo messages start in upper case. Anyway reason for that?


ngeiswei:
The other weird think is an automation event is removed, the undo message recorded is "remove MIDI", even if it's a processor automation. It apparently comes from this line https://github.com/Ardour/ardour/blob/master/gtk2_ardour/editor_ops.cc#L4170

ngeiswei:
Let me close this premature pull request and study further what is going on...

PR #383, opened 2017-10-02
Adding missing option and detail about bugtracker

Varpie:
The french manpage was missing a few elements. This pull requests intends to make it equivalent to the english version.


x42:
This is no longer applicable. The 17+ year old, outdated man-page has been removed.

PR #385, opened 2017-10-18
A better method for "Close Regions Gaps"

Houston4444:
The actual method for "Close Region Gaps" can be improved.
Indeed, in the current code, each region is stretched of "pull back length" to the left,
the region on the left is stretched to overlap this one of "Fade Length".
The problem arises when using this method after split a region at onsets and snap to grid all new regions (Rhytme ferret).
If after "split at Onsets", a gap between two regions is bigger than "Pull Back Length", then the right region will be stretched of Pull Back Lenght to the left, and the left region will be stretched to the right until overlap the right one. And what can we ear on the right side of the left region ? The attack present at the beggining of the right region.

To avoid this, I propose this code which retains the same behavior except that it never stretch the left region to the right. If the left region is not close enought, the right one will be stretched to the left until the end of the left one. This avoids double attacks.

So, we can rename "Pull Back Length", in "Minimum Pull Back Lenght" and set it by defaut to 10ms instead of 30ms, by experience it seems reasonable for drums or guitars.

Also, the code I propose uses truely the fade duration defined by user instead of a simple overlap (for audio regions only). If "Fade Length" is set on 0ms, the default fade will be used (64samples), I think it's good the set it to 0 by default.

To test this, take an audio region of rhytmic guitar or drum, apply Rhytme Ferret, analyze region, split region at Onsets, select all new regions and "snap to grid". Then use "Close Gaps".

### DIFF
--- a/gtk2_ardour/editor_ops.cc
+++ b/gtk2_ardour/editor_ops.cc
@@ -7078,12 +7078,12 @@ Editor::close_region_gaps ()
 	list <boost::shared_ptr<Playlist > > used_playlists;
 
 	RegionSelection rs = get_regions_from_selection_and_entered ();
-
+						
 	if (!_session || rs.empty()) {
 		return;
 	}
-
-	Dialog dialog (_("Close Region Gaps"));
+	
+    Dialog dialog (_("Close Region Gaps"));
 
 	Table table (2, 3);
 	table.set_spacings (12);
@@ -7094,7 +7094,7 @@ Editor::close_region_gaps ()
 	SpinButton spin_crossfade (1, 0);
 	spin_crossfade.set_range (0, 15);
 	spin_crossfade.set_increments (1, 1);
-	spin_crossfade.set_value (5);
+	spin_crossfade.set_value (0);
 	table.attach (spin_crossfade, 1, 2, 0, 1);
 
 	table.attach (*manage (new Label (_("ms"))), 2, 3, 0, 1);
@@ -7105,7 +7105,7 @@ Editor::close_region_gaps ()
 	SpinButton spin_pullback (1, 0);
 	spin_pullback.set_range (0, 100);
 	spin_pullback.set_increments (1, 1);
-	spin_pullback.set_value(30);
+	spin_pullback.set_value(10);
 	table.attach (spin_pullback, 1, 2, 1, 2);
 
 	table.attach (*manage (new Label (_("ms"))), 2, 3, 1, 2);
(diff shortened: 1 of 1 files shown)


luzpaz:
Is this still relevant ? (or can it be closed?)

Houston4444:
> Is this still relevant ? (or can it be closed?)

Sorry for the late answer. IIRC, no, because @pauldavisthefirst was not agree to rework this action this way.
However, I still think that what I proposed for rhythm ferret (this PR and #388) is a good thing to do (probably not exactly the way it has been done). That said, theses two PR are certainly too old for easy merging, it can only be an inspiration source.

I may have done things in the wrong order, implementing before to discuss with upstream devs. I am still interested by this complete feature : something working well to auto-snap to grid percussive impacts even with many tracks. Maybe I should first propose a plan, and discuss to the way it can be implemented.

Houston4444:
Because there is no issue tracker here, I propose the plan in this discussion directly.

improve rhythm ferret to can easily snap onsets to grid on a multitrack percussive instrument
=======

This idea retakes the two old pull requests I submited several years ago (#385 and #388). Theses PRs are of course too old to be merged easily, but I still think that the global idea is good, and if you see the interest, I would like to rework on it.

# user steps

The function should work in 4 steps, once the user selects regions to proceed and goes to rhythm ferret in context menu.

## 1 ) Onsets detection ("Analyze")
    
It is important to filter the onsets in order to have maximum one onset per grid division, Otherwise we will have multiple regions at same points when we will snap regions to grid. It is also important that onsets are shared between equivalent regions, because we need to still have equivalent regions after split. For this reason, it is nicer that when user moves an onset, onsets with same position on equivalent regions are also moved (let's call them equivalent onsets).
    
In the PR #388, the filtered onset was the closest from grid and not the more powerful one. Because it was easier to code, and in the end it worked quite well. It also had the advantage to ignore some offset between grid division.
    
## 2 ) split regions on onsets
    
Nicer if splitted regions are selected once they are splitted.
    
## 3 ) snap regions on grid

Maybe a 'swing' option could be nice to apply some groove (not developed in #388)
    
## 4 ) fill holes between regions

It is important that regions are stretched to the left only, and never to the right, because they could contain the next attack on their right, then we could hear double attacks.
If regions are already overlapping, we can reduce the right of the region on the left, to clear the inaudible content.
The algorythm can adapt fades and region layer to avoid several stacks of regions when track is on stacked layers mode.
We need at least an option to set minimum pull back (10 ms by default by experience).

# GUI dialog and others

Maybe steps 3 and 4 could be done in one step, we could also propose to join regions at end.

On GUI side, it would be nice to have a very clear ferret dialog.

For example, we could have at bottom of the dialog, the following buttons :

[                 Analyze                ]
[Split region] [Snap to grid] [fill holes]

at startup, only "Analyze" is Enabled
"Split region" becomes enabled when onsets appears
Once "split region" is done, "Analyze" and "Split region" becomes disabled, "Snap to grid" becomes enabled.
Once "Snap to grid" is done, "snap to grid" become disabled and "fill holes" becomes enabled.
One "fill holes" is done, dialog could be closed.

# Pull requests
If you agree with a such plan, I suppose that you will appreciate multiple PRs, I suggest to split them this way:
1 ) equivalent onsets (onsets shared between equivalent regions)
2 ) filter onsets on grid division (with a checkbox to enable it in rhythm ferret)
3 ) "fill holes" algorythm, "fill holes between regions" added to the 'Operation' combobox
4 ) Ferret dialog cosmetics (when 1, 2, and 3 are merged)

PR #389, opened 2017-11-21
Nsm switch 

Houston4444:
NSM switch implementation which allows NSM to switch directly from a session to another one without closing and re-launching Ardour.

For better NSM experience, I also change behaviour of  folder argument, it now prefer the last snapshot to the default snapshot (I mean, this is a bug which appends too whit launching ardour6 mysessionfolder).

The main thing to do was to prevent "save or not" dialogs  and any saves during switch (ardour becomes dirty as a JACK port appears or disapears, so we prefer not save after switch order).

### DIFF
--- a/gtk2_ardour/ardour_ui.cc
+++ b/gtk2_ardour/ardour_ui.cc
@@ -974,6 +974,10 @@ ARDOUR_UI::get_transport_controllable_state ()
 void
 ARDOUR_UI::save_session_at_its_request (std::string snapshot_name)
 {
+	if (_nsm_switches) {
+		return;
+	}
+	
 	if (_session) {
 		_session->save_state (snapshot_name);
 	}
@@ -991,6 +995,10 @@ ARDOUR_UI::autosave_session ()
 	}
 
 	if (!Config->get_periodic_safety_backups()) {
+		return 1;
+	}
+	
+	if (_nsm_switches) {
 		return 1;
 	}
 
@@ -1107,7 +1115,7 @@ ARDOUR_UI::starting ()
 			 * The wrapper startup script should set the environment variable 'ARDOUR_SELF'
 			 */
 			const char *process_name = g_getenv ("ARDOUR_SELF");
-			nsm->announce (PROGRAM_NAME, ":dirty:", process_name ? process_name : "ardour6");
+			nsm->announce (PROGRAM_NAME, ":dirty:switch:", process_name ? process_name : "ardour6");
 
 			unsigned int i = 0;
 			// wait for announce reply from nsm server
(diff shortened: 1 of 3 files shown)


gtk2_ardour/ardour_ui.cc
```diff
@@ -3553,11 +3577,11 @@ ARDOUR_UI::get_session_parameters (bool quit_on_cancel, bool should_be_new, stri
 		} else if (likely_new && template_name.empty()) {
 
 			ret = build_session_from_dialog (session_dialog, session_path, session_name);
-
+
```
x42: Trailing whitespace

gtk2_ardour/ardour_ui.cc
```diff
@@ -3553,11 +3577,11 @@ ARDOUR_UI::get_session_parameters (bool quit_on_cancel, bool should_be_new, stri
 		} else if (likely_new && template_name.empty()) {
 
 			ret = build_session_from_dialog (session_dialog, session_path, session_name);
-
+			
 		} else {
 
 			ret = load_session (session_path, session_name, template_name);
-
+
```
x42: trailing whitespace

gtk2_ardour/ardour_ui.cc
```diff
@@ -3575,7 +3599,13 @@ ARDOUR_UI::get_session_parameters (bool quit_on_cancel, bool should_be_new, stri
 			ARDOUR_COMMAND_LINE::session_name = "";
 		}
 	}
+	
+	if (_nsm_switches && ret != 0 ) {
+			/*If there is an error at nsm open or switch, don't loop an error dialog, just quit. */
+			exit (1);
+	}
 
+	_nsm_switches = false;
```
x42: Is this the right place to reset `_nsm_switches` ?
There are various ways to load or construct a session and not all are via `get_session_parameters()`. I don't know if that applies to NSM though. But perhaps `ARDOUR_UI::set_session()` is a better place to reset it.

Houston4444: Hummm, I don't find where set_session() is called in get_session_parameters().
One thing is sure,  NSM always call get_session_parameters() to open or create a session (see nsm.cc line55). And if ardour isn't launch with NSM, _nsm_switches is never set.

Ardour is set as 'ready' for NSM, once get_session_parameters() is terminate (if there is no errror).

x42: set_session() is called whenever the session changes. AFICT it's the only central method that's always called in that case.

Houston4444: Ok cool. Just a little question about acronym... What means AFICT ???

x42: oops. Sorry, I meant AFAICT -- as far as I can tell.

Houston4444:
Replaced with #651

PR #390, opened 2017-11-21
Ignore the grid mode for finding the snaped sample

Houston4444:
Since the function "snap_regions_to_grid" is only called by user. This has to ignore if grid is on/off/magnetic .


pauldavisthefirst:
snap/grid were overhauled since this PR was submitted, and it no longer appears relevant.

PR #391, opened 2017-11-23
change behaviour of equivalent regions selection with modifier key

Houston4444:
This is a way to resolve the problem i mention here:   http://tracker.ardour.org/view.php?id=6826

It concerns selection of equivalents regions with key modifiers (on linux, Ctrl and Shift).
I saw in code that the behaviour is quite weird (maybe because of an old behaviour), if you press Ctrl + click, it select all equivalent regions only if tracks are selected. But there is no interest, because it deselect the tracks !!!

**So after a long reflexion, I've done this:**
if you click on a region with Primary key, it will add/remove to selection:
  - Only the clicked region if there is no selected region or if selection contains isolated equivalent region(s)
  - All equivalent regions in the other case

if you click on a region with Tertiary key, it will apply selection to:
  - only the clicked region if selection contains isolated equivalent region(s)
  - All equivalent regions in the other case

Note that a region with no equivalent is not considered as isolated region.

this way, naturally, if user want to select isolated regions, he starts his selection with Primary Key, else he starts his selection with no modifier key, and it does what user wants (I think).


Houston4444:
Howw, sorry. I (or git) just made a big git error, including a lot of undesired commits.

PR #392, opened 2017-11-30
Fix error when $PATH contains whitespaces

mseeber:
Previously, if $PATH contains whitespaces the startup script would split
the new exported $PATH on the first whitespace. This was observed on an
Ubuntu system where $PATH contained a directory with whitespaces.

The change adds weak quotation around the new path to prevent expansion
of whitespaces. $PATH though, is still expanded.

### DIFF
--- a/tools/linux_packaging/build
+++ b/tools/linux_packaging/build
@@ -252,7 +252,7 @@ fi
 # jack in the places where it might be
 #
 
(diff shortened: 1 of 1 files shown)


mseeber:
Please note, that the change to the build tool itself is not tested by me. Instead, on the target system it was confirmed, that weak quotation fixes the problem in the start script when editing the startup script directly.

PR #393, opened 2017-12-14
install_video_tools.sh: Use https instead of http

yurkobb:
Downloading executables over plain http opens a very wide attack surface on the user's machine, so https should always be used instead.

### DIFF
--- a/tools/videotimeline/install_video_tools.sh
+++ b/tools/videotimeline/install_video_tools.sh
@@ -130,16 +130,16 @@ done
 echo "installing video-tools to '${TARGETDIR}'."
 cd "$TARGETDIR" || exit 1
 
-HARVID_VERSION=$(curl -s http://ardour.org/files/video-tools/harvid_version.txt)
+HARVID_VERSION=$(curl -s https://ardour.org/files/video-tools/harvid_version.txt)
 echo "Downloading harvid-${MULTIARCH}-${HARVID_VERSION}."
 curl -L --progress-bar \
-	http://ardour.org/files/video-tools/harvid-${MULTIARCH}-${HARVID_VERSION}.tgz \
+	https://ardour.org/files/video-tools/harvid-${MULTIARCH}-${HARVID_VERSION}.tgz \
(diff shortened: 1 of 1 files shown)


x42:
This script is not intended to be called by end-users.  It's for automatic builds on machines that ardour.org controls.  ..and no, https by itself won't make any difference: just encrypting the tranport-layer and relying on certs won't help much.

..but if it helps to make you sleep eaiser: http://ardour.org currently re-directs to https:// anyway :)

yurkobb:
Hmm, but it's [mentioned in the docs](http://manual.ardour.org/video-timeline/setup/), that's how I actually arrived at it. Probably the reference should be removed then? Also the link to git.io in the manual is also using http...

x42:
well, I don't have any objections to merge it.. 

I'm just saying it's a lot easier to do a HTTPS MITM attack on a AWS based systems compared to injecting malware into a binary inside a tar.  If someone is really interested to hijack the binary, https won't help. We should probably check a 3rd party server for sha1/sha256 signature integrity.

yurkobb:
Agree. However, there's also a possibility of an infected wifi router or something else on the level of the user's ISP that's not aware about ardour specifically, but is rewriting all common software download formats for instance. It may easily strip the https redirect if it ocours as well. Sounds very hypothetical, but it's just better to have too much https than to have too little.

x42:
Sadly this breaks on the OSX 10.6 build-machine. curl fails with the Let'sEncrypt root certificate used on ardour.org on older mac systems.

yurkobb:
I'm a bit confused about the purpose of this script.

> This script is not intended to be called by end-users. It's for automatic builds on machines that ardour.org controls. 

If it is not meant to be used outside ardour.org infrastructure, maybe it should be removed from the public source altogether?

pauldavisthefirst:
We're not in the habit of hiding stuff. There are a variety of scripts/testing tools that don't make much sense for end-users.

x42:
I'll close this issue. There is no benefit to encrypt the file transfer.

x42:
It is used by CI, the nightly builder.

PR #395, opened 2018-01-31
Create M-Audio_Axiom_AIR_Mini_32.map

shawnpdoherty:
List CC events for transport controls and knobs. List Note events for Trigger Pads. Map transport controls and knobs.


x42:
Looks good. thanks!

PR #396, opened 2018-02-04
Add Midnam for Roland TB-03

7890:
Named controllers according to MIDI implementation chart.
https://www.roland.com/global/support/by_product/tb-03/


x42:
rebased as 6.0-pre0-607 00e1e12ba

PR #398, opened 2018-03-28
Issue 7549 Duplicate Track Order

relascope:
Problem: http://tracker.ardour.org/view.php?id=7549

The duplicated Routes are now ordered like in the UI. 
Previous they were ordered in the order of their selection.


pauldavisthefirst:
basic idea looks sound. but why not just use a sort e.g. libs/ardour/session_route.cc line 39

relascope:
didn't know about it. 
New commit.

PR #402, opened 2018-07-15

cooltehno:
blueberry_milk_correction_150718
Hello Paul & people! Good day to everybody:)
I've got three questions:
1.This is my attempt to change a theme file as an example. If this my example PR is correct and good from your vision, may I make the same to others (cainville, clear-gray & cubasish - illustrations are already prepared :))?
2. Are you planning to add an item 'midi note selected fill', because only the outline of the selected notes are not so visible as it was in ardour 5.x (at my view)? Or may be the outline could be more thick?
3. Is it the right place to edit the ardour6's theme files (I mean the thread or folder or how it's correctly called here in 'github')?

_Specifically to Blueberry-milk:_
This commit changes all the default midi colors (0-5) to the specific design of the theme. Also the default dark color of the selected midi note outline is changing to the bright green color (more visible and matchs to design).

Big thanks!

### DIFF
--- a/gtk2_ardour/themes/blueberry_milk-ardour.colors
+++ b/gtk2_ardour/themes/blueberry_milk-ardour.colors
@@ -115,12 +115,12 @@
     <Color name="meter color7" value="0xff8700ff"/>
     <Color name="meter color8" value="0xff0000ff"/>
     <Color name="meter color9" value="0xff0000ff"/>
-    <Color name="midi color0" value="0x1e7727ff"/>
-    <Color name="midi color1" value="0x52af25ff"/>
-    <Color name="midi color2" value="0x85e524ff"/>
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
1) this is all correct. you don't need to run the colors by me. we are allowing themes to change MIDI meter colors, but do not want them to change audio meter colors.

2) the fill color no longer changes to indicate selection. we could thicken the outline.

3) yes, this is how to edit the files and submit a PR!

big thanks to you.

cooltehno:
_"we are allowing themes to change MIDI meter colors, but do not want them to change audio meter colors."_

the separate sort of color slot ('midi color0-5') has saved a lot of time in my work, bacause I didn't need to check all the design of the theme after the editing of these colors. - Thanks!!

PR #406, opened 2018-07-18
Change the velocity bar inside notes to extend to edges of note

alcomposer:
Keep Paul Davis original design. Having the velocity bar inset from note causes distracting space at beginning of note when viewing a MIDI note that is zoomed out.


pauldavisthefirst:
manually applied (due to merge conflicts). thanks!

PR #411, opened 2018-08-10
Add pick-up mode and use it for knobs and faders

terencode:
The pick up mode ignores new controller values until they match with the current setting of the stripable's ac


pauldavisthefirst:
manually rebased.

PR #412, opened 2018-08-12
Launch Control XL surface: add a mapping for cancel_all_solo

lentferj:


lentferj:
Revoked, better version coming

PR #413, opened 2018-08-13
LCXL: implement mute, solo and record arm toggles

lentferj:
Holding the Device button and pressing one
    of mute, solo and record arm will toggle
    selected tracks (use Track Focus buttons)
    accordingly.


pauldavisthefirst:
manuall rebased.

PR #417, opened 2018-08-18
LCXL: Add knob leds and refine colors

terencode:


x42:
rebased and merged as 6.0-pre0-963- 51b5c01b8

PR #418, opened 2018-08-18
LCXL: Add configuration option for handling master

terencode:
In the Controller's settings you can now
    choose between two operation modes:
    1)  8 track mode
    2)  7 track plus master mode

    In case 2) fader 8 is fixed on the master


x42:
rebased and merged as 6.0-pre0-964 74ad41f8d

PR #420, opened 2018-08-18
LCXL: track select and knobs' leds changes

terencode:


libs/surfaces/launch_control_xl/controllers.cc
```diff
@@ -394,10 +394,14 @@ LaunchControlXL::update_knob_led(uint8_t n)
 
 		case 3:
 		case 7:
-			if (stripable[n] && stripable[n]->is_selected()) {
-				color = AmberFull;
+			if (stripable[n] && stripable[n] == master) {
```
x42: prefer  `stripable[n]->is_master()`

x42: Does there really have to be a check if `stripable[n]` is set in every case?  would it not make more sense to bail out early?

PR #421, opened 2018-08-20
LCXL: fix some more typos

lentferj:


x42:
rebased as 6.0-pre0-967 b80e0793

PR #422, opened 2018-08-22
LCXL: save assignment of 8th fader in config

lentferj:


x42:
rebased as 6.0-pre0-968 d17b6c5ee

PR #423, opened 2018-08-22
LCXL: Add some Mixbus specific mappings

lentferj:


x42:
rebased as 6.0-pre0-970 266c5fbc16b7c

PR #424, opened 2018-08-24
Lua reset_mixer script: fix and expand

lecoqlibre:
The script was broken (regular Ardour 5.12).
I needed to also reset pan, pan width and some automation.
Solo and all processors options wasn't prompted but used. 

- Remove the is_channelstrip() and is_nonbypassable() calls.
- Add options: solo, pan, pan width, gain automation, pan automation and pan width automation.
- Add radio buttons for sends, inserts and plug-ins.
- Arrange prompt dialog: list, names and default values (now false to avoid unwanted behavior).
- Add my name to the authors list.
- Comment and refactor code.

### DIFF
--- a/scripts/reset_mixer.lua
+++ b/scripts/reset_mixer.lua
@@ -1,89 +1,139 @@
 ardour {
 	["type"] = "EditorAction",
 	name = "Reset Mixer",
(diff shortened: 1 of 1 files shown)


x42:
Rebased and merged as 6.0-pre0-983 e546b109c  Thanks!

PR #427, opened 2018-08-30
LCXL: Fix check if ports exist

terencode:


x42:
rebased and merged as 6.0-pre0-991 2a78f0a85

PR #429, opened 2018-09-21
LCXL: Fix a crash on exit

lentferj:


x42:
Merged as 6.0-pre0-1034 e52c8e376

PR #430, opened 2018-09-24
Mackie protocol fixes

agraef:
Both changes are utterly trivial, but they solve two nagging issues with the MCP interface for me, so I hope that they can be merged at some point (also if there ever is another 5.x bugfix release before Ardour 6 officially comes out). I'm using these with the 5.12 release, but as far as I can tell, this part hasn't changed since, and Paul's article on Ardour 6 development doesn't mention any planned changes there either.

The first commit just fixes up the broken feedback for the bank left/right buttons, and should hopefully be uncontroversial.

The second commit is probably a bit more problematic, so let me elaborate. ;-)

Ardour places a restriction on channel-right, it will only move the bank to the right one channel if there's still a full bank there. This has been there for a very long time, and probably for a good reason. But I'm using a single-strip device, the [X-Touch ONE](http://www.musictribe.com/Categories/Behringer/Computer-Audio/Desktop-Controllers/X-TOUCH-ONE/p/P0CAP), together with other MCP-compatible devices such as the nanoKONTROL2 or a full-size X-Touch, which are both connected to Ardour's Mackie port at the same time. Which works great, except that I can't access all strips with the X-Touch ONE with a bank size of 8. Just changing the device description to a bank size of 1 works, but is impossible if I want to use an 8-strip device along with it. Also, being able to flip through a large number of strips more quickly with the bank controls is just *so* convenient even if using just the single-strip device on its own.

This wouldn't be much of an issue if Ardour would allow different MCP devices to be connected at the same time, each with their own device device description, but that isn't possible right now, and I'd consider this overkill (and confusing) anyway. I'd say that the present Mackie interface works just fine as it is except for this single issue. So the easiest way to get single-strip devices working with a >1 bank size is to simply relax the condition on the initial bank number in `MackieControlProtocol::next_track()`, which is what I've done here.

### DIFF
--- a/libs/surfaces/mackie/mackie_control_protocol.cc
+++ b/libs/surfaces/mackie/mackie_control_protocol.cc
@@ -218,7 +218,7 @@ void
 MackieControlProtocol::next_track()
 {
 	Sorted sorted = get_sorted_stripables();
-	if (_current_initial_bank + n_strips() < sorted.size()) {
+	if (_current_initial_bank + 1 < sorted.size()) {
 		switch_banks (_current_initial_bank + 1);
 	}
 }
--- a/libs/surfaces/mackie/mcp_buttons.cc
+++ b/libs/surfaces/mackie/mcp_buttons.cc
@@ -120,7 +120,7 @@ MackieControlProtocol::left_press (Button &)
 LedState
 MackieControlProtocol::left_release (Button &)
 {
-	return none;
+	return off;
 }
 
 LedState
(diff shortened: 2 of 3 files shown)


agraef:
While I was at it, also fixed a minor glitch which caused the feedback for the view selection buttons to be garbled (would only ever light the USER button).

agraef:
I've tested all this on an X-Touch quite extensively now, works all fine for me. I didn't notice any adverse effects from the second commit either.

agraef:
That's fantastic, thanks a bunch for the quick response! I can ask David whether he's willing to add this as a patch to the existing 5.12 Arch package for the time being, since 6.0 is still some time away I guess.

Can't wait to give 6 a try myself, it's going to be an amazing leap forward. But I've seen your advice to Joern and so I'll wait until the dust begins to settle.

pauldavisthefirst:
manually applied and committed after checking on a Mackie Universal Pro. Thanks!

agraef:
Thanks!

PR #431, opened 2018-10-12
Follow fluidsynth's API changes introduced with 2.0.0

schnitzeltony:
Build tested with fluidsynth 1.1.11 and 2.0.1


x42:
Thank you to get the ball rolling! Rebased and applied without #ifdef's  as 5b03c41f6c4.

PR #432, opened 2018-10-17
LCXL: complete overhaul and Mixbus support

lentferj:


x42:
Rebased and merged as 6.0-pre0-1157 05d353959  Thanks!

PR #433, opened 2018-10-17
Prevent excessive meter redraws for inactive meters at zero II

schnitzeltony:
Same as be826f363522dbca96c8263d404c57c4fe9ff0e2

Signed-off-by: Andreas Müller <schnitzeltony@gmail.com>


x42:
Thanks!

PR #438, opened 2018-12-12
LCXL: some more small tweaks

lentferj:
* assign Pan knob to Comp thresh as secondary function
    * make cancel_all_solo and cancel_all_mute work in device mode, too


x42:
Rebased and merged as 6.0-pre0-1410 5d2a03578  Cheers!

PR #439, opened 2018-12-27
Fix Count-in not taking effect when punch-in is enabled.

xaptronic:
While recording some tracks, I noticed that count-in did not work with punch-in enabled.

To fix this, check whether punch-in is enabled and record is enabled in addition to if we are actively recording when deciding whether to run count-in clicks.


x42:
Nice find! Rebased as 6.0-pre0-1466 53c6b7f90

PR #441, opened 2019-01-01
Add Midnam for Yamaha MX 49/61/88

liopic:
Midnam file with preset names, as defined in MX49/MX61 Data List from:
https://ca.yamaha.com/en/products/music_production/synthesizers/mx_series/downloads.html


x42:
Thanks!

PR #442, opened 2019-01-03
Lcxl: another set of small fixes

lentferj:


x42:
Merged as 6.0-pre0-1480 ff0ee75a5  Thanks!

PR #443, opened 2019-02-18
Added MIDI Remap Lua DSP plugin

Linux-cpp-lisp:
Hi all,

This is an Ardour Lua DSP script I put together for remapping arbitrary MIDI notes in real time. I [posted it on the forums](https://discourse.ardour.org/t/lua-dsp-plugin-for-real-time-midi-remapping/100475), and at the suggestion of Robin Gareus, I'm submitting a PR to have it added to Ardour.

It can map any MIDI note to any other. Polyphonic aftertouch should also be remapped, though I haven't tested that, lacking any easy way to produce polyphonic aftertouch data. Timing and velocities are unaffected, and all other MIDI data is passed through untouched. By default it allows up to 10 remappings; just change `N_REMAPINGS` in the code if you need more.

One major improvement -- if possible -- would be to update the `translation_table` only when the parameters have changed, rather than every single buffer. I was unable to find any callback for when the parameters have been modified, but if one does exist, I'll certainly update the code to take advantage of it. 

Thanks,
A.

### DIFF
--- a/scripts/midi_remap.lua
+++ b/scripts/midi_remap.lua
@@ -22,7 +22,7 @@ end
 function dsp_params ()
 
     local map_scalepoints = {}
-    map_scalepoints["Off"] = OFF_NOTE
+    map_scalepoints["None"] = OFF_NOTE
     for note=0,127 do
         local name = ARDOUR.ParameterDescriptor.midi_note_name(note)
         map_scalepoints[string.format("%03d (%s)", note, name)] = note
@@ -72,7 +72,7 @@ function dsp_run (_, _, n_samples)
     local translation_table = {}
     local ctrl = CtrlPorts:array()
     for i=1,N_REMAPINGS*2,2 do
-        if not (ctrl[i] == OFF_NOTE or ctrl[i + 1] == OFF_NOTE) then
+        if not (ctrl[i] == OFF_NOTE) then
             translation_table[ctrl[i]] = ctrl[i + 1]
         end
     end
@@ -87,7 +87,9 @@ function dsp_run (_, _, n_samples)
         if (#d == 3) and (event_type == 9 or event_type == 8 or event_type == 10) then -- note on, note off, poly. afterpressure
             -- Do the mapping - 2 is note byte for these types
             d[2] = translation_table[d[2]] or d[2]
-            tx_midi (t, d)
+            if not (d[2] == OFF_NOTE) then
+                tx_midi (t, d)
+            end
         else
             tx_midi (t, d)
         end
--- a/scripts/midi_remap.lua
+++ b/scripts/midi_remap.lua
@@ -1,0 +1,103 @@
+ardour {
+    ["type"]    = "dsp",
+    name        = "MIDI Remap Notes",
+    category    = "Utility",
+    license     = "MIT",
+    author      = "Alby Musaelian",
+    description = [[Map arbitrary MIDI notes to others. Affects Note On/Off and polyphonic key pressure.]]
+}
+
+-- The number of remapping pairs to allow. Increasing this (at least in theory)
+-- decreases performace, so it's set fairly low as a default. The user can
+-- increase this if they have a need to.
+N_REMAPINGS = 10
+
+function dsp_ioconfig ()
+    return { { midi_in = 1, midi_out = 1, audio_in = 0, audio_out = 0}, }
+end
+
+function dsp_params ()
+    local note_names = {"C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B"}
+
+    local map_scalepoints = {}
+    for note=0,127 do
+        local octave = -5 + math.floor(note / 12)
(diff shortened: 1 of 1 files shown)


scripts/midi_remap.lua
```diff
@@ -0,0 +1,103 @@
...
+        midiout[cnt] = {}
+        midiout[cnt]["time"] = time;
+        midiout[cnt]["data"] = data;
+        cnt = cnt + 1;
+    end
+
+    -- We build the translation table every buffer because, as far as I can tell,
+    -- there's no way to only rebuild it when the parameters have changed.
+    -- As a result, it has to be updated every buffer for the parameters to have
+    -- any effect.
```
x42: You could store it globally and only update it when a parameter has changed. Not sure if it's worth it though.

An alternative is to not initialize the table with a unity matrix but use Lua's sparse array. 

```
  local translation_table = {}
  local ctrl = CtrlPorts:array()
  for i=1,N_REMAPINGS*2,2 do
    translation_table[ctrl[i]] = ctrl[i + 1]
  end
-- .....
  d[2] = translation_table[d[2]] or d[2]
```

scripts/midi_remap.lua
```diff
@@ -0,0 +1,103 @@
...
+
+    -- Restore translation table
+    for i = 0,127 do
+        translation_table[i] = i
+    end
+    -- Update it
+    local ctrl = CtrlPorts:array()
+    for i=1,N_REMAPINGS*2,2 do
+        translation_table[ctrl[i]] = ctrl[i + 1]
```
x42: If there are duplicate mappings, the last one wins.

This is mainly relevant for the default  `000 -> n`.  Try re-mapping note 0 using any but the last control.
One solution would be to add a dedicated  `"off" == -1` case as control parameter.

scripts/midi_remap.lua
Linux-cpp-lisp: Is there a callback for when the parameters have changed? If not, I would have to store the past parameter values and compare them, which takes nearly as many operations as rebuilding the table and introduces an extra branch, so I doubt it would actually change performance (I know that's C-level nitpicking, but either way I doubt the differences are measurable at this level).

I opted against the sparse array since my research indicated that tables indexed with contiguous positive integers were implemented under the hood as C-style arrays, and as a result would hopefully be efficient... but since it looks like most Lua versions [do a hash table anyway](https://stackoverflow.com/questions/26843589/how-does-lua-5-x-represent-sparse-arrays), maybe it's better to just clean up the code and use one.

scripts/midi_remap.lua
Linux-cpp-lisp: Since I'm switching to the sparse arrays, I'll add an dedicated "off" option

scripts/midi_remap.lua
```diff
@@ -0,0 +1,103 @@
...
+
+function dsp_params ()
+    local note_names = {"C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B"}
+
+    local map_scalepoints = {}
+    for note=0,127 do
+        local octave = -5 + math.floor(note / 12)
+        local name = note_names[(note % 12) + 1]
+        map_scalepoints[string.format("%03d (%s%d)", note, name, octave)] = note
+    end
```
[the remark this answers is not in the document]
Linux-cpp-lisp: Great! I didn't know that function existed, but it's much better than this ad-hoc approach.

scripts/midi_remap.lua
Linux-cpp-lisp: Whoops. Actually, the sparse array won't help: say there is a mapping from 2 -> 3 that is later changed to 5 -> 3, for example. If `translation_table` isn't reinitialized to the identity, now *both* 2 and 5 will map to 3... I think I'll leave it as is.

(I'm also pretty sure it's already a global -- omitting `local` in the declaration is sufficient, right?)

x42: > Is there a callback for when the parameters have changed? If not, I would have to store the past parameter values and compare them

yes, you would have to store and compare. It's not worth it.

>  If translation_table isn't reinitialized to the identity, ..

`local translation_table = {}` initializes it -- you can also get rid of the the `dsp_init` function and global definition.

Linux-cpp-lisp:
OK, I've resolved the edge cases, switched to the sparse `translation_table`, added an "off" option, and switched to the localized note names. Thanks for the feedback @x42 !

x42:
Now "to" can be "off" as well.  I'd expect that to drop the note.

What do you think? should "to" not be able to be "off", or should a new feature be added to filter, drop notes in that case?

Linux-cpp-lisp:
Good idea -- I've changed it so that mapping a note to "Off" (which I've renamed "None" for clarity) simply filters out the events, ` > /dev/null` style.

Great!

x42:
Awesome! rebased and merged as 6.0-pre0-1578 f0d227f92

Thanks!

PR #446, opened 2019-03-08
Fixes mistakes in the French and German translation files

trebmuh:
This PR fixes a mistake in the French translation file introduced in https://github.com/Ardour/ardour/pull/445 yesterday.

While building it, I saw that it was complaining about 2 `fatal errors` in the de.po file:

```
Generating po/de.po
po/de.po:10176: 'msgid' and 'msgstr' entries do not both end with '\n'
po/de.po:10201: 'msgid' and 'msgstr' entries do not both end with '\n'
msgfmt: found 2 fatal errors
```

I'm trying here to correct those as well but please @x42 , double check that I'm not introducing another mistake in the German file while doing that.


x42:
Rebased and merged as 6.0-pre0-1727 1df160cbf. Thanks!

PR #447, opened 2019-03-09
fix crash in Keyboard Shortcut Widget / "Print Bindings"

2ny:
How to reproduce the crash:
open the "Keyboard Shortcut Widget", click on "Print Bindings"

The fix add checks on all iterators in the loop, so that each of them is always valid.
I don't have the domain-knowledge to know if an other (underlying) bug exist: should those 3 collections have the exact same number of elements?

### DIFF
--- a/gtk2_ardour/po/de.po
+++ b/gtk2_ardour/po/de.po
@@ -10175,7 +10175,6 @@ msgstr ""
 msgid "  -b, --bindings              Display all current key bindings\n"
 msgstr ""
 "  -b, --bindings                   Alle momentanen Tastaturkürzel ausgeben\n"
-"\""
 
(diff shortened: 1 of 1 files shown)


x42:
This rather seems to hide an issue.

If it is possible that path, keys and labels can get out of sync, that would be an issue elsewhere.

x42:
Looks like this is due to recent incomplete, re-design of the ActionManager. `ActionManager::get_all_actions()` vs `Bindings::get_all_actions()`  -- the former does not add any value to `keys` at all currently.

2ny:
grrr, sorry, seems I've messed-up my repo. My bad. Trying to fix that.

PR #450, opened 2019-05-12
Drag and drop a file from Dolphin is now working even if the file has…

jmtrivial:
… utf8 non-latin1 characters in its name.

This fix corresponds to bug ID 0007253

### DIFF
--- a/gtk2_ardour/editor.cc
+++ b/gtk2_ardour/editor.cc
@@ -3407,7 +3407,7 @@ Editor::convert_drop_to_paths (
 		   THERE MAY BE NO NULL TERMINATING CHAR!!!
 		*/
 
-		string txt = data.get_text();
+		string txt = Glib::convert(data.get_text(), "ISO-8859-1", "UTF-8");
 		char* p;
 		const char* q;
 


x42:
I think that should probably be `Glib::locale_to_utf8()`. ISO-8859-1 is not usually the default.

e.g. In my case the system is using `en_US.UTF-8` and the [ext3] file-system is also UTF8.

What is the output of `locale` on your system?

jmtrivial:
My locale is fully utf8 : 

LANG=fr_FR.UTF-8
LANGUAGE=
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL=

But I think (not sure) that when you drag'n drop a file from dolphin, the file names are encoded in utf8, rather than encoded in uri from nautilus. I suppose that gtk considers that it should be uri encoded (then in latin1 I suppose, in 8 bits), then split the utf8 characters into two characters when it comes from dolphin.
I'm not enough an expert in gtk to check how it works...

jmtrivial:
Try on your system with a file named "téléchargement.wav" for example, and see what appends by writing in the console data.get_text() when the file come from a drag'n drop from nautilus, then from dolphin.

jmtrivial:
Ok, I think I found the correct bugfix. I close this pull request and create a new one.

PR #451, opened 2019-05-13
Fix drag and drop from nautilus/dolphin

jmtrivial:
Drag and drop from nautilus or dolphin is now considered as a list of uri rather than a plain text. 
* Fix bug ID 0007253 https://tracker.ardour.org/view.php?id=7253
* remove hand written code to parse plain-text description of the filenames


x42:
Is it really safe to allow "text/plain" as falback, but remove the parser that generates the uri-list from text?

x42:
I've picked the first commit only (184647612795)  and also updated other DnD receiver (8ae4fe1aa9) accordingly.

PR #452, opened 2019-06-03
added an Ardour-focused smartphone layout for Touch OSC

dayer4b:
At the recommendation of @ovenwerks , I am submitting a Touch OSC layout that I made.  With Len’s layout as a starting point, I designed a new one.

It uses 4 pages in order to split up functions and I relied heavily on the “encoder” form for scrubbing position and gain. I included a page dedicated to “track selection” that covers 42 tracks. I was hoping I could find some dynamic thing there, but this was the best I could come up with.

https://discourse.ardour.org/t/touchosc-on-android/89903/16


x42:
Sorry it took so long -- Rebased as 6.0-pre0-2521 549b2447f8 Thanks!

PR #453, opened 2019-07-14
Add patchfiles/Moog_Slim_Phatty.midnam

jmaibaum:
Adding a MIDI Name Document for the [Moog Slim Phatty](https://www.moogmusic.com/products/slim-phatty).

Includes:
- Factory preset names
- MIDI Control Channel names and values

Lots of the Control Channels in the 0-31 range support 14bit values for finer precision (using control channels 32-63 for LSB), but so far I was not able to tell Ardour that it should send out these higher precision values. All documentation or forum posts that I found online boils down to "14bit MIDI CC values feature is broken by design" (which doesn't really answer if it shouldn't be possible to support it nevertheless).

The `MIDINameDocument10.dtd` included with the `patchfiles` directory also documents possible `14bit`, `RPN`, `NRPN` values for the `Type` parameter, but it seems that neither value changes Ardour's behaviour. It always only sends out values in the 0-127 range on a single CC channel.

I also tried to extend the range of Values with:

```xml
<Control Type="14bit" Number="19" Name="VCF Cutoff">
  <Values Min="0" Max="16383"/>
</Control>
```

But since all of this didn't help, I now went for separate "Coarse" and "Fine" `7bit` controls for every control that supports higher precision values. Unless there is indeed a way to support `14bit` controls in Ardour right now, I'm happy to have this merged as it is.

### DIFF
--- a/patchfiles/Moog_Slim_Phatty.midnam
+++ b/patchfiles/Moog_Slim_Phatty.midnam
@@ -1,0 +1,489 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE MIDINameDocument PUBLIC "-//MIDI Manufacturers Association//DTD MIDINameDocument 1.0//EN" "http://www.midi.org/dtds/MIDINameDocument10.dtd">
+<MIDINameDocument>
+  <Author>Johannes Maibaum</Author>
+  <MasterDeviceNames>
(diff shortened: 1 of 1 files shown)


x42:
Thanks! rebased and merged as 6.0-pre0-2351 5963104378

PR #454, opened 2019-10-14
Made sense of Region > Trim > Trim to (range)

MrHeadwar:


gtk2_ardour/editor_ops.cc
```diff
@@ -3804,65 +3804,79 @@ Editor::trim_region (bool front)
...
 
-		RouteTimeAxisView* tav = dynamic_cast<RouteTimeAxisView*> (&rv->get_time_axis_view());
-		if (!tav) {
-			return;
-		}
+			rv->region()->clear_changes ();
 
-		samplepos_t start;
-		samplepos_t end;
+			if (rv->region()->first_sample() < loc.start() || (expand && rv->region()->first_sample() > loc.start())) {
```
x42: Instead of custom comparisions, it would be preferable to use Region::coverage()  and check for `OverlapNone`

With the previous code's switch statement it might be as easy as
```
case Evoral::OverlapNone:
  rv->region()->playlist()->remove_region (rv->region());
  continue;
```

MrHeadwar: It should have indeed. It relates to bug [7275](https://tracker.ardour.org/view.php?id=7275). In the current master, "Trim to [punch|loop]" does not trim any region if the region does not overlap both the range's start and end, hence a region that is partly in the range is not trimmed at all. Plus, a region that is selected and not in the range at all is left untouched, and I think the user should be trusted to know what he is doing, so if he trims a group of regions and one region is outside the range, this region should be destroyed. That is what this patch does.

MrHeadwar: I've also added a parameter ("expand") which is unused atm, that allows for a "fit to [loop|range]", cutting what exceeds the range and (trying to) expand the regions inside the range to fill the range.

PR #457, opened 2019-11-09
a-fluidsynth segfault: don't use midnam:update extension in hosts that do not support it

SpotlightKid:
Also logs a warning, if the hosts does not support the midnam or bankpatch extensions.

This allows this plugin to be safely used in other LV2 hosts, even when compiled with `LV_EXTENDED` defined.


x42:
rebased and merged as 6.0-pre0-2841 8fe978a8e5 -- Thanks!

PR #458, opened 2019-11-14
Add script to create LFO-like plugin automation

danielappelt:


x42:
This is awesome! Thanks
rebased and merged as 6.0-pre0-2860 5fb949267d

x42:
PS. It would be nice to have an Icon for this.

Also perhaps a call to `al:thin(20)` near the end (before the save undo call) may be helpful in case, there are very dense events for short regions with many repeats. Which could cause issues with per-sample automation.

I can look into both.

danielappelt:
> I can look into both.

This would be very cool!

PR #459, opened 2019-11-24
Make wscripts python3 compatible (waf 2.0.19 compatible)

dvzrv:
These changes allowed me to successfully build ardour using the most recent waf (2.0.19).
Some of them might be weird and will probably need refactoring now or later.

The [misc.py](https://gitlab.com/ita1024/waf/raw/waf-1.6.11/waflib/extras/misc.py) script is the only fragment from the currently used waf (1.6.11), that is *required* because of includes to build ardour (and when/if updating the compiled version of waf in this repo, it *has to be added*). However, its use should be refactored and removed after this pull request is merged.

All of this should still work with the current waf setup!
Note: On Arch Linux I'm using a [system-wide waf](https://www.archlinux.org/packages/community/any/waf/), but this workflow should work with the upstream waf just the same (after all, I'm only telling it where its components are through packaging).

Now, to how I made this build (after applying the changes in this pull request):

```
  # making sure, that subdirectories can be picked up with local includes
  touch __init__.py
  # making ancient 'misc' include available to system installed waf                        
  sed -e "s/('misc')/('misc', tooldir='tools')/" \                       
      -i {tools/luadevel,gtk2_ardour,session_utils,libs/fst,headless}/wscript             
  # make custom 'autowaf' include compatible with system installed waf
  find . -type f \                                      
         -iname "*wscript*" \                            
         -exec sed -e 's/from waflib.extras import autowaf/from tools import autowaf/g' \  
                   -e 's/import waflib.extras.autowaf/from tools import autowaf/g' \
                   -i {} \;
```

These should be all required changes to test this with a system installed waf and incorporating the old `misc.py`. The `autowaf.py` in use is still in place below `tools/`.
Aftwards the usual `waf configure`, `waf build` and `waf install` (+ parameters) applies.

### DIFF
--- a/gtk2_ardour/wscript
+++ b/gtk2_ardour/wscript
@@ -622,6 +622,8 @@ def build(bld):
         obj.source    = list(gtk2_ardour_sources)
         obj.target = 'ardour-' + str (bld.env['VERSION'])
         obj.includes = ['.']
+        obj.includes += bld.env['INCLUDES_LILV']
+        obj.ldflags = ['-l' + lib for lib in bld.env['LIB_LILV']]
         obj.ldflags = ['-no-undefined']
 
         if bld.is_defined('WINDOWS_VST_SUPPORT'):
--- a/gtk2_ardour/wscript
+++ b/gtk2_ardour/wscript
@@ -509,7 +509,6 @@ def build(bld):
         obj.source    = list(gtk2_ardour_sources)
         obj.target    = 'luadoc'
         obj.includes  = ['.', '../libs']
-        obj.ldflags   = ['-no-undefined']
         obj.use       = [
                 'libpbd',
                 'libardour',
@@ -549,7 +548,6 @@ def build(bld):
 #        obj.source    = [ 'ardour_button.cc', 'ui_config.cc', 'tooltips.cc' ]
 #        obj.target    = 'canvas_test'
 #        obj.includes  = ['.', '../libs']
-#        obj.ldflags   = ['-no-undefined']
 #        obj.use       = [
 #                'libpbd',
 #                'libardour',
@@ -624,7 +622,6 @@ def build(bld):
         obj.includes = ['.']
         obj.includes += bld.env['INCLUDES_LILV']
         obj.ldflags = ['-l' + lib for lib in bld.env['LIB_LILV']]
-        obj.ldflags = ['-no-undefined']
 
         if bld.is_defined('WINDOWS_VST_SUPPORT'):
         # Windows VST support mingw
--- a/libs/pbd/wscript
+++ b/libs/pbd/wscript
@@ -159,8 +159,7 @@ def build(bld):
         obj.source += 'debug_rt_alloc.c'
 
     obj.export_includes = ['.']
-    obj.includes = ['.']
-    obj.includes += bld.env['INCLUDES_XML']
+    obj.includes = ['.'] + bld.env['INCLUDES_XML']
     obj.name         = 'libpbd'
     obj.target       = 'pbd'
     obj.uselib       = 'GLIBMM SIGCPP XML UUID SNDFILE GIOMM ARCHIVE CURL'
--- a/libs/pbd/wscript
+++ b/libs/pbd/wscript
@@ -159,7 +159,7 @@ def build(bld):
         obj.source += 'debug_rt_alloc.c'
 
     obj.export_includes = ['.']
-    obj.includes     = ['.']
+    obj.includes = '. /usr/include/libxml2'
     obj.name         = 'libpbd'
     obj.target       = 'pbd'
     obj.uselib       = 'GLIBMM SIGCPP XML UUID SNDFILE GIOMM ARCHIVE CURL'
--- a/libs/pbd/wscript
+++ b/libs/pbd/wscript
@@ -159,7 +159,7 @@ def build(bld):
         obj.source += 'debug_rt_alloc.c'
 
     obj.export_includes = ['.']
-    obj.includes = '. /usr/include/libxml2'
+    obj.includes = ['.', bld.env['INCLUDES_XML']]
     obj.name         = 'libpbd'
     obj.target       = 'pbd'
     obj.uselib       = 'GLIBMM SIGCPP XML UUID SNDFILE GIOMM ARCHIVE CURL'
--- a/libs/pbd/wscript
+++ b/libs/pbd/wscript
@@ -159,7 +159,8 @@ def build(bld):
         obj.source += 'debug_rt_alloc.c'
 
     obj.export_includes = ['.']
-    obj.includes = ['.', bld.env['INCLUDES_XML']]
+    obj.includes = ['.']
+    obj.includes += bld.env['INCLUDES_XML']
     obj.name         = 'libpbd'
     obj.target       = 'pbd'
     obj.uselib       = 'GLIBMM SIGCPP XML UUID SNDFILE GIOMM ARCHIVE CURL'
--- a/libs/ptformat/wscript
+++ b/libs/ptformat/wscript
@@ -38,7 +38,7 @@ def build(bld):
         obj.cflags = [ bld.env['compiler_flags_dict']['pic'] ]
         
     obj.export_includes     = ['.']
-    obj.includes     = ['.']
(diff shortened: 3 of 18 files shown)


gtk2_ardour/wscript
```diff
@@ -509,7 +509,6 @@ def build(bld):
         obj.source    = list(gtk2_ardour_sources)
         obj.target    = 'luadoc'
         obj.includes  = ['.', '../libs']
-        obj.ldflags   = ['-no-undefined']
```
x42: What is the motivation to remove `-no-undefined` LDFLAG?

Did this sneak in by accident, or is there some relation to python3 compat?

gtk2_ardour/wscript
```diff
@@ -622,7 +620,8 @@ def build(bld):
         obj.source    = list(gtk2_ardour_sources)
         obj.target = 'ardour-' + str (bld.env['VERSION'])
         obj.includes = ['.']
-        obj.ldflags = ['-no-undefined']
+        obj.includes += bld.env['INCLUDES_LILV']
+        obj.ldflags = ['-l' + lib for lib in bld.env['LIB_LILV']]
```
x42: Does `obj.uselib += ['LILV']` help instead of this?

libs/pbd/wscript
```diff
@@ -159,7 +159,7 @@ def build(bld):
         obj.source += 'debug_rt_alloc.c'
 
     obj.export_includes = ['.']
-    obj.includes     = ['.']
+    obj.includes = '. /usr/include/libxml2'
```
x42: Absolute paths must not be used.

libs/ptformat/wscript
```diff
@@ -38,7 +38,7 @@ def build(bld):
         obj.cflags = [ bld.env['compiler_flags_dict']['pic'] ]
         
     obj.export_includes     = ['.']
-    obj.includes     = ['.']
+    obj.includes = '. /usr/include/glib-2.0 /usr/lib/glib-2.0/include'
```
x42: Ditto. Another absolute path, Prefer pkg-config or obj.uselib

x42:
I guess we can cherry-pick `misc.py` without the other changes, @pauldavisthefirst  what do you think?

gtk2_ardour/wscript
dvzrv: I was unable to compile with this added. I'm not entirely sure why that is. Potentially there was a concatenation problem.

libs/pbd/wscript
dvzrv: alright, will have a look at it!

gtk2_ardour/wscript
dvzrv: I'll check and update

dvzrv: Seems this didn't work, because:

`AttributeError: 'task_gen' object has no attribute 'uselib'`

libs/ptformat/wscript
dvzrv: `obj.uselib` does not work here.
I've now pushed the use of `bld.env['INCLUDES_GLIB']`, but lists may not be added to the includes list like that. Will have to use a for loop

gtk2_ardour/wscript
dvzrv: As context, this is what happens, when the those flags are still there:

```
[ 685/1125] Linking build/tools/bb/bb
g++: error: unrecognized command line option ‘-no-undefined’
```
The verbose output is:

```
Build failed
 -> task in 'bb' failed with exit status 1: 
        {task 107607835508512: cxxprogram bb.cc.1.o,gui.cc.1.o,misc.cc.1.o -> bb}
['/usr/bin/g++', '-pthread', 'tools/bb/bb.cc.1.o', 'tools/bb/gui.cc.1.o', 'tools/bb/misc.cc.1.o', '-o/home/dave/Downloads/ardour/build/tools/bb/bb', '-Wl,-Bstatic', '-Wl,-Bdynamic', '-Llibs/ardour', '-Llibs/midi++2', '-Llibs/evoral', '-Llibs/audiographer', '-Llibs/ptformat', '-Llibs/pbd', '-Llibs/temporal', '-lardour', '-lmidipp', '-levoral', '-laudiographer', '-lptformat', '-lpbd', '-ltemporal', '-ljack', '-lgtkmm-2.4', '-latkmm-1.6', '-lgdkmm-2.4', '-lgiomm-2.4', '-lpangomm-1.4', '-lglibmm-2.4', '-lcairomm-1.0', '-lsigc-2.0', '-lgtk-x11-2.0', '-lgdk-x11-2.0', '-lpangocairo-1.0', '-latk-1.0', '-lcairo', '-lgdk_pixbuf-2.0', '-lgio-2.0', '-lpangoft2-1.0', '-lpango-1.0', '-lgobject-2.0', '-lglib-2.0', '-lharfbuzz', '-lfontconfig', '-lfreetype', '-no-undefined']
```

session_utils/wscript
```diff
@@ -86,7 +86,7 @@ def build(bld):
     utils = bld.path.ant_glob('*.cc', excl=['example.cc', 'common.cc'])
 
     for util in utils:
-        fn = str(util)[:-3]
+        fn = os.path.basename(str(util))[:-3]
```
SpotlightKid: I would suggest to use `os.path.splitext(os.path.basename(str(util)))[0]` here.

(And maybe do `from os.path import basename, dirname, join, normpath, splitext, ...` at the top of the script.)

dvzrv:
It seems my changes somehow affect the way the `manifest.ttl` files of the example lv2 plugins are created:

```
* Node /build/ardour-git/src/ardour/build/libs/LV2/a-comp.lv2/manifest.ttl is created more than once. The task generators are:                                                                                                                                                                                                                                        
  1. bld(source=[], target='../../LV2/a-comp.lv2/manifest.ttl',
  meths=['apply_subst', 'check_err_features', 'check_err_order',
    'process_subst', 'process_rule'], features=['subst'],
    path=/build/ardour-git/src/ardour/libs/plugins/a-comp.lv2, idx=1,
    tg_idx_count=53, install_path='${LV2DIR}/a-comp.lv2', chmod=420,
    LIB_EXT='.so', _name='../../LV2/a-comp.lv2/manifest.ttl', posted=True,
    fun=<function subst_func at 0x686077fcdf80>, dict={'LIB_EXT': ''},
    install_task=None) in /build/ardour-git/src/ardour/libs/plugins/a-comp.lv2
  2. bld(source=[], target='../../LV2/a-comp.lv2/manifest.ttl',
  meths=['apply_subst', 'check_err_features', 'check_err_order',
    'process_subst', 'process_rule'], features=['subst'],
    path=/build/ardour-git/src/ardour/libs/plugins/a-comp.lv2, idx=1,
    tg_idx_count=53, install_path='${LV2DIR}/a-comp.lv2', chmod=420,
    LIB_EXT='.so', _name='../../LV2/a-comp.lv2/manifest.ttl', posted=True,
    fun=<function subst_func at 0x686077fcdf80>, dict={'LIB_EXT': ''},
    install_task=None) in /build/ardour-git/src/ardour/libs/plugins/a-comp.lv2
If you think that this is an error, set no_errcheck_out on the task instance
```

While `LIB_EXT='.so'` is setup properly, it somehow seems, that `dict={'LIB_EXT': ''}` overwrites it again. I can't seem to find where this is set...

session_utils/wscript
dvzrv: Fair point! I was just being pragmatic ;-)

session_utils/wscript
```diff
@@ -86,7 +86,7 @@ def build(bld):
     utils = bld.path.ant_glob('*.cc', excl=['example.cc', 'common.cc'])
 
     for util in utils:
-        fn = os.path.basename(str(util))[:-3]
+        fn = os.path.splitext(os.path.basename(str(util)))[0]
```
x42: session-utils must be C++ files with a "`.cc`" extension and compiled as such (see session_utils/README).

The `ant_glob` three lines above will only provide files with that end with ".cc". So it can never happen that the file extension will be something other than precisely 3 chars `.cc`.

That being said,  stripping the extension is fine regardless.

x42:
Cleaned up, rebased and applied with 6.0-pre0-3299 5a841ef462

PR #460, opened 2019-11-26
Fix thread-safety issue in a-fluidsynth.

swesterfeld:
This fix ensures that the a-fluidsynth "synth" object is not used in two threads at the same time during midi event handling (run() in RT thread vs. load_sf2() in worker thread), which could result in crashes.


x42:
Thanks. rebased and merged as 6.0-pre0-2926 be1012d64e

PR #464, opened 2020-01-20
tools: make ardour.keys generation reproducible

jelly:
By default perl hashes are random when iterating over them this makes
ardour not reproducible. As it ships with a random order generated
ardour.keys file, simply sorting the keys before iterating over them
makes the file reproducible.

Motivation: https://reproducible-builds.org


jelly:
Problem can be spotted here https://tests.reproducible-builds.org/archlinux/community/ardour/ardour-5.12-13-x86_64.pkg.tar.xz.html

pauldavisthefirst:
thanks! very clear explanation, just the way a PR should be.

PR #467, opened 2020-02-09
LCXL: Adopt pan knob behaviour

lentferj:
* "left/right" changed in A6, adopt for this
    * add a pick_up_rev function to honour this


x42:
Modified (fixed indenting and whitespace) and rebased as 6.0-pre0-3388 cc7eca7453

Thanks!

PR #468, opened 2020-02-12
Add Sun backend option for JACK

niacat:
JACK's Sun backend is used on NetBSD, illumos, GNU/HURD (?), others (?).

### DIFF
--- a/libs/backends/jack/jack_utils.cc
+++ b/libs/backends/jack/jack_utils.cc
@@ -68,6 +68,7 @@ namespace ARDOUR {
 	const char * const coreaudio_driver_name = X_("CoreAudio");
 	const char * const alsa_driver_name = X_("ALSA");
 	const char * const oss_driver_name = X_("OSS");
+	const char * const sun_driver_name = X_("Sun");
 	const char * const freebob_driver_name = X_("FreeBoB");
 	const char * const ffado_driver_name = X_("FFADO");
 	const char * const netjack_driver_name = X_("NetJACK");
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
I'd prefer to see the basic conditionals that make this available be wrapped in compile-time OS ident #ifdefs

niacat:
It's not really OS-specific since it's available through rump, but yeah I can do this.

krytarowski:
sun backend is OS independent.

pauldavisthefirst:
It is not OS independent. It uses ioctl's that are OS dependent.

krytarowski:
These ioctls can be rump syscalls (hypercalls). rump is a NetBSD kernel built as a literal standalone library. It can be executed in userland on Windows, Linux, Darwin, inside a bootloader in HURD etc. No need for real syscall ioctls. rump runs in theory on anything more or less with a C99/C11 compiler.

pauldavisthefirst:
That maybe true, but that would in turn require different build options for JACK on the non-Sun-ish/BSD host platform, since the configure time check is for the presence of sys/audioio.h (JACK1) and/or python's sys.platform saying "sunos" (JACK2).

Ardour's "JACK configuration" options are absolutely not intended to be all-encompassing. There are other JACK front-ends that can and should be used for unusual situations. If you're running on Linux, the "sun" driver for JACK is not a sensible option to present to users.

niacat:
In any case, can we get this merged? It's only going to show up on platforms where it's the primary audio API baked into the kernel, in the current state of this PR.

pauldavisthefirst:

PR #472, opened 2020-02-13
LCXL: make sends assignable for 32C

lentferj:
* add an option to the controller configuration gui
      to assign the 3x2 send banks in mixer mode
      to either the upper (1-6) or lower (7-12) Mixbus sends
    * This was a user request to better support the workflow
      of the CoMondo Mix system


x42:
Rebased and merged as 6.0-pre0-3409 d7fed7c52e  Thank you!

PR #474, opened 2020-02-23
Fixing midi out issue in Lua scripts

Rob8000:
The midi out ports weren't being iterated over correctly when running a Lua script with more than one midi out.

### DIFF
--- a/libs/ardour/luaproc.cc
+++ b/libs/ardour/luaproc.cc
@@ -725,12 +725,15 @@ LuaProc::connect_and_run (BufferSet& bufs,
 
 			// copy back midi events
 			if (_has_midi_output && lua_midi_sink_tbl.isTable ()) {
-				bool valid;
-				const uint32_t idx = out.get(DataType::MIDI, 0, &valid);
-				if (valid && bufs.count().n_midi() > idx) {
-					MidiBuffer& mbuf = bufs.get_midi(idx);
-					mbuf.silence(0, 0);
-					for (luabridge::Iterator i (lua_midi_sink_tbl); !i.isNil (); ++i) {
+				uint32_t count = 0;
+				uint32_t idx;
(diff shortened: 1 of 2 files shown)


x42:
I am torn. Not only does this change break existing scripts that use a single port, but it also breaks the assumption that by default there should only be a single MIDI port in a single route's data-flow.

We really should discourage using multiple MIDI ports. Most notably many other plugin standards only support for single MIDI ports, and MIDI synths only have a single input.  Ardour would replicate those plugins.

`dsp_run()` is intended to "make the common case fast", while  `dsp_runmap()` explses Ardour's plain method (with zero overhead). So if you must use multiple ports for one reason, the latter is likely to remain the only valid way.

Rob8000:
> I am torn. Not only does this change break existing scripts that use a single port, but it also breaks the assumption that by default there should only be a single MIDI port in a single route's data-flow.
> 
> We really should discourage using multiple MIDI ports. Most notably many other plugin standards only support for single MIDI ports, and MIDI synths only have a single input. Ardour would replicate those plugins.
> 
> `dsp_run()` is intended to "make the common case fast", while `dsp_runmap()` explses Ardour's plain method (with zero overhead). So if you must use multiple ports for one reason, the latter is likely to remain the only valid way.

I'm pretty sure that it doesn't break existing scripts. Scripts currently have access to a table representing the multiple midi output ports, but they can only use the first one as it is. The method that I changed was trying to iterate over all of the ports in this table, but only processing the first one. The change I made fixes that.

x42:
`Author: R <@gmail.com>`  what kind of attribution is that? :)

x42:
Rebased and merged as 6.0-pre0-3538 b10d9cf09b 

I've split up in two commits (code and example script:   8702ff2189665 b10d9cf09bf6150f0b), and fixed the example script: The line `seq = (seq + 1) % 2 + 1`always produced `1` (note-off events were never sent).

Cheers!

PR #476, opened 2020-02-24
And removing even more unused forward declarations.

chaot4:
Another minor cleaning up of unused forward declarations.


x42:
Rebased and merged as 6.0-pre0-3536 d1302ae934

PR #478, opened 2020-02-26
Reject HTTP connections to the websockets surface

lucianoiam:
According discussion on the IRC channel some users tried to connect to the websockets surface using a web browser. The surface currently does not serve any content over HTTP; to avoid confusion this patch detects HTTP connections and sends back a meaningful error message.

### DIFF
--- a/libs/surfaces/websockets/server.cc
+++ b/libs/surfaces/websockets/server.cc
@@ -411,9 +411,11 @@ WebsocketsServer::lws_callback (struct lws* wsi, enum lws_callback_reasons reaso
 		case LWS_CALLBACK_UNLOCK_POLL:
 		case LWS_CALLBACK_WS_PEER_INITIATED_CLOSE:
 		case LWS_CALLBACK_FILTER_HTTP_CONNECTION:
+#if LWS_LIBRARY_VERSION_MAJOR >= 3
 		case LWS_CALLBACK_HTTP_BIND_PROTOCOL:
 		case LWS_CALLBACK_ADD_HEADERS:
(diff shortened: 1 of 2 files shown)


x42:
rebased and merged as 6.0-pre0-3542 ef512d412d
Thanks again!

Are you planning to allow to serve some files (e.g. `share/websockets_client/`) in the future?

PS. The message says "Refer to Ardour documentation for further information."  Did you just volunteer to write that doc? ;-p

lucianoiam:
Haha! sure, why not? the protocol is rather simple; will work on docs later.

I am not planning to serve files in the short term; I primarily wrote this to plug Ardour into a separate project that already includes a web server. But it is an interesting addition that could become a priority if someone writes a decent web interface. I would add zeroconf support so the whole thing “just works” by enabling the surface.

The current surface feature set is limited to manipulate Ardour mixer strips so probably any derivative work will orbit around the idea of a mixer, unless more “nodes” are added.

I do plan to write some client tools, in fact I have an almost complete Python client for the surface which makes it easy to interact with the mixer from a script (is this of interest for the project? can create a PR later).  Next will be a collection of mixer-related widgets. Still not sure what tech to pick up, if it’s HTML/JS then it could make its way into an official web UI served directly by Ardour. The other candidate I am considering is Flutter since it now also targets the web. A Flutter package could potentially allow to create multiplatform implementations of Ardour “custom micro-surface apps” very easily. Well, at least mixer-tasks related apps.

Summarizing maybe I will never write an “official web UI” but I am happy to contribute with any reusable tools that result from my own work, like the python scripts or widgets.

PR #479, opened 2020-03-01
Add Python and Flutter/Dart client and demos for the WebSockets server surface

lucianoiam:
This aims to make it easy to interact with the Ardour mixer from a script or mobile app.

More stuff on the way...


x42:
I'm not sure if adding all that python code directly to upstream ardour/git is a good idea.
That'll be a lot to maintain and also tie release cycles to Ardour's. 
Most control surface frontends are in separate repositories for those reason.

Then again it'd be nice to have ardour's surface serve some client via http.
One approach would be to bundle it when packaging ardour (like 3rd party plugins are bundled).

lucianoiam:
All these makes sense, I will better move all these to a separate repo. 

Regarding the hypothetical web client, is a basic web-based mixer a good candidate for an official release?

x42:
On 3/1/20 11:37 PM, Luciano Iam wrote:

> Regarding the hypothetical web client, is a basic web-based mixer a good candidate for an official release?

Yes, I expect so.

Given time that could also grow into a more complete mixer interface,
similar to https://openstagecontrol.ammd.net/ (which uses ardour's OSC
surface).

Another candidate could be some transport-control for small screens
(record control from a phone).

The top-level HTTP URL could list installed web-clients, maybe from
Ardour's config-dir e.g. `~/.config/ardour6/web-clients/*/` in addition
to some bundled clients.

What do you think?

lucianoiam:
I like the idea of a transport control in a phone, it would turn the WS surface into something useful for end users now.

Then it could grow over time into a more complex UI like you said. I wasn't aware of the Open Stage Control project, it's great! I can borrow some stuff from there.

I think implementing a list of available clients is a great idea because it wouldn't limit the surface to a single "official" version. Does Ardour already define a directory for bundled content? thanks.

x42:
On 3/2/20 10:33 AM, Luciano Iam wrote:

> Does Ardour already define a directory for bundled content?
So far 3rd party bundled content are only LV2 plugins (and demo
sessions, Mixbus only). They're in specific folders e.g. [1].

I think we should follow what generic MIDI control surface does for midi
maps [2]: share/midi_maps/ and USER-CONFIG-DIR/midi_maps/


How about `share/web-surface/*/` and `USER-CONFIG-DIR/web_surface/*/` ?

so `.../web-surface/UNIQUE-ID-OF-THE-CTRL/<stuff here>`

maybe along with some mandatory meta-data file:
 `.../web-surface/UNIQUE-ID-OF-THE-CTRL/.description`



[1]
https://github.com/Ardour/ardour/blob/master/tools/linux_packaging/build#L725

[2]
https://github.com/Ardour/ardour/blob/master/libs/surfaces/generic_midi/generic_midi_control_protocol.cc#L181-L204

lucianoiam:
Thanks for pointing out the examples. I agree with naming as it follows what has been done for another surface.

I will create new PRs as I make progress; I think this one can be closed now. For anyone interested in this patch, I moved the files to the places where they belong:

PyPi
`https://pypi.org/project/pyardourclient/0.1.1/`

Dart pub
`https://pub.dev/packages/dart_ardour_client/versions/0.1.0`

PR #482, opened 2020-03-28
Initialize keyboard layout based on user preference

ngeiswei:
Otherwise the Simple QWERTY layout is loaded in the Virtual MIDI Keyboard
regardless of what is set in

Preferences->MIDI->Virtual Keyboard->Virtual Keyboard Layout

### DIFF
--- a/gtk2_ardour/virtual_keyboard_window.cc
+++ b/gtk2_ardour/virtual_keyboard_window.cc
@@ -51,7 +51,7 @@ VirtualKeyboardWindow::VirtualKeyboardWindow ()
 
 	_piano.set_flags (Gtk::CAN_FOCUS);
 
-	_piano.set_keyboard_layout (APianoKeyboard::S_QWERTY);
+	select_keyboard_layout (UIConfiguration::instance().get_vkeybd_layout ());
 	_piano.set_annotate_octave (true);
 	_piano.set_grand_piano_highlight (false);
 	_piano.set_annotate_layout (true);


x42:
Odd, here it works  due to `ParameterChanged`. 

I'm also surprised that the proposed change works for you, the window gets constructed from ArdourUI' constructor. That is usually too early for `UIConfiguration`.

In other places this is sometimes handled by explicitly calling `parameter_changed()` to trigger a change.

ngeiswei:
For having placed `printf` in VirtualKeyboardWindow::parameter_changed`, I can tell that it is not called upon the creation of the Virtual keyboard, which, for having placed `printf` in `VirtualKeyboardWindow::VirtualKeyboardWindow`, I can tell that it is created when the user clicks on `Window->Virtual Keyboard`, much after `UIConfiguration` has been instantiated, I suppose.

I tried with a pristine config (i.e. erased `~/.config/ardour*`), with the master head (rev 6a930c81eb) and I'm having this problem, so something happening wrongly on my system (Manjaro 19.0.2).

ngeiswei:
> In other places this is sometimes handled by explicitly calling parameter_changed() to trigger a change.

OK, so, @x42 , you're saying that, if a fix of that nature is required, the following

```c++
select_keyboard_layout (UIConfiguration::instance().get_vkeybd_layout ());
```

should be replaced by

```c++
parameter_changed ()
```

right?

ngeiswei:
For what it's worth, my Desktop Environment is Xfce.

x42:
I'm sorry, scratch what I have said. `WM::ProxyWithConstructor virtual_keyboard_window` is indeed not immediately constructed.

Your original commit is fine. rebased and merged as 6.0-pre1-22 4001ed7515. Thanks again!

PR #483, opened 2020-04-04
Move MIDI Virtual Keyboard layout in its own class

ngeiswei:
This PR doesn't change anything behaviorally for the MIDI Virtual Keyboard, only moves the layout code in its own class.

The main motivation for doing that is to have the tracker interface reuse the MIDI Virtual Keyboard layout. I initially tried to have the tracker interface use the whole MIDI Virtual Keyboard, not just its layout, but it didn't work out because of numeric keys (edit values like velocity, etc) clashing with note keys for non single layouts.

I know you are trying to get Ardour 6.0 out and that sort of change is probably unwelcome at the moment. And of course there is no urge in merging it (my tracker interface is best taken care of post-Ardour 6.0), but I wanted to put it out now before someone decides to make substantial changes to the MIDI Virtual Keyboard code.

If you'd rather not merge it at all, ever, I would move that code to the tracker side, but I think from a programing design viewpoint, it's better to avoid duplication when possible. The class split has also a small intrinsic value I would argue as well.

Thanks for your understanding!

### DIFF
--- a/gtk2_ardour/keyboardlayout.cc
+++ b/gtk2_ardour/keyboardlayout.cc
@@ -32,7 +32,7 @@ KeyboardLayout::~KeyboardLayout ()
 }
 
 void
-KeyboardLayout::set_keyboard_layout (Layout layout)
+KeyboardLayout::set_layout (Layout layout)
 {
 	switch (layout) {
 		case QWERTY:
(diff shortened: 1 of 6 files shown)


x42:
Looks good from a cursory reading. The only thing I'm worried about is having a symbol
`KeyboardLayout` in the global namespace.   We should rename that.

x42:
Rebased and merged as 6.0-pre1-177 1eb98316a3
I've also `s/KeyboardLayout/PianoKeyBindings/` to prevent polluting the global namespace

Cheers!

ngeiswei:
Thank you so much!

(didn't have to respond or help earlier due to being busy, thanks for handling the whole thing).

PR #484, opened 2020-04-06
WebSockets: add support for static http content

lucianoiam:
Implements a http file server to access web surfaces from a browser; relies on the same libwebsockets so no new dependencies are required. Includes a dynamically generated index page available at `http://localhost:9000` that lists built-in and user defined surfaces.


x42:
Wow, that's nice! Thanks for continuing development on that front!

I think this should be 3-or 4 commits:
 * NO-OP rename header guards
 * Update html/css websocket frontend
 * Implement websocket http-server
maybe another one? cleanup/prepare for http

What do you think?  I can have a go at it later, unless you're faster.

libs/surfaces/websockets/server.cc
```diff
@@ -59,9 +60,29 @@ WebsocketsServer::WebsocketsServer (ArdourSurface::ArdourWebsockets& surface)
...
+	// list of surfaces is available as a dynamically generated json file
+	memset (&_lws_mnt_index, 0, sizeof (lws_http_mount));
+	_lws_mnt_index.mountpoint      = "/";
+	_lws_mnt_index.mountpoint_len  = strlen (_lws_mnt_index.mountpoint);
+	_lws_mnt_index.origin_protocol = LWSMPRO_FILE;
+	_lws_mnt_index.origin          = _resources.index_dir ().c_str ();
+
+	// user defined surfaces in the user config directory
+	memset (&_lws_mnt_user, 0, sizeof (lws_http_mount));
+	_lws_mnt_user.mountpoint      = "/user";
```
x42, line 75: Isn't there a trailing slash missing?  `/user` (file) vs  `/user/` (dir)

Instead of `/user/` perhaps use `/~/` ? To avoid hardcoding English?

libs/surfaces/websockets/server.cc
```diff
@@ -279,35 +312,100 @@ WebsocketsServer::write_client (Client wsi)
...
-	const char *html_body = "<p>This URL is not meant to be accessed via HTTP; for example using"
-		" a web browser. Refer to Ardour documentation for further information.</p>";
-	lws_return_http_status (wsi, 404, html_body);
+	std::string index = _resources.scan ();
+
+	char body[MAX_INDEX_SIZE];
+	//lws_strncpy (body, index.c_str (), sizeof(body));
+	memset (body, 0, sizeof (body));
+	strncpy (body, index.c_str (), sizeof(body) - 1);
+	int len = strlen (body);
```
x42, line 394: Why is this needed?

Can't you just use `index.c_str(), index.length()` ?

lucianoiam:
Regarding the split commits yes I agree, this PR does too much. Should I open different PRs or reopen a new PR with several commits? pardon my lack of GitHub etiquette.

libs/surfaces/websockets/server.cc
lucianoiam: You are right, directories should be represented using a trailing slash.

The `/~/` suggestion is cleaner but note there is also a directory for serving built-in surfaces (residing within Ardour's data directory) hardcoded as `/builtin/`. This is not visible in code but in the filesystem (see `{GIT_ROOT}/share/web_surfaces/builtin`). Originally the code implemented a third mount point for these but later I decided that was unneeded complexity because the `/` mount point can already serve `/builtin/` by relying on the filesystem structure. Web clients will make requests to either `/~/...` or `/builtin/...` depending the surface location on disk. So there is a further English hardcoded string. What do you think?

libs/surfaces/websockets/server.cc
lucianoiam: True, no need to copy that string. I guess I copied this from the pure websockets data write part where you need to provide a buffer with initial padding, and didn't think too much. Good catch.

lucianoiam:
I accidentally clicked "re-request" review

x42:
> Should I open different PRs or reopen a new PR with several commits?

Whatever is easiest for you.
Since you're currently still the only author on this branch, you could even force-push.

libs/surfaces/websockets/server.cc
x42: When reading over it, I thought  `/` is used for bundled/built-in content, and `~/` for per-user surfaces.

Anyway this is not a blocker, for this PR. Keeping `/user/` is fine with me.
If needed, we can still change this later (pre-release) after further discussion.

libs/surfaces/websockets/server.cc
lucianoiam: Just pushed separate commits. Was going to address this in a further new commit then realized that `lws_write()` wants a pointer to mutable data and `c_str()` returns const. What's the recommended strategy there? a C cast (in the case of HTTP writes it can be assumed that libwebsockets does not touch the buffer) or blindly adhere to the API and keep the copy?

libs/surfaces/websockets/server.cc
lucianoiam: Seems libwebsockets does not like the trailing slash; appending it breaks the links to user defined data (404). Cannot find docs for the expected format or examples where the mount point is not the trivial "/". Guess it is not a blocker as well.

https://libwebsockets.org/lws-api-doc-master/html/structlws__http__mount.html#aa2391bfcada0b7a290b3c6651f64586c 

https://github.com/warmcat/libwebsockets/blob/master/minimal-examples/http-server/minimal-http-server/minimal-http-server.c

x42:
Thanks that looks, good. I'll merge it.

I get a few compiler warnings, but those can be addressed separately:

```
../libs/surfaces/websockets/server.cc:364:29: warning: ignoring return value of ‘int lws_add_http_header_status(lws*, unsigned int, unsigned char**, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_add_http_header_status (wsi, HTTP_STATUS_OK, &p, end);
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc:365:31: warning: ignoring return value of ‘int lws_add_http_header_by_token(lws*, lws_token_indexes, const unsigned char*, int, unsigned char**, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_add_http_header_by_token (wsi, WSI_TOKEN_HTTP_CONTENT_TYPE,
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    reinterpret_cast<const unsigned char*> ("application/json"), 16, &p, end);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc:367:31: warning: ignoring return value of ‘int lws_add_http_header_by_token(lws*, lws_token_indexes, const unsigned char*, int, unsigned char**, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_add_http_header_by_token (wsi, WSI_TOKEN_CONNECTION,
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
    reinterpret_cast<const unsigned char*> ("close"), 5, &p, end);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc:369:31: warning: ignoring return value of ‘int lws_add_http_header_by_token(lws*, lws_token_indexes, const unsigned char*, int, unsigned char**, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_add_http_header_by_token (wsi, WSI_TOKEN_HTTP_CACHE_CONTROL,
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    reinterpret_cast<const unsigned char*> ("no-store"), 8, &p, end);
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc:371:27: warning: ignoring return value of ‘int lws_finalize_http_header(lws*, unsigned char**, unsigned char*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_finalize_http_header (wsi, &p, end);
  ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc: In member function ‘int WebsocketsServer::send_index_body(Client)’:
../libs/surfaces/websockets/server.cc:400:33: warning: ignoring return value of ‘int lws_http_transaction_completed(lws*)’, declared with attribute warn_unused_result [-Wunused-result]
  lws_http_transaction_completed (wsi);
```

x42:
Rebased and merged as 6.0-pre1-229 891c63fe89

x42:
oops

```
 mounting file://./../share:./../build:./../gtk2_ardour:./../build/gtk2_ardour:./web_surfaces
```

lucianoiam:
Never tried `ardev` before now I see what's happening.

Is updating `ardev_common_waf.sh` an acceptable solution? `resources.cc` already reads an environment variable `ARDOUR_WEBSURFACES_PATH` that overrides `ARDOUR_DATA_PATH` when present. Akin to what is done for the midi surface.

x42:
Already fixed in be023ef32d9eb2ed7

PR #485, opened 2020-04-10
Web surfaces: prevent never-ending load of index.json

lucianoiam:
61692ae1f8de1b40100bffb650e2112274a7bab5 causes a timeout on HTTP clients when serving `index.json` (line 402), this PR addresses that and makes these kind of errors visible in the web-based index


lucianoiam:
Added some commits that improve the websockets surface resources.

- Now the locations of surfaces on disk are displayed in the index page
- There is a new shared directory in the `web_surfaces` data path that will contain useful resources like a simple websockets JS client. Currently the demos interact with the WS stream at a low level, having a simple official client like here https://github.com/Ardour/ardour/pull/479#issuecomment-593916294 but for JS should encourage development of new surfaces including the upcoming simple Transport control
- The surface data directory has been cleaned up, the index page resource files were moved to a single subdirectory to avoid clutter

x42:
Thanks. This is coming along nicely!
Rebased and merged as 6.0-pre1-281 68ecf6c0b3

PR #486, opened 2020-04-12
Add a JavaScript client library for the WebSockets surface

lucianoiam:
The library is intended to be used by the built-in web surfaces and also available for the user defined ones.

All layers in the WebSockets component are now complete and ready for creating stuff that can be used in the real world.

Also included:

- Better index page design
- Surfaces got a version attribute
- Minor improvements here and there

### DIFF
--- a/libs/surfaces/websockets/manifest.cc
+++ b/libs/surfaces/websockets/manifest.cc
@@ -54,10 +54,12 @@ SurfaceManifest::SurfaceManifest (std::string path)
 			_name = value;
 		} else if (name == "Description") {
 			_description = value;
+		} else if (name == "Version") {
+			_version = value;
 		}
 	}
 
-	if (_name.empty () || _description.empty ()) {
+	if (_name.empty () || _description.empty () || _version.empty ()) {
 #ifndef NDEBUG
 		std::cerr << "SurfaceManifest: missing properties in " << xml_path << std::endl;
 #endif
@@ -76,6 +78,7 @@ SurfaceManifest::to_json ()
 		<< "\"path\":\"" << Glib::path_get_basename (_path) << "\""
 		<< ",\"name\":\"" << _name << "\""
 		<< ",\"description\":\"" << _description << "\""
+		<< ",\"version\":\"" << _version << "\""
 		<< "}";
 
 	return ss.str ();
--- a/libs/surfaces/websockets/manifest.h
+++ b/libs/surfaces/websockets/manifest.h
@@ -24,7 +24,7 @@
 class SurfaceManifest
 {
 public:
-	// all ardour control surfaces store presets in xml format
+	// all ardour control surfaces implement presets using xml format
 	SurfaceManifest (std::string);
 
 	bool valid () { return _valid; }
@@ -32,6 +32,7 @@ public:
 	std::string path () { return _path; }
 	std::string name () { return _name; }
 	std::string description () { return _description; }
+	std::string version () { return _version; }
 
 	std::string to_json ();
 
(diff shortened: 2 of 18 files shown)


pauldavisthefirst:
This is an aside/non-central comment, but we need to think about how/if to try to unify the accessible "command space" that OSC and websocket support can address.

x42:
```
error: Failed to merge in the changes.
Patch failed at 0001 WebSockets: fix timeout when sending HTTP body
```

x42:
It'd be great if you could rebase your changes, not "Merge https://github.com/Ardour/ardour"
There should not be any merge commits on ardour's master branch.

lucianoiam:
I agree both surfaces should not diverge too much. I created this mainly for supporting mixer-like functionality; the current OSC surface goes beyond that. Code is clean enough to add/modify commands once a set is defined.

lucianoiam:
I will need to spend some time with the Git Book; the merge commits are gone from the history but not sure about the conflicts above. They look all the same:

```
<<<<<<< master
	if (_name.empty () || _description.empty () || _version.empty ()) {
=======
	if (_name.empty () || _description.empty ()) {
>>>>>>> master
```

The uppermost option is the correct for all files but then committing the changes creates a new merge commit, all using the GitHub interface.

lucianoiam:
> It'd be great if you could rebase your changes, not "Merge https://github.com/Ardour/ardour"
> There should not be any merge commits on ardour's master branch.

Fixed

x42:
Nice work. +1 for using the PBD::* API, too.
This can now be merged/rebased cleanly. The conflict is gone. Thanks!

Looks like you're still busy pushing commits to this PR, is there more coming, or should I merge it?

lucianoiam:
Thanks! it is safe to merge, I will be silent for a couple of days.

x42:
Rebased and merged as 6.0-pre1-330 83353b3125

lucianoiam:
@pauldavisthefirst this is a summary of the commands supported by the WebSockets surface:

```
tempo                       (bpm: double)
position_time               (seconds: double)
transport_roll              (on_off: bool)
record_state                (on_off: bool)
strip_gain [strip_id: int]  (db: double)
strip_pan [strip_id: int]   (position: double)
strip_mute [strip_id: int]  (on_off: bool)
strip_plugin_enable [strip_id: int, plugin_id: int] (on_off: bool)
strip_plugin_param_value [strip_id: int, plugin_id: int, param_id: int] (value: double)
```

These represent read-only states and are used for feedback:

```
strip_meter                      (db: double)
strip_description[strip_id: int] (name: string)
strip_plugin_description [strip_id: int, plugin_id: int] (name: string)
strip_plugin_param_description [strip_id: int, plugin_id: int, param_id: int] (name: string, value_type: char, min: int, max: int, is_log: bool)
```

Using the WS surface way of naming things the format is read like:

`node [address] (values)`

Regarding a potential command-space unification, do the above interfaces differ too much from the current OSC? I would be happy to introduce some modifications to keep things uniform if necessary. There is already a difference in how the various object identifiers (strip_id, plugin_id ,…) are treated separately from the arguments/values and that cannot be changed (this is not OSC after all) but the values’ order/count/ranges could be revised. Or the command names themselves.

For example the pan handles an artificial range [-1;1] just for the sake of symmetry. There is already a code comment about that (strips.cc @ 94).

Also probably the lower level interaction with the internal Ardour objects should be revised, I grabbed inspiration from the existing surfaces but some choices might be odd.

pauldavisthefirst:
This would be my goto comparison point:

https://github.com/Ardour/ardour/blob/master/libs/surfaces/osc/osc.cc#L425

I'm not as familiar with the "route observer" side of our existing OSC support, Len Ovens has worked on that more.

PR #487, opened 2020-04-16
Fix potential crash in WebSockets Server surface

lucianoiam:
- Fixed inbound messages that were not completely validated and led to crashes
- libwebsockets < 3 does not include MIME type for `.svg` files and results in 404 for those... SVG could be used for serving graphical assets －fixed
- Improve JS client and update mixer-demo


x42:
rebased as  6.0-pre1-394 28dd253231

PR #489, opened 2020-04-19
Add a web surface for basic transport control

lucianoiam:
Currently it implements transport roll on/off, toggling record state and position time feedback.

Regarding transport roll there is a single play/pause button that tries to mimic the spacebar behavior. There is a comment in BasicUI code that encourages calling `toggle_roll()` to achieve this:

https://github.com/Ardour/ardour/blob/ef206d80667f5fe1bfa68dd569c5f3a207fb7f44/libs/surfaces/control_protocol/control_protocol/basic_ui.h#L108 

However:

`toggle_roll(true)`  →  first call works (starts rolling), subsequent fail (does not stop)
`toggle_roll(false)`  →  always works, but sometimes the playhead is reset to the starting point when playback stops

The spacebar does not seem to reset the playhead location when transport stops. Either there is a bug in the WS surface or `toggle_roll()` does not work as advertised.

How to perfectly replicate the spacebar action?


x42:
Rebased and merged as  6.0-rc1-30 744762f343

PR #490, opened 2020-04-21
Mark user-visible messages for translation

prokoudine:
Just a handful of forgotten user-visible messages, some of which are extremely visible :)


x42:
Rebased and merged as 6.0-rc1-36 47e4216012

PR #491, opened 2020-04-21
Windows related bugfixes for the WebSockets Server

lucianoiam:
Also adds the ability to interact with VCA strips


lucianoiam:
The Windows build still has issues when serving files as of 6.0 RC1-44. Assuming that build includes commit cc08a2d in this PR

Unless index.html is appended to each surface URL a 404 is returned. This is handled by libwebsockets and never happened on other platforms.

Probably needs dedicated debugging on a Windows environment.

win

ubuntu

mac

lucianoiam:
The "Escape path strings in surfaces.json" patch was also needed and worked, though.

PR #492, opened 2020-04-22
Honour forward/rewind option when already rolling #8031

mister-benjamin:
The options ForwardSlow, Forward and ForwardFast and their
respective Rewind options were not honoured in the execution
of forward/rewind operations when the transport was already
rolling at a slower speed than the forward or rewind option's
speed with same direction as the requested direction.


pauldavisthefirst:

PR #493, opened 2020-04-22
Fix ardour.js client library

lucianoiam:
Bugs did not affect shipped surfaces but made control APIs in the JS client unusable. Also added some comments.


x42:
Nice clean git commit history! rebased and merged 6.0-rc1-55 ea16fc51d3

PR #494, opened 2020-04-22
Update Russian translation

prokoudine:
This isn't a complete translation but it's _way_ better than what would get into the final release. Another update is possible depending on how soon the release is :)


x42:
Rebased and merged as 6.0-rc1-74 2aa5e9029
Спасибо!

PR #496, opened 2020-04-25
Recognize dist, distcheck better to set app name

nphilipp:
Previously, `dist` had to be the first argument to waf. If it came in a
different position (e.g. `APPNAME=Ardour ./waf tarball dist`) or
`distcheck` was used instead of `dist`, the resulting archive was named
`noname-....tar.bz2`.

Signed-off-by: Nils Philippsen <nils@tiptoe.de>


pauldavisthefirst:

PR #497, opened 2020-04-26
Update Japanese translation for 6.0

inatus:
I have updated the Japanese translation files towards the version 6.0 release. It would be appreciated if you could confirm and merge this.


x42:
Since I don't speak Japanese, I cannot confirm the translation..
But it's rebased and merged 6.0-rc1-138 cce6ac3e73 now.
Thank You!

PR #498, opened 2020-04-27
Fix default midi patch model

ngeiswei:
I noticed control change names were replaced by `Controller N`, as shown in the screenshot below (from 24d3bf25a9)

before-fix

That PR introduces a possible fix, see below

after-fix

I obtained that fix more from trial and error than deep understanding of the code, so be careful.

### DIFF
--- a/gtk2_ardour/midi_time_axis.cc
+++ b/gtk2_ardour/midi_time_axis.cc
@@ -440,7 +440,7 @@ MidiTimeAxisView::setup_midnam_patches ()
 		model_changed ("");
 	} else if (model.empty() || ! MIDI::Name::MidiPatchManager::instance ().master_device_by_model (model)) {
 		/* invalid model, switch to use default */
-		model_changed ("");
+		model_changed (DEFAULT_MIDNAM_MODEL);
 	} else {
 		model_changed (model);
 	}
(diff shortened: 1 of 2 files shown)


gtk2_ardour/midi_time_axis.cc
```diff
@@ -437,10 +437,10 @@ MidiTimeAxisView::setup_midnam_patches ()
 	std::string model = gui_property (X_("midnam-model-name"));
 	if (model.empty() && _route->instrument_info().have_custom_plugin_info ()) {
 		/* use plugin's MIDNAM */
-		model_changed ("");
```
x42, line 440: This is intentional.  Empty means to use default.

x42:
When a plugin provides a MIDNAM, Ardour only shows names that the plugin provides.
(and not generic midi)

ngeiswei:
> When a plugin provides a MIDNAM, Ardour only shows names that the plugin provides.
(and not generic midi)

That's really cool! I don't understand how the plugin provides that though, it doesn't seem to be in the lv2 spec, and I couldn't find a single midnam file among my installed plugins. Or do you mean that it must be included in Ardour's patchfiles folder?

ngeiswei:
>  I don't understand how the plugin provides that [MIDNAM] though

Never mind, I found a plugin I can study to understand that http://x42-plugins.com/x42/x42-avldrums :-)

ngeiswei:
I pushed something but it doesn't work. Will resume work on that later...

x42:
Can you check if a47fac7638dc69051ff4778e08dfccef539195bb solves this issue?

As for the LV2 MIDNAM extension,  a-fluidsynth, sfizz or drumgizmo may be better examples.
the avldrums embedd a fixed hardcoded midnam.

ngeiswei:
Yep a47fac7 solves the issue. Doesn't interfere with the plugin midinam either. All seems well, thanks!

PR #499, opened 2020-05-08

cooltehno:
This Blueberry theme is totally newly generated from the 'dark-ardour.colors'-file in order to match maximum to the default theme structure. The new conception is more contrasting&bright.
A 4min video show:
https://vimeo.com/416166712


ahellquist:
Looks really slick. I have never been able to use the alternate themes myself but this theme might be possible to use after all. I my opinion, Ardour needs a little spicing and theming to attract more users. It looks like it always did and that is not necessarily wrong but it isn't necessarily right either :-)

cooltehno:
@ahellquist  Thanks)

cooltehno:
I've closed it because in my changes I've deleted all the comments (in color definition lines). Also need some checking about some "alert: greenish" color.

PR #500, opened 2020-05-09
Compatibility with libwebsockets 4

lucianoiam:
The WebSockets Server surface did not work when compiling against lws 4. This is a simple patch that fixes that.

~~I plan to better integrate lws by leveraging its built-in glib event loop integration, new in 4.0.0
https://github.com/warmcat/libwebsockets/blob/224b59c00b134f3aa5d6f10fb833c9ab30720352/changelog~~

Leverages libwebsockets built-in glib loop integration.

### DIFF
--- a/libs/surfaces/websockets/server.cc
+++ b/libs/surfaces/websockets/server.cc
@@ -38,6 +38,12 @@
 #endif
 #endif
 
+/* libwebsockets version of LWS_LIBRARY_VERSION_NUMBER also appends
+   LWS_LIBRARY_VERSION_PATCH which can contain non-numeric characters
+   rendering it unsuitable for numeric version checks  */
+#define LWS_LIBRARY_VERSION_NUM (LWS_LIBRARY_VERSION_MAJOR * 1000000) + \
+								(LWS_LIBRARY_VERSION_MINOR * 1000)
+
 #define MAX_INDEX_SIZE	65536
 
 using namespace Glib;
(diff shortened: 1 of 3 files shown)


x42:
Fails to compile (debian/stable)
```
../libs/surfaces/websockets/server.cc: In member function ‘virtual int WebsocketsServer::start()’:
../libs/surfaces/websockets/server.cc:121:21: error: ‘LLL_USER’ was not declared in this scope
  lws_set_log_level (LLL_USER | LLL_ERR | LLL_WARN | LLL_DEBUG, 0);
                     ^~~~~~~~
../libs/surfaces/websockets/server.cc:121:21: note: suggested alternative: ‘LLL_PARSER’
  lws_set_log_level (LLL_USER | LLL_ERR | LLL_WARN | LLL_DEBUG, 0);
                     ^~~~~~~~
                     LLL_PARSER
../libs/surfaces/websockets/server.cc: In static member function ‘static int WebsocketsServer::lws_callback(lws*, lws_callback_reasons, void*, void*, size_t)’:
../libs/surfaces/websockets/server.cc:404:9: error: ‘lws_callback_http_dummy’ was not declared in this scope
    rc = lws_callback_http_dummy (wsi, reason, user, in, len);
         ^~~~~~~~~~~~~~~~~~~~~~~
../libs/surfaces/websockets/server.cc:404:9: note: suggested alternative: ‘lws_callback_function’
    rc = lws_callback_http_dummy (wsi, reason, user, in, len);
         ^~~~~~~~~~~~~~~~~~~~~~~
         lws_callback_function
```

lucianoiam:
Seems lws < 2.1.0 does not define the `LLL_USER` log level; I removed its usage since it is not really necessary.

Using `lws_callback_http_dummy()` avoids manually handling lots of cases in the libwebsockets callback, something that caused trouble across different lws versions and configurations. I copied this approach from the lws examples, but seems < 2.1.0 does not implement this function. In such case the previous approach is used, that is manually handle more cases in the callback `switch`.

lucianoiam:
Just in case, the latest two commits should fix the build issue. Forgot to mention in previous comment.

x42:
Rebased and merged as 6.0-25 0a19e87646

x42:
This breaks Windows at run-time. The following message is printed twice:

> ArdourWebsockets: check your libwebsockets was compiled with LWS_WITH_GLIB or LWS_WITH_EXTERNAL_POLL enabled.

x42:
It used to work on windows XP, Windows 7 and Window 10 with libwebsockets 3.0.99 before this change.

x42:
I'll see if we can update to 0.4.15. we'll also need to compile it `-DLWS_WITH_GLIB=YES` ?

lucianoiam:
Only `LWS_WITH_GLIB` is needed. The other flag `LWS_WITH_EXTERNAL_POLL` is not recommended anymore https://github.com/warmcat/libwebsockets/blob/2a772776a92743172191af271345d9d1059d8658/CMakeLists.txt#L229, it was required to enable a custom integration with Glib when `LWS_WITH_GLIB` was not available in libwebsockets (pre 4x.).

lucianoiam:
Here is a patch that avoids updating the Windows library, it replaces the compile time feature check with the previous runtime check:

```
--- ./libs/surfaces/websockets/server.cc	2020-06-10 10:37:49.735958899 +0200
+++ /home/user/Desktop/server.cc	2020-06-10 10:37:24.044099218 +0200
@@ -116,13 +116,6 @@
 int
 WebsocketsServer::start ()
 {
-#if !defined(LWS_WITH_GLIB) && !defined(LWS_WITH_EXTERNAL_POLL)
-	PBD::error << "ArdourWebsockets: check your libwebsockets was compiled"
-	              " with LWS_WITH_GLIB or LWS_WITH_EXTERNAL_POLL enabled"
-	           << endmsg;
-	return -1;
-#endif
-
 #ifndef NDEBUG
 	lws_set_log_level (LLL_ERR | LLL_WARN | LLL_DEBUG, 0);
 #endif
@@ -145,6 +138,17 @@
 		return -1;
 	}
 
+#ifndef LWS_WITH_GLIB
+	/* sometimes LWS_WITH_EXTERNAL_POLL is missing from lws_config.h
+	   but the feature is still available, hence this runtime check */
+	if (_fd_ctx.empty ()) {
+		PBD::error << "ArdourWebsockets: check your libwebsockets was compiled"
+		              " with LWS_WITH_GLIB or LWS_WITH_EXTERNAL_POLL enabled"
+		           << endmsg;
+		return -1;
+	}
+#endif
+
 	return 0;
 }
```

[prev_feat_check.diff.zip](https://github.com/Ardour/ardour/files/4757228/prev_feat_check.diff.zip)

PR #501, opened 2020-05-10
Basque translation for Ardour 6, 100% completed

Porrumentzio:
We think it's all correct, except some messages Terminal answers, such as:
"invalid name: porrumentzio@riseup.net, @ is not a valid character"
or
"adwaita-theme: no such color found"

They are only metada and color issues, so, as we are not developers, it would be great if someone could say whether is there any problem.

If it's okay, we would be so happy to add this translation to the official version :)


pauldavisthefirst:
manually merged&rebased. thanks!

PR #503, opened 2020-05-11
Build gtk2_ardour for Linux users

FrameBassman:


x42:
Not a bad start, there is a very long way to go, still. This does not even create a version that runs or can be installed. But since the script only zips up the GUI source folder, that doesn't seem to have been your intention so far.  Could you elaborate what the your goal is?


In order to provide a version that works for all GNU/Linux users, we cannot rely on `apt`, either. The resulting binary would only run on a specific distribution.  I don't see much of a point, at least binaries compiled against debian libs won't help us with development. They'll have different bugs.

We hand-pick and patch some dependencies: https://nightly.ardour.org/list.php#build_deps to make it possible to provide builds that run on any distro, and we've tried to use travis in the past for proper nightly builds but it regularly timed out.  Perhaps we can revisit this, and upload a pre-built build-stack.  That'd be kinda cool.

All that being said, it could be useful for CI, just to test if Ardour commits builds. 
Don't provide binaries just the compile-log, and perhaps run some unit-tests, too. For this use-case, debian's apt would be fine. -- Have a look at the `.dsc` file from https://packages.debian.org/source/sid/ardour  there are few  optional buid-deps missing, also A6 has some new ones (e.g. readline, websockets).

FrameBassman:
Initial motivation was: As a Ardour Ubuntu-based Developer, I want to know, how I can build a project. I think that this build instructions will be helpful for that goal.

I also agree about this version of implementation useless for a full deployment process - but it can be useful for CI. Probably, I want to try automate tests runs and this is just first step of that long process.

Therefore, I suggest remove `deploy` section and add simple test run for that pipeline. What do you think?

PR #513, opened 2020-05-21
Allow CD marker at session start #8029

mister-benjamin:
This removes the policy that a CD marker cannot be placed at (or before) session start. The current policy is enforced inconsistently, as it is not encapsulated/established as an invariant. Imho the current policy doesn't solve any, but introduces problems.

According to my analysis, the current behavior is this:

* You cannot add a CD marker before or at session start in the markers ruler. That is only the case since the recent commit 82eba76c8f, before you could add them but they were lost when reopening the session.
* CD marker tickbox in the Locations window is only deactivated for markers at session start, but not before session start.
* When you move the session start before any CD marker (or to coincide with session start), the CD markers are kept. But if you then open the Locations window, the CD flag is being removed from them.
* The intended policy to restrict CD markers is only enforced by the GUI, but not by libardour.
* You can put a CD marker one CD frame after session start. I understand the intention of this whole policy as an attempt to enforce the Red Book standard (2 seconds pregap). The policy only prevents to place CD markers at or before session start. Also, in my understanding, the fulfilment of this aspect of the standard doesn't have to be taken into account in CUE sheets at all (see below).

According to [this document](https://github.com/libyal/libodraw/blob/master/documentation/CUE%20sheet%20format.asciidoc#61-msf), the MSF is relative to the last FILE command. MSF is defined without the 2 seconds MFS offset.

Also, when looking at examples of cue sheets [here](http://wiki.hydrogenaud.io/index.php?title=Cue_sheet), all of the **Track 1, Index 0 (or Index 1 without hidden track)** times start at 00:00:00.

The same goes for [the CUE sheet article](https://en.wikipedia.org/wiki/Cue_sheet_(computing)) on Wikipedia. The author(s) also mention, that *"...INDEX 00 is optional and denotes the pregap. The pregap of Track 1 is used for Hidden Track One Audio (HTOA)"*.

I also added some details on that matter in the [bug report](https://tracker.ardour.org/view.php?id=8029).

In my first version to remove this restriction from Ardour, I changed the policy from "not at session start" to "not before session start", thereby also resolving the inconsistencies mentioned above. E.g. removing the CD flag from a CD marker when moving the session start to a position at a later point in time than the marker. But I realized, that it doesn't even matter and all special cases could be entirely removed, because when exporting the session range with CD markers in it as a CUE, CD markers outside of that range are not used anyhow. Also, that doesn't require updating any .po files, as the corresponding messages can just be removed.

I extensively tested this and exported and burned a couple of CDs with it.
* [DDP tools](http://ddp.andreasruge.de) tools handles it correctly and satisfies the Red Book standard in any case
* **cdrdao** generates two seconds pregap before track 1 and adds any audio material (silence or not) between session start and the first CD marker to it
* **wodim**/**cdrecord** generates two seconds pregap before track 1 and fails when *Index 0* for Track 1 is given

Thus:
* **DDP tools** does the right thing with or without Ardour's current policy
* With **cdrdao**, I always end up with more than two seconds of pregap when burning from CUE sheets exported by Ardour with its current policy
* **wodim** doesn't allow me to burn CDs from CUE sheets exported by Ardour with its current policy

Merging this wouldn't affect anybody's established workflow, but imho it allows for more control without removing anything that enforced standard compliance. It also removes inconsistent special casing from Ardour.

If of interest, here's a transcript of my experimentation:

### r1_session.wav.cue (with two seconds silence between session start and first CD marker)

```
# cat r1_session.wav.cue
REM Cue file generated by Ardour
TITLE "session"
FILE "r1_session.wav" WAVE
  TRACK 01 AUDIO
    FLAGS DCP 
    TITLE "unnamed1"
    INDEX 00 00:00:00
    INDEX 01 00:02:00
  TRACK 02 AUDIO
    FLAGS DCP 
    TITLE "unnamed2"
    INDEX 00 00:18:00
    INDEX 01 00:20:00
...
```

All good with two seconds pregap:
```
# cue2ddp r1_session.wav.cue
=== CUE IMAGE SUMMARY ===
Cue file : r1_session.wav.cue
*** Audio data ***
File name: r1_session.wav
File type: Microsoft Wave
Starts at: 00:00:00.000
Ends at  : 00:01:40.845
Length   : 00:01:40.845
Padding : 00:00:00.007 (1344 bytes)
*** PQ Listing ***
UPC/EAN: 
Trk Idx  Start Time         Track Length       ISRC          PRE DCP 4CH SCMS
--- ---  -----------------  -----------------  ------------  --- --- --- ----
 01  00  00:00:00 (     0)
 01  01  00:02:00 (   150)  00:16:00 (  1200)                 N   Y   N   N
 02  00  00:18:00 (  1350)
 02  01  00:20:00 (  1500)  01:20:64 (  6064)                 N   Y   N   N
...
```

This leads to four seconds pregap:
```
cdrdao write --driver generic-mmc-raw r1_session.wav.cue
```

This fails:
```
wodim -audio -useinfo -text -dao --cuefile='r1_session.wav.cue'
```

### session.wav.cue (session start and first CD marker coincide)
```
# cat session.wav.cue
REM Cue file generated by Ardour
TITLE "session"
FILE "session.wav" WAVE
  TRACK 01 AUDIO
    FLAGS DCP 
    TITLE "unnamed1"
    INDEX 01 00:00:00
  TRACK 02 AUDIO
    FLAGS DCP 
    TITLE "unnamed2"
    INDEX 00 00:16:00
    INDEX 01 00:18:00
...
```

All good with two seconds pregap:
```
# cue2ddp session.wav.cue
=== CUE IMAGE SUMMARY ===
Cue file : session.wav.cue
*** Audio data ***
File name: session.wav
File type: Microsoft Wave
Starts at: 00:00:02.000
Ends at  : 00:01:40.845
Length   : 00:01:38.845
Padding : 00:00:00.007 (1344 bytes)
*** PQ Listing ***
UPC/EAN: 
Trk Idx  Start Time         Track Length       ISRC          PRE DCP 4CH SCMS
--- ---  -----------------  -----------------  ------------  --- --- --- ----
 01  00  00:00:00 (     0)
 01  01  00:02:00 (   150)  00:16:00 (  1200)                 N   Y   N   N
 02  00  00:18:00 (  1350)
 02  01  00:20:00 (  1500)  01:20:64 (  6064)                 N   Y   N   N
...
```

All good with two seconds pregap:
```
cdrdao write --driver generic-mmc-raw session.wav.cue
```

All good with two seconds pregap:
```
wodim -audio -useinfo -text -dao --cuefile='session.wav.cue'
```

### DIFF
--- a/gtk2_ardour/location_ui.cc
+++ b/gtk2_ardour/location_ui.cc
@@ -298,12 +298,6 @@ LocationEditRow::set_location (Location *loc)
 		cd_check_button.set_active (location->is_cd_marker());
 		cd_check_button.show();
 
-		if (location->start() == _session->current_start_sample()) {
-			cd_check_button.set_sensitive (false);
-		} else {
-			cd_check_button.set_sensitive (true);
-		}
-
 		hide_check_button.show();
 		lock_check_button.show();
 		glue_check_button.show();
@@ -533,18 +527,6 @@ LocationEditRow::cd_toggled ()
 		return;
 	}
 
-	//if (cd_check_button.get_active() == location->is_cd_marker()) {
-	//	return;
-	//}
-
-	if (cd_check_button.get_active()) {
-		if (location->start() <= _session->current_start_sample()) {
-			error << _("You cannot put a CD marker at the start of the session") << endmsg;
-			cd_check_button.set_active (false);
-			return;
-		}
-	}
-
 	location->set_cd (cd_check_button.get_active(), this);
 
 	if (location->is_cd_marker()) {
@@ -659,12 +641,6 @@ LocationEditRow::start_changed ()
 	i_am_the_modifier++;
 
 	start_clock.set (location->start());
-
-	if (location->start() == _session->current_start_sample()) {
-		cd_check_button.set_sensitive (false);
-	} else {
-		cd_check_button.set_sensitive (true);
-	}
 
 	i_am_the_modifier--;
 }
--- a/gtk2_ardour/po/cs.po
+++ b/gtk2_ardour/po/cs.po
@@ -10,7 +10,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: gtk-ardour 0.347.2\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-05-20 14:36-0600\n"
+"POT-Creation-Date: 2020-05-21 22:10+0200\n"
 "PO-Revision-Date: 2020-05-13 16:55+0200\n"
 "Last-Translator: Pavel Fric <pavelfric@seznam.cz>\n"
 "Language-Team: Czech <translation-team-cs@lists.sourceforge.net>\n"
@@ -971,7 +971,7 @@ msgstr "Přidat stopy/sběrnice"
 msgid "About"
 msgstr "O programu"
 
-#: ardour_ui.cc:331 location_ui.cc:1215
+#: ardour_ui.cc:331 location_ui.cc:1191
 msgid "Ranges|Locations"
 msgstr "Polohy"
 
@@ -2905,7 +2905,7 @@ msgstr ""
 msgid "Prompter"
 msgstr "Otázka"
 
-#: ardour_ui_startup.cc:167
+#: ardour_ui_startup.cc:168
 msgid ""
 "%4This is a session from an older version of %3%5\n"
 "\n"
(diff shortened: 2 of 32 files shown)


bachstudies2020:
Somewhat related to my request here: https://tracker.ardour.org/view.php?id=7883

Being able to have CD marker at session start is a great step forward given DDPtools does add the 2 seconds if missing. I believe HOFA DDP apps also add the silent 2 second pregap if missing otherwise respects any existing 2-second+ pregap of silence or audio.

mister-benjamin:
I rebased the branch on master.

Note: I ran `waf i18n_po` again... If that should interfere with anything, I would also be happy if you only cherry-pick the commit which changes the code...

x42:
Don't bother about translation changes.  They're not supposed to be updated during normal development cycle.  -- I'll grab this without the translation changes and make it two separate commits.

(one for `libs/*` and one for `gtk2_ardour/*` - we try not to modify the GUI in the same git commit as changing a library)

x42:
rebased and merked (w/o the translation changes) as  6.0-12 974ae5100f and 4c8b9daa0f

mister-benjamin:
Thanks.

PR #514, opened 2020-05-23
Update Japanese translation

inatus:
I have updated Japanese translation to latest.


x42:
Rebased and merged as 6.0-83 e848afe2ec  Thanks!

PR #515, opened 2020-05-24
Add ControlList docs that are relevant for Lua

danielappelt:


x42:
I think you're the first to contribute doxyen documentation by itself! Awesome!

rebased and merged as 6.0-18 522d6d769

PR #517, opened 2020-05-28
Fix build for clang - libs/pbd/pbs/compose.h

Chaz6:
Without this change, build failes on clang with the following error:-

error: call to function 'operator<<' that is neither visible in the template definition nor found by argument-dependent lookup


Chaz6:
Er, sorry no, this does not fix it.

PR #522, opened 2020-06-03
Fix compiling cycles.h on ppc64, don't reference unused external `cacheflush_time` variable

nphilipp:
When attempting to build Ardour on ppc64le, I ran into the error linked below because `cycles_t` is undefined. I also noticed that nothing ever uses `cacheflush_time` and removed it.

* https://koji.fedoraproject.org/koji/taskinfo?taskID=45321920 
* https://kojipkgs.fedoraproject.org//work/tasks/1920/45321920/build.log

### DIFF
--- a/libs/ardour/ardour/cycles.h
+++ b/libs/ardour/ardour/cycles.h
@@ -74,6 +74,7 @@ static inline cycles_t get_cycles (void)
 
 #elif defined(__powerpc64__)
 
+typedef uint64_t cycles_t;
 static inline cycles_t get_cycles(void)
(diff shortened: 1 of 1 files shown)


x42:
This may need additional special casing for with `ifdef __APPLE__`. 
So far PPC builds (32 and 64bit) assumed to be OSX.

x42:
I was wrong above, this doesn't affect Apple/PPC builds.
Rebased and merged as 6.0-88 f6e38219bc
Thanks you.

PR #523, opened 2020-06-07
 Updated and reworked Basque translation with a problem fixed

Porrumentzio:
Added some translations that weren't made and corrected some errors, including that on /libs/ardour/po/eu.po which created crashes when importing audio files.

Translation made by Porrumentzio & Enara Larraitz.


x42:
Rebased and merged as 6.0-158 2ca1f55d6b. Thank You!

PR #524, opened 2020-06-08
pthread fixes (ticket #8200)

mvf:
This PR addresses some issues with the recent pthread changes for [ticket 8200](https://tracker.ardour.org/view.php?id=8200).

### DIFF
--- a/libs/ardour/rt_tasklist.cc
+++ b/libs/ardour/rt_tasklist.cc
@@ -16,7 +16,6 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-#include <cerrno>
 #include <cstring>
 
 #include "pbd/pthread_utils.h"
@@ -92,7 +91,7 @@ RTTaskList::reset_thread_list ()
 			pbd_pthread_create (PBD_RT_STACKSIZE_HELP, &thread_id, _thread_run, this);
 
 		if (rv) {
-			PBD::fatal << _("Cannot create thread for TaskList! (") << strerror(errno) << ')' << endmsg;
+			PBD::fatal << _("Cannot create thread for TaskList! (") << strerror(rv) << ')' << endmsg;
 			/* NOT REACHED */
 		}
 		pbd_mach_set_realtime_policy (thread_id, 5. * 1e-5);
--- a/libs/ardour/rt_tasklist.cc
+++ b/libs/ardour/rt_tasklist.cc
@@ -85,11 +85,13 @@ RTTaskList::reset_thread_list ()
 	g_atomic_int_set (&_threads_active, 1);
 	for (uint32_t i = 0; i < num_threads; ++i) {
 		pthread_t thread_id;
-		int rv = AudioEngine::instance()->is_realtime () ?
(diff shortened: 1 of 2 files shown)


x42:
Can you elaborate the motivation of this?

libs/pbd/pthread_utils.cc
```diff
@@ -23,7 +23,8 @@
 #include <stdint.h>
 #include <string>
 
-#ifndef PLATFORM_WINDOWS
+#if !defined PLATFORM_WINDOWS && defined __GLIBC__
+#include <climits>
```
x42, line 27: interesting, why is `climits` needed?

libs/ardour/rt_tasklist.cc
```diff
@@ -86,13 +85,15 @@ RTTaskList::reset_thread_list ()
...
+		int rv = 1;
+		if (AudioEngine::instance()->is_realtime ()) {
+			rv = pbd_realtime_pthread_create (PBD_SCHED_FIFO, AudioEngine::instance()->client_real_time_priority(), PBD_RT_STACKSIZE_HELP, &thread_id, _thread_run, this);
+		}
+		if (rv) {
+			rv = pbd_pthread_create (PBD_RT_STACKSIZE_HELP, &thread_id, _thread_run, this);
+		}
 		if (rv) {
-			PBD::fatal << _("Cannot create thread for TaskList! (") << strerror(errno) << ')' << endmsg;
+			PBD::fatal << _("Cannot create thread for TaskList! (") << strerror(rv) << ')' << endmsg;
```
x42, line 96: What is the issue with using `? :` operators?

libs/pbd/pthread_utils.cc
mvf: Without it there's potential for silent future breakage.

`climits` wraps `limits.h`, which defines `PTHREAD_STACK_MIN` [1] (used by `pbd_stack_size`). Currently `PTHREAD_STACK_MIN` is only defined by accident because one of the headers included via `pdb/thread_utils.h` includes `limits.h` somewhere. If any of these headers change, the `PTHREAD_STACK_MIN` definition could disappear, unless we include `climits`.

To illustrate, this snippet only includes the system headers directly included by `pthread_utils.{cc,h}`. It doesn't compile for me:
```cpp
#define _GNU_SOURCE
#include <pthread.h>
#include <signal.h>
#include <string>
#include <stdint.h>
#include <cstring>
#include <set>
#include <dlfcn.h>

#ifndef PTHREAD_STACK_MIN
#error PTHREAD_STACK_MIN not defined
#endif
```
[1] https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

libs/ardour/rt_tasklist.cc
mvf: I don't know if this was intentional, but the `? :` changed the behavior. 6.0 would fall back to a non-realtime thread when it couldn't create a realtime one. Pseudocode:
```cpp
if (!AudioEngine::instance()->is_realtime () || pbd_realtime_pthread_create())
    pthread_create();
```
Since 09aa0a3d1ad53f53eb4a7da0cde6644205b28f4d, it's immediate `PBD::fatal()` time if realtime thread creation fails (typically due to `EPERM`). This prevents Ardour from starting into a session when the user doesn't have realtime scheduling capability.

x42:
Thanks! I should have read the excellent long commit messages.

Rebased and merged as 6.0-80 e3dcd1f5d7

PR #525, opened 2020-06-14
WebSockets: improve client API

lucianoiam:
This patch mostly does two things:

- Makes the JS client object observable just like the rest of the observable components (track, plugins...) so one can listen to events like `connected`, `message` using `client.on(...)` instead of passing a bunch of callbacks to the constructor.

- Improves some ES6 related syntax


lucianoiam:
Working on the client gave a chance to further test the WS server. I think `LWS_WITH_GLIB` can led to 100% CPU after some traffic. This never happened with the custom Glib integration (`LWS_WITH_EXTERNAL_POLL`). Research pending.

x42:
Rebased and merged (6.0-156-gc4ead80613).
Thanks, as usual.

PR #526, opened 2020-06-17
French translation review/update

trebmuh:
- adds several new translations
- fixes incorrect capital letter in several translated strings
- change the translation of "tempo map" (EN) from _table de tempo_, for _carte de tempo_ (better translation) across the whole file (ie: consistency)


x42:
Rebased and merged as 6.0-120-ge34d940ab3

PR #527, opened 2020-06-17
French translation update

trebmuh:
- fr.po file update
- min/MAJ fixes (main work)
- bargraphes -> indicateurs
- loudness -> sonie (consistency)
- inline -> en-ligne (consistency)
- préset -> pré-réglage (consistency)
- marker -> repère (consistency)
- fixed several "fuzzy"
- several new translations
- a few errors fixed
- remove unneeded translations (when there is no change)

### DIFF
--- a/gtk2_ardour/po/fr.po
+++ b/gtk2_ardour/po/fr.po
@@ -18,7 +18,7 @@ msgstr ""
 "Project-Id-Version: Ardour 6\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-05-20 14:36-0600\n"
-"PO-Revision-Date: 2020-06-13 22:13+0200\n"
+"PO-Revision-Date: 2020-06-18 00:35+0200\n"
 "Last-Translator: Olivier Humbert <trebmuh@tuxfamily.org>\n"
 "Language-Team: French <>\n"
 "Language: fr\n"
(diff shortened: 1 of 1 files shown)


x42:
Merci! Rebased and merged as  6.0-148 b9ae71a35e.

Re min/majuscule.  If a word is not in a complete sentence, but rather a heading or a list or dropdown, we use upper-case for all words except articles. Isn't that the same in French?   e.g.  dropdown: "Audio Bus"   tooltip:  "This creates a new audio bus".

trebmuh:
> Re min/majuscule. If a word is not in a complete sentence, but rather a heading or a list or dropdown, we use upper-case for all words except articles. Isn't that the same in French?

Not that I know of. A majuscule in French is obeying to specific (quite strict rules). It includes the first letter for a title or sub-title (which is very close to a heading/dropdown/list) even if that's not a "complete" (ie: subject/verb/complement) but more likely a nominal group. [This article](https://www.lalanguefrancaise.com/general/guide-complet-usage-majuscules-francais/) is pretty good it seems if you're interested by this topic.

Then, in French, it'll usually be written as "Bus audio".

Please, note that, I've already spotted other uppercase/lowercase errors in the Ardour French translation. I'm waiting a bit to give myself time to spot others, and I will submit a new PR.

PR #528, opened 2020-06-18
Fix an issue (reported by waf i18n_pot) in cs.po

trebmuh:
This one was causing a fatal error in msgfmt.
HTH


x42:
Thanks! rebased as 6.0-150 49352d0732

PR #529, opened 2020-06-26
Adds a midi map for the M-Audio Oxygen 25 3rd Gen controller

caprenter:
The existing midi map for the M-Audio Oxygen controller did not work for me.
I note that my controller says 3rd Gen on the back
I used the documentation: https://manual.ardour.org/using-control-surfaces/generic-midi/midi-binding-maps/ to create a new map.
I think adding it to the rest of the files might be better than overwriting the old one, but happy for others to decide.
Maps the transport control buttons on the device only. They are labelled C10 to C15 on the controller, and map to 113 to 118 in the file.


x42:
Rebased and merged as 6.0-197 b09d17ebc8. Thanks!

PR #530, opened 2020-06-29
update german translation

edogawa23:


jeythekey:
May I make a quick suggestion (instead of creating a new issue for this):
Currently, the drop-down menu of the edit mode button (slide, ripple, lock) has no German translation. The label shown on the button itself (after selection) is translated, though. Would thus probably be an easy fix?
I also found [this](https://tracker.ardour.org/view.php?id=7924) today, maybe that can also be fixed in this PR?

Thanks!

edogawa23:
@jeythekey thanks for your feedback, it's all there now... the strings mentioned in the Mantis issue have changed a bit since 5.12, and for now I preferred to keep a literal translation, maybe Paul or Robin will consider changing the source string to mention edit point...

x42:
Thank you both! -- rebased and merged as 6.0-189 d0681f8013

I'll have a look at https://tracker.ardour.org/view.php?id=7924 later today.

jeythekey:
Thanks to @edogawa23 also from my side, that was probably one of the fastest fixes I encountered so far ;)

PR #534, opened 2020-07-09
Editing about.cc

smart2128:
Adding Italian translator to about.cc


pauldavisthefirst:
manually rebased and merged. sorry for not adding this myself.

PR #535, opened 2020-07-10
show name when audio track is inactive

iurienistor:
For now it fixes for audio tracks only. I hope I am going in the right direction. Ideally would to be able to re-attache the label from control table to inactive table somehow.
Fixing issue : https://tracker.ardour.org/view.php?id=8298

### DIFF
--- a/gtk2_ardour/audio_time_axis.h
+++ b/gtk2_ardour/audio_time_axis.h
@@ -87,6 +87,9 @@ public:
 	void show_existing_automation (bool apply_to_selection = false);
 	void hide_all_automation (bool apply_to_selection = false);
 
+ protected:
+	bool can_edit_name () const override;
+
 private:
 	friend class AudioStreamView;
 	friend class AudioRegionView;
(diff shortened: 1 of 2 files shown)


gtk2_ardour/audio_time_axis.h
```diff
@@ -87,6 +87,9 @@ class AudioTimeAxisView : public RouteTimeAxisView
 	void show_existing_automation (bool apply_to_selection = false);
 	void hide_all_automation (bool apply_to_selection = false);
 
+ protected:
+	bool can_edit_name () const override;
```
x42, line 91: This is not C++98, but since the parent method is declared `virtual` "override" isn't needed.

This should also be AND modified with parent `RouteTimeAxisView::can_edit_name()`

x42:
Thanks for tracking this down!
I've split this into two and moved the `!active()` test into `RouteTimeAxisView` (that's used for all tracks and busses).
0708cdb6b4a3868361 and c398d85f0b2e6d

I think this issue can be closed now.

x42:
I'm just investigating MIDI.

Looks like `inactive_label.show()` is missing in the c'tor of TimeAxisView.

iurienistor:
Yes, I have just tested and there is a need also for MIDI tracks to be fixed.

x42:
Fixed in 6.2-7 68082abc22

PR #536, opened 2020-07-11
double click to expand / contract the track

iurienistor:
Issue: https://tracker.ardour.org/view.php?id=8297

Double click to expand / contract the track. When is minimized I set it to normal height. Ideally to be able to set it to small height. But small height has little area that is not related to name editing.

Ideally I would like double click to do this:

Return the track to smallest height, or make it maximum (the same as F shortcut is doing for expansion).

Proposal: maybe to add more space to the track header or left side, so there will be easy to double click when the track is small


x42:
Nice! Thanks.
Rebased and merged as 6.2-15 4e6a15938c

PR #540, opened 2020-07-17
Follow symlinks for stored paths in generic UIs

paulfd:
When using the `abstract_path` feature in LV2 plugins, there is a symlink chain created from the saved state of the plugin to the original file. This seems like an UX inconvenience. This patch attempts to follow the symlinks to the original file.

It's not entirely perfect though; a user voluntarily setting a symlink as the path parameter might get irritated by this behavior.

Code organization wise I put a free function there but if it's not the proper style please tell me!


x42:
This must go to https://github.com/lv2/lilv/issues/13

PR #541, opened 2020-07-18
fix memory leak when undo/redo view state

iurienistor:


gtk2_ardour/editor.cc
```diff
@@ -4471,7 +4479,8 @@ Editor::VisualState::VisualState (bool with_tracks)
 
 Editor::VisualState::~VisualState ()
 {
-	delete gui_state;
+	if (gui_state)
```
x42, line 4482: See https://ardour.org/styleguide.html item 25

> Don't write: `if (foo) delete foo;` ... C++ handles `delete (0);` just fine, so just `delete foo` will do.

iurienistor: Ok, thanks you, learned something new.

iurienistor: Do I need to make another pull request?

PR #543, opened 2020-07-18
remove parenthesis from the disabled track name

iurienistor:
I propose to remove parenthesis from the disabled track name.

### DIFF
--- a/gtk2_ardour/route_time_axis.cc
+++ b/gtk2_ardour/route_time_axis.cc
@@ -430,7 +430,7 @@ RouteTimeAxisView::label_view ()
 		name_label.set_text (x);
 	}
 
-	inactive_label.set_text (string_compose("(%1)", x));
+	inactive_label.set_text (string_compose("%1", x));
 	inactive_label.show ();
 
 	const int64_t track_number = _route->track_number ();


pauldavisthefirst:
there needs to some visual indication of inactive status

iurienistor:
> there needs to some visual indication of inactive status

Ok. Maybe a grey out text instead of parenthesis. But since I am still familiarizing with the code... I couldn't find how to user the grey from the theme colors.

pauldavisthefirst:
You can't use "grey". You can get a color by name, and it should preferably be functional name, not an actual color name. UIConfig::color ()

Actually using it generally harder.

iurienistor:
@pauldavisthefirst I don't know if it is the exact file but in themes/dark-ardour.colors there are defined aliases <ColorAliases> and <Colors> used by aliases. There is widget:gray color, but I am afraid it is not intended to be used as "grayout color" in <ColorAliases>. Also, I don't see anything related to track label name in the <ColorAliases>. Do I need to add it there and define also a new color for gray out stuff?...

x42:
What's wrong with using braces? That leaves the name clearly visible, while still providing an indication.

iurienistor:
@x42 My intention was to remove parenthesis and grayout. I thought usually in GUI when something is disabled everything is just grayed out but I am not aware of Ardour specifications.

x42:
Usually only actions that are sensitive are made insensitive (aka grayed out) when they re not available.

Easiest way to mark it grayed-out is:
```
inactive_label.set_sensitive (false);
```

iurienistor:
@x42 I've just tested, the contrast unfortunately is bad. I think probably it should be used in the combination with colors defined in the themes that will overwrite the widget styles.

PR #544, opened 2020-07-18
Add MIDI mapping for AKAI_MPK249

Perfexion:


x42:
Thank you!
Rebased and merged as Ardour 6.2-242 917923512e

PR #545, opened 2020-07-19
WebSockets: implement a better mixer demo

lucianoiam:
Sets the foundation for a real mixer control surface. Includes GPLv3 code from https://github.com/DeutscheSoft/toolkit for the JavaScript audio widgets. I thought reinventing the wheel on this topic was not worth the effort as this library nicely implements what is needed for a mixer. Still there is a simple abstraction layer for the widgets in case they are replaced with an in-house implementation in the future (see `widgets.js`).

This should be considered work in progress, strip labels and FX controls are still missing. 

The traffic logger was moved to a separate surface called "Protocol".

The surface protocol was updated to use native Ardour pan range [0;1.0]. This change had been already commented in the surface C++ code by x42.


x42:
Applied as 6.2-73 52e80db8d

PR #546, opened 2020-07-19
French translation update

trebmuh:


x42:
Merci.

Applied as 6.2-75 d2f74456c

PR #547, opened 2020-07-20
Set JACK transport bar_start_tick field

swesterfeld:
I found that there is an XXX comment which says `still need to set bar_start_tick` in Ardour JACK transport code. I found this while testing SpectMorph JACK transport synchronization, the field is simply not set by Ardour right now. So I added code to fill out the field.


x42:
Rebased and merged as 6.2-183 ad8eb4d05b
Thank you!

PR #549, opened 2020-07-20
add maximized region mode

iurienistor:
In this commit is the idea, maybe you'll like it.

When the user makes "Ctrl + double click" on the region
the region view will be maximized and the visual state just before this will be kept.
This will make like entering into a maximized region mode.
The next "Ctrl + double click" will return visual state to the kept one,
i.e. leaving the maximized region mode.
When the visual state is in maximized mode and there vertical scroll is done
than the mode will be just cleared and the visual state will not be changed.

### DIFF
--- a/gtk2_ardour/editor.cc
+++ b/gtk2_ardour/editor.cc
@@ -252,6 +252,7 @@ Editor::Editor ()
 	, _playlist_selector (0)
 	, _time_info_box (0)
 	, no_save_visual (false)
+	, _minimized_visual_state (0)
 	, _leftmost_sample (0)
 	, samples_per_pixel (2048)
 	, zoom_focus (ZoomFocusPlayhead)
@@ -887,6 +888,7 @@ Editor::~Editor()
 	delete selection;
 	delete cut_buffer;
 	delete _cursors;
+	delete _minimized_visual_state;
 
 	LuaInstance::destroy_instance ();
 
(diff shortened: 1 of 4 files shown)


x42:
Why have you closed it?

iurienistor:
There was no feedback or any comment, thus, I thought probably there is no interest in this.

x42:
I think the average time to review/merge github pull-requests is maybe 2 months, usually we investigate and reviews shortly before the bi-monthly releases.

pauldavisthefirst:
Also, I'm in the middle of reimplementing the nutempo "branch" which requires deep immersion and a large amount of line by line code changes. I would normally have reviewed this by now, but I can't spare the mental bandwidth until nutempo v2 gets closer to completion.

iurienistor:
Ah, ok, thank you. :) Sorry for this, I didn't know about how review process takes place in Ardour project. I am afraid I can't reopen the pull request, probably someone that has right access or I need to make another pull request.

x42:
It cannot be re-opened. There's an error:

> The repository that submitted this pull request has been deleted.

x42:
We can probably grab it directly 
```
curl https://patch-diff.githubusercontent.com/raw/Ardour/ardour/pull/549.patch | git am -
```

iurienistor:
Thank you. I'll make a new pull request. Probably even with additional changes.

PR #550, opened 2020-07-25
WebSockets: updates to the mixer demo surface

lucianoiam:
- Add label strips
- Add a fullscreen toggle
- Improve layout

IMG_32DBCB560304-1


x42:
That looks awesome!

rebased and merged as 6.2-180 b298f44468

PR #551, opened 2020-07-29
Next/Previous preset buttons

akimaze:
Hi, 
Many plugins do not have a convenient UI for quickly changing presets (some don't have UI like noize maker). In case of a long list, turning presets in a combo takes a long time. This PR adds the next/previous preset buttons to plugin window. 

Often when I'm looking for inspiration, I set up loop (L) and change presets in plugin window. Thanks to this buttons, this way of working is very convenient and fast. I used `ArdourIcon::NudgeRight` and `ArdourIcon::NudgeLeft` for icons maybe should be changed.

This is my first contribution for Ardour, so maybe should be done in other way.

noizemaker_next_prev_plugin

### DIFF
--- a/gtk2_ardour/lv2_plugin_ui.cc
+++ b/gtk2_ardour/lv2_plugin_ui.cc
@@ -319,6 +319,8 @@ LV2PluginUI::LV2PluginUI(boost::shared_ptr<PluginInsert> pi,
 	_ardour_buttons_box.pack_end (delete_button, false, false);
 	_ardour_buttons_box.pack_end (save_button, false, false);
 	_ardour_buttons_box.pack_end (add_button, false, false);
+	_ardour_buttons_box.pack_end (next_button, false, false);
+	_ardour_buttons_box.pack_end (prev_button, false, false);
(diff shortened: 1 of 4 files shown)


x42:
I'm currently traveling and will review next week when I'm back (unless @pauldavisthefirst  gets there first).

A quick heads up: Using the scroll-wheel on the preset dropdown allows to quickly zap through presets.
For Ardour 6 we worked hard to reduce the number of buttons and decrease the min. width of the plugin-toolbar. Adding new buttons should not be done lightly.

Context: http://libregraphicsworld.org/blog/entry/week-recap-28-august-2019#music

Ardour5: image

vs.

Ardour6: image

akimaze:
> A quick heads up: Using the scroll-wheel on the preset dropdown allows to quickly zap through presets.

@x42  Well, I wish I had known that earlier.  I think change preset with scroll-wheel should be enough so I close that PR.

ngeiswei:
These two small buttons are still worth considering IMO, I know I would like it.

akimaze:
@ngeiswei @x42 I didn't delete the branch, so it can be reopened/merged if the Ardour Team consider it's worth to do it.

PR #553, opened 2020-08-01
Add option to set note range in MIDI editor to one note.

akimaze:
Hi!
That tiny change makes working with MIDI percussion tracks a lot more convenient. It allows you to use the midi editor as a simple pattern editor by simply set note range to one note:

onenoterange

Useful for kick, snare, hihat etc. Allows faster editing and saves screen space.

### DIFF
--- a/gtk2_ardour/piano_roll_header.cc
+++ b/gtk2_ardour/piano_roll_header.cc
@@ -460,13 +460,16 @@ PianoRollHeader::on_expose_event (GdkEventExpose* ev)
 			Cairo::TextExtents te;
 			cr->get_text_extents(note_name, te);
 
-			if (is_white_key)
-				cr->set_source_rgb(0.0f, 0.0f, 0.0f);
-			else
-				cr->set_source_rgb(1.0f, 1.0f, 1.0f);
+			double r, g, b;
 
+			if (is_white_key)
+				note_name_color(white, r, g, b);
+			else
+				note_name_color(black, r, g, b);
+
+			cr->set_source_rgb(r, g, b);
 			cr->move_to(2.0f + font_size , y + note_height - 1.0f - (note_height - te.width) / 2.0f);
-			cr->rotate(-1.5707); // -90 degrees
+			cr->rotate(M_PI / -2.0); // -90 degrees
 			cr->show_text(note_name);
 		}
 		else if (oct_rel == 0) {
@@ -474,7 +477,10 @@ PianoRollHeader::on_expose_event (GdkEventExpose* ev)
 			double y = floor(_view.note_to_y(i)) - 0.5f;
 			double note_height = floor(_view.note_to_y(i - 1)) - y;
 
-			cr->set_source_rgb(0.0f, 0.0f, 0.0f);
+			double r, g, b;
+			note_name_color(white, r, g, b);
+
+			cr->set_source_rgb(r, g, b);
 			cr->move_to(2.0f, y + note_height - 1.0f - (note_height - font_size) / 2.0f);
 			cr->show_text(ParameterDescriptor::midi_note_name(i));
 		}
@@ -483,7 +489,15 @@ PianoRollHeader::on_expose_event (GdkEventExpose* ev)
 	return true;
 }
 
+void
+PianoRollHeader::note_name_color(const PianoRollHeader::Color& key_color, double& r, double& g, double& b)
+{
+	Gtkmm2ext::Color text_color = Gtkmm2ext::rgba_to_color(key_color.r, key_color.g, key_color.b, 1.0f);
+	text_color = Gtkmm2ext::contrasting_text_color(text_color);
 
+	double a;
+	Gtkmm2ext::color_to_rgba(text_color, r, g, b, a);
+}
 
 bool
 PianoRollHeader::on_motion_notify_event (GdkEventMotion* ev)
--- a/gtk2_ardour/piano_roll_header.cc
+++ b/gtk2_ardour/piano_roll_header.cc
@@ -22,6 +22,7 @@
 #include <iostream>
 #include "evoral/midi_events.h"
 #include "ardour/midi_track.h"
+#include "ardour/parameter_descriptor.h"
 
 #include "gtkmm2ext/keyboard.h"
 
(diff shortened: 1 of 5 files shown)


ngeiswei:
@akimaze, how does that differ from `Note Range->Fit Contents` when there is only one fixed note on the track?

I did notice though that `Note Range->Fit Contents` doesn't always work, maybe instead of adding an extra menu item, this should be fixed?

Also, just in case (since your screenshot does not show it), I would like to mention the existence `Note Mode->Percusive` for displaying drum midi data.

P.S: I'm just an Ardour enthusiast without merging rights. Paul and/or Robin feedback is what really matters.

akimaze:
@ngeiswei I think `Note Range->Fit Contents` gives always minimum one octave view. I'm new in Ardour and I thought it was intentional. And IMO for usual MIDI clip editing this is better than shrink to one note.

> Also, just in case (since your screenshot does not show it), I would like to mention the existence Note Mode->Percusive for displaying drum midi data.

I saw the Percusive view but I think the notes are cut off in this mode (they last very short). I don't know maybe this is a bug or I don't know something about Ardour. So I go back to normal mode and made notes longer.

x42:
> I thought it was intentional. 

It is. the minimum range of the scroomer (when manually zooming) is one octave, to provide context. At least one "C" with an octave label is always visible.

https://github.com/Ardour/ardour/blob/40fd2055665350ef5071f21e7e8259fed97da9aa/gtk2_ardour/midi_scroomer.cc#L38-L39

Then, when the piano-header is displayed, there is an additional constraint that can lead to additional notes:

https://github.com/Ardour/ardour/blob/40fd2055665350ef5071f21e7e8259fed97da9aa/gtk2_ardour/midi_streamview.cc#L396-L404

ngeiswei:
If min range is intentional, then I can see the utility of such menu item. It would probably entail to set a max size of the diamond in percussive mode, cause otherwise, when the track is large (and even when it's minimized), diamonds start to overlap too much.

akimaze:
@x42 Oh, I missed that division, I think in that case range should be set to 1. I will fix that. Also I try add note name.

@ngeiswei  My intention was to simulate the LMMS Beat Editor:

beat_editor

I think that with this option (one note range in MIDI editor) Ardor is even more useful because you can easily humanize/swing notes without loosing fast "pattern-like editing" (In LMMS you need to open track in midi editor and when you change note position it disappear from beat editor).

Percussive mode not working for me. For example when I get "Kick" preset from ZynAddSubFX (lv2) it's played too short, is this working for you when you set crystal notes? Should I configure something?

x42:
I like the simplicity of this! As opposed to other in this direction, this is nice and clean.

Then again, I also think this needs a more general solution. e.g. for a hi-hat track you may want to show 3 notes.
Maybe also on @alcomposer can explain his ideas to only show named MIDI notes (using MIDNAM) and dedicated scales. He made some nice videos/gifs to showcase the idea.

I also think it's a very poor replacement for LMMS Beat Editor. Have a look at origin/beatbox3 branch. That's work-in progress to add a MIDI-region editor/creator in the likes of FL and Numerology. Perhaps @pauldavisthefirst can elaborate.

pauldavisthefirst:
I like the general idea quite a lot, but haven't had a chance to test it out yet. 

the newbbgui branch is the one most related to what x42 mentioned above (the beat editor), but it's not currently in a state where trying it out will reveal anything (the engine part is there, but the GUI is not)

akimaze:
> In any case it would be nice to show the note-name.

I added note name since the area is small I had to make it vertically (A#3 is too long):

note_name

> Then again, I also think this needs a more general solution. e.g. for a hi-hat track you may want to show 3 notes.

This solution is really only a stronger zoom and does not undermine the sense of introducing more advanced solutions, but I think that in many cases it will be sufficient/helpful. Not only with drums but also in any other situation where we only use one midi note.

> Maybe also on @alcomposer can explain his ideas to only show named MIDI notes (using MIDNAM) and dedicated scales. He made some nice videos/gifs to showcase the idea.

Do you have any links? Can't find anything.

> Also related to #477 

I compiled that, very helpful I can't wait when will be merged :)

> I like the general idea quite a lot, but haven't had a chance to test it out yet.

I think it fits well with the philosophy of working in MIDI at Ardor :) I'm the one who likes integrated MIDI editor (although I didn't like it at first, I had to work few hours to start liking it). And this option is a good complement to this idea.

> the newbbgui branch is the one most related to what x42 mentioned above (the beat editor), but it's not currently in a state where trying it out will reveal anything (the engine part is there, but the GUI is not)

I was looking for this branch `beatbox3` but found only very old `beatbox`. Thanks for clarification. Are there any docs about how the beat editor will look? 

@x42 @pauldavisthefirst I worked hard on this today, because I am leaving tomorrow and will be available after 11 August 2020. So if any changes will be needed, feel free to make them or wait for me to come back.

Sorry for bad English.

gtk2_ardour/piano_roll_header.cc
```diff
@@ -443,25 +449,42 @@ PianoRollHeader::on_expose_event (GdkEventExpose* ev)
...
-			s << "C" << cn;
+			cr->set_font_size(font_size);
+
+			Cairo::TextExtents te;
+			cr->get_text_extents(note_name, te);
+
+			if (is_white_key)
+				cr->set_source_rgb(0.0f, 0.0f, 0.0f);
+			else
+				cr->set_source_rgb(1.0f, 1.0f, 1.0f);
```
x42, line 466: Have a look at `Gtkmm2ext::contrasting_text_color()` - in general we prefer slightly off-white text (and it may eventually become themeable)

gtk2_ardour/piano_roll_header.cc
```diff
@@ -443,25 +449,42 @@ PianoRollHeader::on_expose_event (GdkEventExpose* ev)
...
+
+			cr->move_to(2.0f + font_size , y + note_height - 1.0f - (note_height - te.width) / 2.0f);
+			cr->rotate(-1.5707); // -90 degrees
```
x42, line 469: Prefer `M_PI / -2.0`

akimaze: Fixed

gtk2_ardour/piano_roll_header.cc
akimaze: Fixed

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #554, opened 2020-08-02
Sort plugins by name (label) in the preset combo selection

enorrmann:


x42:
This used to be the case in the past. In 4e1464e70fc73b this was intentionally reverted, and factory presets are ordered in the way the vendor specifies.  This change was motivated for plugins that have many presets that are nicely grouped.

The general sort order is: factory presets as ordered by the vendor, then user-presets, ordered by name.

PS. Since control-surfaces also allow access to plugin-presets, the change has to be done in the backend (not the GUI).
Although the GUI could offer an option to sort by label.

enorrmann:
> then user-presets, ordered by name

Well it seems that this is not the case (as far as I understand).
I have several custom saved presets and they are displayed (apparently) randomly at the gui, or may be sorted by creation date, I'm not sure... but definitively not by name.

x42:
Which plugin standard? AU? LV2?

enorrmann:
VST... same happens in the "favorite plugins" explorer 
https://i.imgur.com/1oekOMG.png

PR #555, opened 2020-08-02
WebSockets: properly release strips and plugins when removed

lucianoiam:
This should fix some crashes reported by the beta testing team

### DIFF
--- a/libs/surfaces/websockets/component.h
+++ b/libs/surfaces/websockets/component.h
@@ -57,7 +57,7 @@ public:
 	PBD::EventLoop*              event_loop () const;
 	Glib::RefPtr<Glib::MainLoop> main_loop () const;
 	ARDOUR::Session&             session () const;
-	ArdourMixer&                mixer () const;
+	ArdourMixer&                 mixer () const;
 	ArdourTransport&             transport () const;
 	WebsocketsServer&            server () const;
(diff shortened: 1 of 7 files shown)


x42:
Rebased and merge as 6.2-181 f007ba6b46 Thanks!

x42:
Reverted as 6.2-186 dfe2270c9f  due to C++11isms.

`std::unique_ptr`, `std::move` is not available in C++98.

Instead of explicitly managed Connections, it would seem easier to create a Stripable class that _has-a_ ScopedConnectionList. That way when the Stripable is removed (Drop references), the connection-list leaves scope and all signals are unsubscribed (also all plugin signals).

lucianoiam:
Thanks for reviewing, sounds like a smart concept and effectively led to a cleaner implementation. These latest commits drop the hackish approach in favor of that. As a side effect the internal server mixer interface now resembles the client counterpart more, probably making the whole design a bit more self-explainable.

There are new strip and plugin wrapper objects that contain their connections.

Everything should adhere to C++98 this time

x42:
Rebased and merged as 6.2-225 8eb4dcb675

I've also fixed C++11ims (`ArdourMixerNotFoundException`), and added a namespace in `af79240c81ccca`.
 This prevent global symbols such as "Json", those will only cause trouble when combined with libraries.

PR #557, opened 2020-08-03
fixed typo "sontrollables"

enorrmann:


x42:
Thanks for catching that. Rebased and merged as 6.2-182 d21337a3a4

PR #558, opened 2020-08-03
One more typo fix in loudness dialog

bachstudies2020:


x42:
Rebased and merged as 6.2-134 bc84562699

PR #559, opened 2020-08-09
Adding AVX optimized routines for Linux

ashafq:
This commit adds AVX optimized routines for the following
procedures below:

*_compute_peak
*_find_peaks
*_apply_gain_to_buffer
*_mix_buffers_with_gain
*_mix_buffers_no_gain

AVX optimized routine has the prefix of: x86_sse_avx_

Note: mix_buffer_with_gain and mix_buffers_no_gain may prefer
SSE implementaion over AVX if source and destination pointers
are aligned to 16 byte boundaries. Therefore, it will be optimal if
_all_ audio buffers are allocated to 32 byte boundaries to take
full advantage of AVX ISA extension.

### DIFF
--- a/gtk2_ardour/midi_region_view.cc
+++ b/gtk2_ardour/midi_region_view.cc
@@ -39,6 +39,7 @@
 
 #include "pbd/memento_command.h"
 #include "pbd/stateful_diff_command.h"
+#include "pbd/unwind.h"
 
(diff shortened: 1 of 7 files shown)


x42:
Squashed and rebased as 6.2-161 03abc1076e

Thanks you very much!

PR #562, opened 2020-08-23
Enable neon detection

ashafq:


x42:
Rebased and slightly re-worked as 1f878636c85

Cheers!

PR #565, opened 2020-09-01
WS: crash bugfix related to missing strip panner

lucianoiam:
Replaced internal mixer strip method `is_vca()` for `has_pan()`. Checking `has_pan()` before `pan()` when building the initial state for a recently connected client to avoid try/catch.

In the current state of things the GUI does not really need to know a low level detail like a track being a VCA to determine if a pan control and/or plugins should be presented. Revealing track types can be a future enhancement, ie. exposing the `PresentationInfo` flags.

Now the availability of a pan control is explicit through the second boolean arg value in the WS `strip_description` message (previously indicated if the track was VCA). For the plugins check it is enough to check if the plugins array returned by the JS client for a certain strip is empty or not.

Side note, MIDI tracks (unsurprisingly) use a linear scale for the level fader and the demo mixer strips are hardcoded to dB. Since mixing audio and MIDI is common practice exposing the track type could be the next update, replacing the has_pan field and expecting the client to be smarter.

### DIFF
--- a/libs/surfaces/websockets/dispatcher.cc
+++ b/libs/surfaces/websockets/dispatcher.cc
@@ -65,19 +65,16 @@ WebsocketsDispatcher::update_all_nodes (Client client)
 		
 		ValueVector strip_desc = ValueVector ();
 		strip_desc.push_back (strip.name ());
-		strip_desc.push_back (strip.is_vca ());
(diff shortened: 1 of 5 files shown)


x42:
Nice! Rebased and merged as 6.2-240 9c08c058a3

PR #566, opened 2020-09-05
WS: properly support MIDI strips

lucianoiam:
This patch exposes strip flags to the client, allowing it to choose the correct scale for strip faders and meters.


x42:
Sorry it took so long.
Rebased and merged as 6.3-341 9ee828b47b

lucianoiam:
It is understandable given all the latest great progress on Ardour. I yet have to write some documentation for this.

PR #569, opened 2020-09-16
Fix for the French translation

trebmuh:


x42:
Applied as 6.3-193 b527e47165

PR #570, opened 2020-09-19
Audio track: show spectrogram in regions

liuanlin-mx:


x42:
Wow. That's a nice surprise. Well done!

It might benefit from keeping keeping the fftw buffers as class members instead of re-allocating them for every drawing request. -- Compare to the outline for this feature: https://github.com/Ardour/ardour/tree/spectrowave

Have you tried if it scales to a few hundred tracks?

PS. It'd also be great to split this into two (or more) commits. The libwaveform and lib API changes first and then the GUI.
but we could also do that when re-basing.

1123852253:
I only tried a dozen audio tracks
draw_spectrum is not efficient (region->read() takes a long time. Maybe the cache is invalid?)

prokoudine:
I did test this with a hundred tracks where only ten had spectrum mode enabled. On average, 3 times as slow playback, and uneven at that (goes faster, slower, then faster again). Sometimes, the playhead would reverse its direction, then frantically skip forward. Overall, good fun, but not exactly usable in large sessions :)

pauldavisthefirst:
This is nice work, but I strongly believe that FFT data should be cached on disk for *sources* similar to what we do for peakfiles and transient analysis.  I think it is inappropriate to regenerate anything on a per-draw basis.

PR #573, opened 2020-10-17
update german translation

edogawa23:


x42:
Rebased and merged as 6.3-218-g1f7e4ee76a

PR #575, opened 2020-10-18
updated german translation

edogawa23:


x42:
rebased and merged as 6.3-290 3d1fe88506

PR #576, opened 2020-10-21
Update script documentation

Stemby:


x42:
Thanks. I've applied the link-fix 6.3-228 e96c7e9b84.

There are 27 "README" files and all are effectively plain text, so I did not pick up the markdown change.

keryell:
Just looking at the `README` file, what is actually the markup language used there?

x42:
None, just plain text

keryell:
Ah, thanks.
In `share/scripts/README` there are some `^_`, `^__`, `^_-*.lua` so I thought it was a mark-up language... :-)

x42:
Those are file-name pattern.

Stemby:
The `^` character isn't part of a filename pattern.

x42:
it's a [glob](https://en.wikipedia.org/wiki/Glob_(programming)).  here `^` matches the beginning of a file's name (similar to regular expression boundary match)

https://github.com/Ardour/ardour/blob/f3fbb96d6a2b49cd305d7ca3af5601352fa49104/share/scripts/wscript#L12

PR #577, opened 2020-10-22
Delete confusing character

Stemby:
The `^` character isn't part of a filename pattern.

### DIFF
--- a/share/scripts/README
+++ b/share/scripts/README
@@ -8,19 +8,19 @@ https://github.com/Ardour/ardour/tree/master/share/scripts
 
 Script Naming conventions:
 
-^_
+_
  A script filename with a leading underscore indicates an example script.
  These scripts are only available from ardour's git repository and not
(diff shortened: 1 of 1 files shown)


x42:
What's confusing about this? It's a glob matching the one used in various places (unit tests, wscript, gitignore):

https://github.com/Ardour/ardour/blob/f3fbb96d6a2b49cd305d7ca3af5601352fa49104/share/scripts/wscript#L12

keryell:
It is usual syntax in the UNIX world since the 70s for regular expressions and this is pervasive in most regular expressions nowadays.
So it depends on the culture of the reader and if the reader is in this context while reading it.
But the real question I have here is: what is the value to have actually `^__` or `__` in front of each paragraph since it explained in English in each paragraph instead of just using an itemized list of paragraphs?

x42:
Rebased and merged as 6.6-454 67cce59f4e

PR #578, opened 2020-11-07
Updated Basque translation

Porrumentzio:
Here it goes.
I only updated metadata because I don't manage to update new strings, although I can see them in new Ardour versions. Aren't they prepared for translation? Running `./waf i18n` does not add any new string.

Also, please add Basque translators into credits (`~/gtk2mm__ardour/*.po`)


x42:
What is the goal of this PR? 
Since the translation has not been changed, the meta-data is also still correct, no?

or have you forgotten to `git add` some new files?

Porrumentzio:
The metadata has changed because a email address and the group name have changed.
____
Also, as I can not create an account on the bugtracker, I wanted to ask to add Basque translators into credits

x42:
I've squashed the individual commits into a single one: 6.3-339 a8d924d9e1

(mainly because commit messages like "Tiny correction" are not really useful)

PR #579, opened 2020-11-08
Fix Tom's Loop

7890:
-There's no 'printf' (!)
-New: Use LuaDialog to tell error cases instead of printing help to the log window.


x42:
rebased and merged as 6.3-294 275a37ff39. Thanks!

PR #582, opened 2020-11-10
RCU: Fix race condition between writer/reader

marcan:
This uses an atomic counter and spins only on the writer side, which preserves realtime behavior on the reader side. The spinning yields (by using the same Boost function from Boost spinlocks) to be scheduler-friendly.

Fixing this bug also lets us be able to confidently drop garbage early in the writer if appropriate, so do that and avoid keeping dead wood if possible.


x42:
rebased and squashed as 6.3-299 f87de76b9f

PR #584, opened 2020-11-12
Fix normalization in Temporal::Beats for times between -1 and 0 beats

marcan:
Also fix normalization of times with negative beats and positive ticks


pauldavisthefirst:

PR #586, opened 2020-11-20
Make sure VST3 projectTimeMusic is updated

selenologist:
Although Ardour claims `kProjectTimeMusicValid`, it never actually set `context.projectTimeMusic`. This resulted in e.g. this issue in Surge https://github.com/surge-synthesizer/surge/issues/2998 . This patch fixes that.

Something similar will need to be done for LV2, this only fixes VST3 :smiley:


x42:
Thanks rebased and merged as 6.3-348-g8f00826803
I've also added support for  `barPositionMusic` which was missing likewise.

PS. LV2 already does that. Also LV2 plugins are informed about the complete tempo-map for each process cycle.

PR #587, opened 2020-12-02
Disable vertical autoscroll for UI elements where it doesn't make sense

tastank:
Fixes issue 6396 as well as a few other instances where autoscroll could be annoying.


x42:
Nice one!
Rebased and merged as 6.5-25 0cfee47867

PR #588, opened 2020-12-07
Fix a few strings in Czech translation

martinvlk:
I fixed a few incorrect translations, after consulting with the lead Czech translator pavelfric@seznam.cz


pafri:
Could the changes be approved and committed? Please. Thanks.

x42:
> Could the changes be approved and committed?

We usually merge translations just before an upcoming release, but no matter this is a small change.

Done. rebased and merged as  6.5-93 e84cec0842
Thanks you!

PR #589, opened 2020-12-13
Directly use HTTPS

BrunoVernay:
Easy: more performant and more secure.


x42:
Thanks!
Rebased and applied as 6.6-172 3f50b6d639

PR #590, opened 2020-12-28
Add Moog Grandmother midnam

Ratfink:
Based mostly on the midnam for the Moog Slim Phatty.  The Grandmother doesn't have any presets, so all this provides is the CCs from the manual, as well as the new ones listed in the release notes for firmware version 1.1.3.  Relevant documents can be [downloaded here](https://www.moogmusic.com/downloads?type=158).

It may be worth pointing out, in case this is checked over with a fine-toothed comb, that the manual contains a typo for CC 75, Oscillator 2 Octave (they're actually 16' through 2', as printed on the synth itself).


x42:
Rebased and merged as 6.5-94 381ebeab13
Thank you!

PR #591, opened 2021-01-01
Add support for AMD/Intel FMA extension

ashafq:


x42:
Rebased, squashed the last two commits and applied as 6.5-78 c856a862af
Thank you. That's an other awesome addition!

PR #592, opened 2021-01-27
add fade_in_length() and fade_out_length() methods to audioregion

Houston4444:
Hi.
This has been discussed with Robin on IRC, this PR adds fade_in_length() and fade_out_length() methods to audioregion.

The goal is to give access to theses values from a Lua script.

I also give a script that makes a new playlist in selected tracks with only audible regions. That means it cuts or removes regions if they are below other ones (except during their fade of course). The new playlist is named track_name._MAIN_ because I attempt to do another script to copy selected regions to the track_name._MAIN_ playlist.
The idea came making music, with many takes on a guitar track playlist, and always have to move many regions to get a silence at a moment. This script allows to have a very clean playlist easier to edit with exactly the same audible contents.

The script has no undo method because I didn't find any way to remove a playlist.
Cheers.


Houston4444:
mmmh, all things considered I probably made a mistake.
my fade_in_length() and fade_out_length() methods returns a double, then I use math.floor() Lua method in my script to convert it in int.
Length here is a number of samples, so the C++ method should directly return an integer, don't you think ?

pauldavisthefirst:
yes, these methods should return samplecnt_t not double

Houston4444:
Ok that is done.
I don't know your politic about embedded Lua scripts, this script copies current playlist of selected tracks, renames the new playlists to track_name._MAIN_, and clean their contents to only audible regions. This way another script can easily copy selected regions to playlist named track_name._MAIN_. I want to try this workflow.

Maybe you will prefer that the script only clear the playlist of selected regions, and this action could be undo.

pauldavisthefirst:
I was going to rebase + merge this, but there's a merge conflict even after fixing the rebase conflict. If you don't mind fixing that up against master, I'll be happy to merge it.

Houston4444:
conflict fixed !

libs/ardour/audioregion.cc
```diff
@@ -1243,6 +1243,20 @@ AudioRegion::fade_out_is_default () const
 	return _fade_out->size() == 2 && _fade_out->when(true) == 0 && _fade_out->when(false) == 64;
 }
 
+samplecnt_t
+AudioRegion::fade_in_length ()
+{
+	samplecnt_t fade_in_length = (samplecnt_t) _fade_in->when(false);
```
x42, line 1249: Why use a temporary variable instead of ` return _fade_in->when(false)`?

ControlList::when (false) returns the position of the last point, which is only the length if  `ControlList::when(true) == 0`.

Either name the function `fade_in_end_position()`, or subtract   `_fader_in->when(true)`.

libs/ardour/audioregion.cc
```diff
@@ -1243,6 +1243,20 @@ AudioRegion::fade_out_is_default () const
...
+	return fade_in_length;
+}
+
+samplecnt_t
+AudioRegion::fade_out_length ()
+{
+	samplecnt_t fade_out_length = (samplecnt_t) _fade_out->when(false);
+	return fade_out_length;
```
x42, line 1257: see above comment for `AudioRegion::fade_in_length()`

share/scripts/new_clean_playlists.lua
```diff
@@ -0,0 +1,260 @@
...
+						else
+							s[2] = cut_point_left
+						end
+
+					else
+						-- rg:      B___C
+						-- s :  A___________D
+
+						-- worst case, compared region rg is above r,
+                        -- rg starts after and finish before r.
```
x42, line 163: whitespace (should use tabs for indent)

x42 (commented):
NIce very elaborate example script too!

Few comments inline

libs/ardour/audioregion.cc
Houston4444: > Why use a temporary variable instead of return _fade_in->when(false)?

TBH, I can't remember, indeed it's useless.

I didn't know it is possible that fade in could start before or after the region start. Is it really possible ? I don't see how to do it in the GUI. The script doesn't manage this possibility.

share/scripts/new_clean_playlists.lua
```diff
@@ -0,0 +1,260 @@
...
+			if is_audio then
+				local ra = r:to_audioregion()
+
+				if ra:fade_in_active() then
+					r_no_cut_before = r_front + ra:fade_in_length() + 64
+				end
+
+				if ra:fade_out_active() then
+					r_no_cut_after = r_end - ra:fade_out_length() - 64
+				end
```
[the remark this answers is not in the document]
Houston4444: regions shorter than 128 samples are not copied to the new playlist, because an isolated region with a such length could only be noise. But at this point of the script we just check from when we could cut the compared regions.

pauldavisthefirst:
Any plans to polish this up for use in 7.0?

Houston4444:
replaced with #969

PR #593, opened 2021-02-01
Equivalent region selection (new)

Houston4444:
Hi. This is the same PR than #391 , github made something very strange because I updated my fork, this one seems to be ok.
I've done just a little code clean, it still works fine.

--- old text ---

This is a way to resolve the problem i mention here: http://tracker.ardour.org/view.php?id=6826

It concerns selection of equivalents regions with key modifiers (on linux, Ctrl and Shift).
I saw in code that the behaviour is quite weird (maybe because of an old behaviour), if you press Ctrl + click, it select all equivalent regions only if tracks are selected. But there is no interest, because it deselect the tracks !!!

So after a long reflexion, I've done this:
if you click on a region with Primary key, it will add/remove to selection:
* Only the clicked region if there is no selected region or if selection contains isolated equivalent region(s)
* All equivalent regions in the other case

if you click on a region with Tertiary key, it will apply selection to:
* only the clicked region if selection contains isolated equivalent region(s)
* All equivalent regions in the other case
Note that a region with no equivalent is not considered as isolated region.

this way, naturally, if user want to select isolated regions, he starts his selection with Primary Key, else he starts his selection with no modifier key, and it does what user wants (I think).

### DIFF
--- a/gtk2_ardour/editor.h
+++ b/gtk2_ardour/editor.h
@@ -763,6 +763,7 @@ private:
 	bool set_selected_control_point_from_click (bool press, Selection::Operation op = Selection::Set);
 	void set_selected_track_from_click (bool press, Selection::Operation op = Selection::Set, bool no_remove=false);
 	void set_selected_track_as_side_effect (Selection::Operation op);
+	bool selection_contains_isolated_equivalent_region ();
 	bool set_selected_regionview_from_click (bool press, Selection::Operation op = Selection::Set);
 
 	bool set_selected_regionview_from_map_event (GdkEventAny*, StreamView*, boost::weak_ptr<ARDOUR::Region>);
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
I will admit to not really understanding what the goal is here. Could you expand more on what issue/problem this is addressing?

luzpaz:
bump @Houston4444

Houston4444:
Ohh, totally forgotten this, sorry. 
This has been discussed on IRC with both Paul Davis and Robin Gareus 2years ago. I said that I will rewrite it with what we said, and I have never done this, sorry 😐.

The idea was to not allow single selection of a grouped region anymore, this case is rare enough to let user unselect group if (s)he really wants to do that. I'll do this soon if I find time for, feel free to work on of course.

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #595, opened 2021-02-11
Chinese translation for Ardour 6.6 

YQ-YSY:
use New Chinese translation is OK.


pauldavisthefirst:
manually rebased and merged. Thank you very much!

PR #596, opened 2021-02-17
Fix musl build

tibequadorian:
With musl libc __gnu_linux__ is not defined, but __linux__ is.

Patch from https://git.alpinelinux.org/aports/tree/testing/ardour/fix-platform-detection.patch


x42:
This file is originally from https://github.com/steinbergmedia/vst3_pluginterfaces/blob/master/base/fplatform.h#L98
(they use `#elif __gnu_linux__ || __linux__`) now.

We should probably just run `tools/update_vst3.sh` and pull this in from upstream

PR #600, opened 2021-02-28
Fix build with GCC 11

tartina:
Atomic fails to build with GCC 11 if it's volatile,
https://developer.gnome.org/glib/2.67/glib-Atomic-Operations.html#g-atomic-pointer-compare-and-exchange

### DIFF
--- a/libs/pbd/pbd/rcu.h
+++ b/libs/pbd/pbd/rcu.h
@@ -100,7 +100,14 @@ protected:
 	 */
 	union {
 		boost::shared_ptr<T>*     rcu_value;
+	/* Atomic fails to build with GCC 11 if it's volatile,
+	 * see https://developer.gnome.org/glib/2.67/glib-Atomic-Operations.html#g-atomic-pointer-compare-and-exchange
+	 * */
+#if __GNUC__ > 10
+		mutable gpointer gptr;
+#else
 		mutable volatile gpointer gptr;
+#endif
 	} x;
 
 	inline bool active_read () const {
@@ -108,7 +115,11 @@ protected:
 	}
 
 private:
+#if __GNUC__ > 10
+	mutable gint _active_reads;
+#else
 	mutable volatile gint _active_reads;
+#endif
 };
 
 /** Serialized RCUManager implements the RCUManager interface. It is based on the


tartina:
I think it's safe to remove volatile for every compiler, according to glib latest docs

x42:
The problem is that we don't use latest glib.
Ardour binaries use glib 2.42.0 -  https://nightly.ardour.org/list.php#build_deps

tartina:
I don't think glib version matters, it seems something about atomic operations changed in the latest (unstable) gcc. Glib devs just pointed it out in the latest docs. I was compiling Ardour 6.6 for the unstable version of Fedora, which uses gcc 11 and glib 2.67.4.
I also got some warnings about this being null, i.e.:
```
In file included from ../libs/pbd/stateful.cc:37:
../libs/pbd/pbd/properties.h: In instantiation of ‘void PBD::PropertyTemplate<T>::apply_changes(const PBD::PropertyBase*) [with T = std::__cxx11::basic_string<char>]’:
../libs/pbd/pbd/properties.h:158:7:   required from here
../libs/pbd/pbd/properties.h:159:74: warning: ‘this’ pointer is null [-Wnonnull]
  159 |                 T v = dynamic_cast<const PropertyTemplate<T>* > (p)->val ();
      |                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
../libs/pbd/pbd/properties.h:98:18: note: in a call to non-static member function ‘const T& PBD::PropertyTemplate<T>::val() const [with T = std::__cxx11::basic_string<char>]’
   98 |         T const& val () const {
      |                  ^~~
```

x42:
> I don't think glib version matters, it seems something about atomic operations changed in the latest (unstable) gcc. 

How about all other `volatile` declarations that Ardour uses for glib atomics? There are around 100 in Ardour's codebase.
Can GCC11 cope with those, and it's only the `g_atomic_pointer_` ones causing issues?

BTW, clang is fine, doesn't warn about volatile, but prints warning about unused results:
```
warning: expression result unused [-Wunused-value] (__extension__ ({ typedef char _GStaticAssertCompileTimeAssertion_9[(sizeof 
                 *(&_active_reads) == sizeof (gint)) ? 1 : -1] __attribute__((__unused__)); (void) (0 ? *(&_active_reads) ^ (-1) : 0); (gint) __sync_fetch_and_add ((&_active_reads), (-1)); 
                 }));
```

tartina:
FWIU the error is thrown when you use g_atomic_pointer_compare_and_exchange ()
This is the error I get

```
In file included from /usr/include/glib-2.0/glib/gthread.h:32,
                 from /usr/include/glib-2.0/glib/gasyncqueue.h:32,
                 from /usr/include/glib-2.0/glib.h:32,
                 from /usr/include/glibmm-2.4/glibmm/threads.h:30,
                 from ../libs/ardour/ardour/session.h:53,
                 from ../libs/ardour/default_click.cc:20:
../libs/pbd/pbd/rcu.h: In instantiation of ‘boost::shared_ptr<X> RCUManager<T>::reader() const [with T = std::__cxx11::list<boost::shared_ptr<ARDOUR::Route> >]’:
../libs/ardour/ardour/session.h:313:25:   required from here
/usr/include/glib-2.0/glib/gatomic.h:112:19: error: argument 2 of ‘__atomic_load’ must not be a pointer to a ‘volatile’ type
  112 |     __atomic_load (gapg_temp_atomic, &gapg_temp_newval, __ATOMIC_SEQ_CST); \
      |     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/pbd/pbd/rcu.h:78:47: note: in expansion of macro ‘g_atomic_pointer_get’
   78 |                 rv = *((boost::shared_ptr<T>*)g_atomic_pointer_get (&x.gptr));
      |                                               ^~~~~~~~~~~~~~~~~~~~
../libs/pbd/pbd/rcu.h: In instantiation of ‘boost::shared_ptr<X> RCUManager<T>::reader() const [with T = std::vector<boost::shared_ptr<ARDOUR::Bundle> >]’:
../libs/ardour/ardour/session.h:346:27:   required from here
/usr/include/glib-2.0/glib/gatomic.h:112:19: error: argument 2 of ‘__atomic_load’ must not be a pointer to a ‘volatile’ type
  112 |     __atomic_load (gapg_temp_atomic, &gapg_temp_newval, __ATOMIC_SEQ_CST); \
      |     ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../libs/pbd/pbd/rcu.h:78:47: note: in expansion of macro ‘g_atomic_pointer_get’
   78 |                 rv = *((boost::shared_ptr<T>*)g_atomic_pointer_get (&x.gptr));
      |                                               ^~~~~~~~~~~~~~~~~~~~
```

Then compilation aborts.
Patching the union is sufficient to get a successful build.
I also modified _active_reads because I read the new glib advice, but it's not needed to build Ardour

waddlesplash:
Pretty much all atomics should not use volatile qualifiers (and this has been true for some time); on the opposite side of things, that means you must *never* access or manipulate atomics outside of the intrinsic functions. (We have used non-volatile atomics in @haiku, even in kernel mode, for the better part of a decade.)

x42:
@waddlesplash  spot on and guilty as changed.

- Ardour still uses C++98 and supports old systems, incl. gcc-4.2 builds on OSX 10.5/6 and WinXP. This will change in the not too distant future though.
- There are various cases where the gint is modified directly, not using `g_atomic_*` function calls. Usually only during initialization. Still volatile is required until this is cleaned up.

waddlesplash:
> Ardour still uses C++98 and supports old systems, incl. gcc-4.2 builds on OSX 10.5/6 and WinXP. This will change in the not too distant future though.

Haiku still builds the core system on 32-bit x86 using GCC *2* (2.95, specifically) for ABI compatibility with BeOS. (We have inline assembly for atomics there, though, as gcc2 does not have `__atomic_*` intrinsics.)

x42:
The "volatile" git branch should address this. Please test:
https://github.com/Ardour/ardour/tree/volatile

tartina:
Testing volatile branch: playing a medium sized session ok.

x42:
OK merged as 6.6-164 8b4edaa950

PR #602, opened 2021-03-07
VST3 set minimum resize

ericfont:
It was previously not possible to resize a resizeable vst3 plugin to be smaller than the default initial width & height.  This PR first finds what the vst3 plugin's minimum size by calling checkSizeConstraint() with a {0,0,0,0} rectangle, which returns the vst3's minimum permitted size in that rect.  And then calls set_size_request with the width and height of that returned rect.

Here is the previous behavior on windows when trying to resize smaller than the initial size:

sonobus-137-resise-in-ardour-windows

Here is the previous behavior on linux when trying to resize smaller than the initial size:

sonobus-resize-ardour6p6-bug

And after applying this commit, here is the new behavior on linux (for a JUCE plugin which has a minimum size of 400x300 set with setResizeLimits):

sonobus-after-set-size-request

I haven't tested on windows yet (and I probably don't want to set up a dev environment in windows), but I think this commit would work for windows too.

### DIFF
--- a/gtk2_ardour/vst3_plugin_ui.cc
+++ b/gtk2_ardour/vst3_plugin_ui.cc
@@ -52,6 +52,16 @@ VST3PluginUI::VST3PluginUI (boost::shared_ptr<PluginInsert> pi, boost::shared_pt
 
 	pack_start (_ardour_buttons_box, false, false);
 	_ardour_buttons_box.show_all ();
+
+    // determine minimum size
+    IPlugView* view = _vst3->view ();
+    ViewRect rect;
+    if (view->checkSizeConstraint (&rect) != kResultTrue) {
+        view->getSize (&rect);
+    }
+
+    // set minimum size
+    set_size_request(rect.right - rect.left, rect.bottom - rect.top);
 }
 
 VST3PluginUI::~VST3PluginUI ()


x42:
I think there was a reason why we didn't do that. Probably plugins that don't have a resize handle themselves.

Usually one resizes the Ardour window (not the plugin). and `set_size_request()` gets in the way.

x42:
As I suspected, with that patch applied VST3 plugins that do not have their own resize handle (e.g. all Steinverg vst4gui) do no longer correctly resize.

x42:
That patch basically overrides the call to `VST3X11PluginUI::view_size_request()` and `_req_height`, `_req_height` become irrelevant.

ericfont:
oh, ok, sorry.

ericfont:
So maybe can I test first to see if there is a resize handle, and if so use this patch?

ericfont:
Sorry I'm trying with other VST3 plugins that don't resize and see that I broke their sizing.

ericfont:
> Usually one resizes the Ardour window (not the plugin).

Ok, now I've figured out what you mean by this sentence.  And I'm noticing (on ardour 6.6 release) if I resize with the plugin's internal gui's resize ability by dragging from near the corner, then the resize is choppy:

ardour6-resize-with-plugins-gui

While if I resize with the ardour gtk2 window by carefully pressing my mouse down when I at the precise corner pixel of the gtk2 window to drag, then the resizing is more smoother:

Peek 2021-03-07 15-17

And I'm noticing trying to resize using the gtk2 window corner dragging doesn't try to redraw the gui smaller than ardour's fixed minimum.



I'm going to keep looking at how to set the minimum resize correctly.

ericfont:
(I'm also wondering if maybe there is a way to force resizes done by the plugin's resize corner area to somehow instead be handled by the gtk window, cause that seems smoother.  Maybe that is the plugin responsibility...I don't know...I have to familiarize myself a bit more)

ericfont:
(I will close this PR till I find an actual solution)

ericfont:
Seems working now.  Complied on latest git of SonoBus (develop branch commit 5bc2d00) and latest git of Ardour (master e07b3104bb23f4bb0812f680b9920105f9867d12), I see that dragging the little diagonal lines in the corner seems to now obey the minimum size setting and reduce the ardour window size...with the minimum size being set according to the plugin I believe:

image

Though I am noticing the the plugin pixels are scaled larger than the normal standalone.

And also I notice if I drag the gtk2 window resize (by carefully draging the actual edge of the window instead of the plugin), then I can drag it to appear smaller than the minimum:

image

Though in actuality this seems to be the minimum pixel size, but just that the pixels are scaled to be larger than they are.

Doesn't seem like a big deal, though just pointing out.

x42:
WIth the patch all the Steinberg VST3s can be resized below their min size. -- With current git-master they have a correct min. size

ericfont:
ok gosh I totally confused myself...I mistakenly wasn't actually running the ardour git master when I made my last comment.

I am going to try to be more careful before I speak again.

PR #605, opened 2021-03-23
Update cubasish-ardour.colors

cooltehno:
This PR changes grid line major and minor from light to noticeable dark colors:
Cubasish change 230321


x42:
Theme update hat-trick?! :) 
Rebased and applied as 6.6-171  49301785d6f08 Thanks!

PR #607, opened 2021-03-30
Fix invalid conversion from 'volatile void*' to 'void*'

jpcima:
The build otherwise fails on gcc 10.2.0, glib 2.68.0.

The origin of this problem is found in `gatomic.h` macros, more precisely the c++11 specific variant.
The volatile flag on `gapcae_oldval` is kept, whereas the other one would discard it.

https://github.com/GNOME/glib/blob/622b31f69e72fdb05b813820e9cd8087e88a47bf/glib/gatomic.h#L206

### DIFF
--- a/libs/pbd/pbd/rcu.h
+++ b/libs/pbd/pbd/rcu.h
@@ -193,7 +193,7 @@ public:
 		 * XXX but how could it? we hold the freakin' lock!
 		 */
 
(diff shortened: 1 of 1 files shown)


x42:
see #600

x42:
I've update the patch from #600 correctly detect glib and C++11: https://github.com/Ardour/ardour/commit/ea01ec5988185228c41b7799b45c5051f0b80d15

PR #610, opened 2021-04-10
Add Alesis VI25 MIDI mappings

MathiasKandelborg:
This enables the usage of Alesis VI25 buttons:

- Stop
- Play
- Toggle loop (I don't know how/if this works but I added it anyway, I didn't get any response from it)
- Toggle record enable


x42:
Rebased and squased as 6.6-256 21f3c6980f
Thank you!

PR #611, opened 2021-04-21
Update compile.sh

Tiefseetauchner:
cd in line 4 would not work if path has blank as it would have two arguments
Example:
run in /home/LukeChriswalker/projects to build/Ardour/tools/x-win
./compile.sh: line 4: cd: too many arguments
Then reaches line 5, and exits with code 1 (not mine to judge, but maybe change that exit code?) without explanation


lodenrogue (approved):

pauldavisthefirst:
manually rebased and merged. thanks.

PR #613, opened 2021-05-01
Freedesktop install

edogawa23:
I realized that a locally compiled and installed ardour6 does not show up in KDE application menu, even when configured with --freedesktop

This PR adds install instructions for freedesktop files (.desktop, mimeinfo xml, appdata xml and application icons) to gtk2_ardour/wscript.

I don't know python well, nor the waf ecosystem, so I've tried and deduced my changes from other content in wscript. I'm not sure if this is in any way complete, e.g. doing this only for Ardour. Mixbus and Tracks to my knowledge don't get packaged by linux distros anyway, so it might be sufficient.

This change would also make it possible to simplify the package building process (RPM spec and similar build scripts), even if that is of zero interest for Paul, for obvious reasons. But as he told me on IRC, contributions are welcome, so I thought I'd give it a go.

### DIFF
--- a/gtk2_ardour/wscript
+++ b/gtk2_ardour/wscript
@@ -941,6 +941,14 @@ def build(bld):
         set_subst_dict(obj, freedesktop_subst_dict)
         bld.install_files (os.path.join (bld.env['PREFIX'], 'share/appdata'), obj.target)
 
+        # install desktop icon files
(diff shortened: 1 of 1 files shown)


x42:
Danke!
Rebased and merged as 6.6-453 a103a89f4f

PR #614, opened 2021-05-13
German translation

edogawa23:
update german translation


pauldavisthefirst:

PR #615, opened 2021-05-15
Update export_format_dialog.cc

ElizabethHarmon:
Small addition to match "limit to" dropdown option.


x42:
Rebased as 6.6-546 0e0327b1cd Thanks!

PR #616, opened 2021-05-16
fix OSC "/marker" feedback

KannebTo:
This PR solves an issue with the feedback on "/marker" OSC address. Before this, only the value "No Marks" was published. Now it works correctly when the playhead is before the first, between two, after the last or exactly at a location marker.


ovenwerks:
Thank you. This has been merged and can be closed.

PR #617, opened 2021-05-21
minor german translation update

edogawa23:


x42:
Applied as 6.7-146 f81d3ce333

PR #619, opened 2021-05-23
Update cs.po

pafri:
I've completed (100 %) the translation file. Please commit it. Greetings, Pavel Fric


x42:
Rebased and merged as 6.7-45 ac1fd6d448. Thank you very much!

PR #620, opened 2021-05-28
Update Basque translation

Porrumentzio:
There was a tiny error


x42:
9 commits to change a single string? I took the liberty  to git --squash them.

Rebased and merged as 6.7-44 1bce8b57d4. Thanks!

PR #622, opened 2021-06-08
Add M4A support to the audio import dialog using FFMPEG.

mkruisselbrink:
This uses the existing optional runtime ffmpeg dependency to add support
for m4a files (and in theory whatever other file formats ffmpeg
supports) to the import audio dialog.

I didn't realize when writing this that the same functionality is mostly
already available via Session -> Open Video (even though m4a isn't
currently included as one of the video formats and isn't really a video format,
it still works). Having this in the import audio dialog however seems much more
user friendly, so this might still be useful.

### DIFF
--- a/libs/ardour/ardour/ffmpegfileimportable.h
+++ b/libs/ardour/ardour/ffmpegfileimportable.h
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2021 Marijn Kruisselbrink <mek@google.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef _ardour_ffmpegfile_importable_source_h_
+#define _ardour_ffmpegfile_importable_source_h_
+
+#include "pbd/g_atomic_compat.h"
+#include "pbd/ringbuffer.h"
+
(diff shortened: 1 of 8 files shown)


pauldavisthefirst:
Any particular reason why you used a ringbuffer between the ffmpeg process and ardour? Seems unnecessary ...

pauldavisthefirst:
I like the idea of import via ffmpeg. But if I am reading this correctly, the PR also creates a new type of "FileSource", FFMPegFileSource, which reads from a pipe. I don't think that this is appropriate (willing to hear arguments the other way). The import should create a new libsndfile-compatible audio file on disk, and then we read from that. What am I missing?

mkruisselbrink:
Thanks for the very quick review.

> I like the idea of import via ffmpeg. But if I am reading this correctly, the PR also creates a new type of "FileSource", FFMPegFileSource, which reads from a pipe. I don't think that this is appropriate (willing to hear arguments the other way). The import should create a new libsndfile-compatible audio file on disk, and then we read from that. What am I missing?

I took most of the inspiration for this particular approach from how mp3 import works. I think the FileSource is needed to support "audition" support in the import dialog; i.e. the import dialog uses `SourceFactory::createExternal` to create a AudioFileSource, and then plays from there. Like the Mp3FileSource most of the AudioFileSource api isn't supported, only those bits that are needed to be able to play the audio file (it seems like it should be possible to implement "audition" support in terms of `ImportableSource` rather then `AudioFileSource` as well, since that does provide seekable reading of audio files as well, not sure what it would take to refactor the import dialog/audition code to actually make that possible; one source per channel for AudioFileSource vs one interlaced source for ImportableSource would probably make it a little bit more complicated).

Re the actual import code, it does seem conceptually simpler indeed to use ffmpeg to convert to some other file format and then use the existing libsndfile support to do the real import, it does seem like that would require some rearchitecting of how audio import works today though, and as such it might actually end up more complicated than the current version of this PR where I "merely" implement the existing interfaces the current import code expects to fine (i.e. ImportableSource).

> Any particular reason why you used a ringbuffer between the ffmpeg process and ardour? Seems unnecessary ...

That seemed the simplest way to adapt the "push" model of the stdout of a SystemExec object with the "pull" model of the `ImportableSource` interface. This way ffmpeg can push chunks of decoded audio into the ringbuffer, blocking/pausing ffmpeg if it is decoding data quicker than ardour consumes it, while to ardour things appear as the same "pull" interface that `ImportableSource` provides for the other supported file formats as well.

x42:
Rebased as 6.7-204 0a54f96a44, then a few fixes for C++98 compat.

Well done and thank you!

PR #623, opened 2021-06-10
Fix crash if SourceFactory::createExternal fails to create a source.

mkruisselbrink:
SoundFileBox::audition handles the case when
SourceFactory::createExternal throws a failed_constructor exception.
However createExternal actually signals failure in most cases by
returning a nullptr, which wasn't handled by SoundFileBox, and thus
would result in a crash.

This isn't usually a problem, since the audition UI would be disabled
unless Ardour had already successfully parsed metadata for the file
being selected, which usually means that creating an AudioFileSource for
the file will succeed as well.

### DIFF
--- a/gtk2_ardour/sfdb_ui.cc
+++ b/gtk2_ardour/sfdb_ui.cc
@@ -533,6 +533,10 @@ SoundFileBox::audition ()
 					SourceFactory::createExternal (DataType::AUDIO, *_session,
 											 path, n,
 											 Source::Flag (ARDOUR::AudioFileSource::NoPeakFile), false));
+				if (!afs) {
+					throw failed_constructor();
+				}
+
 				if (afs->sample_rate() != _session->nominal_sample_rate()) {
 					boost::shared_ptr<SrcFileSource> sfs (new SrcFileSource(*_session, afs, _src_quality));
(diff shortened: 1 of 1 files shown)


gtk2_ardour/sfdb_ui.cc
```diff
@@ -533,6 +533,10 @@ SoundFileBox::audition ()
 					SourceFactory::createExternal (DataType::AUDIO, *_session,
 											 path, n,
 											 Source::Flag (ARDOUR::AudioFileSource::NoPeakFile), false));
+				if (!afs) {
+					throw failed_constructor();
```
mkruisselbrink, line 537: not sure what you'd prefer, this or alternatively duplicating the code from the catch below?

x42:
After a quick look, it seems wrong that `SourceFactory::createExternal` can return a nullptr.
It should `throw failed_constructor ()` instead.

x42:
How did you produce this issue? Is there a specific file that triggers it? Is it fixed by cbbcdb4fe62e21727b0adc5f5f72aaed364a63b5 ?

pauldavisthefirst:
fixed with separate change. Thanks for noticing this issue.

PR #624, opened 2021-06-12
Fix Upsampler new/delete mismatch

napcode:
This fixes a small new[]/delete[] mismatch that occurs when exporting a project.


napcode:
Thanks Paul!

PR #625, opened 2021-06-13
Allow WebSockets surface to run on more libwebsockets configurations 

lucianoiam:
Namely when the system has a build of libwebsockets installed which was not compiled with LWS_WITH_EXTERNAL_POLL enabled. Relevant link https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=977637

This is an issue for the combination of stock ardour and libwebsockets packages on Ubuntu 20.04, which is a popular distro. This patch should fix the issue for it and potentially for other distributions.


x42:
Keep in mind that debian already fixed the issue and binaries from ardour.org/download are not affected.
Still nice change. will merge

lucianoiam:
That's great news!

x42:
Rebased and merged as 6.7-169 e3569b6469

Thank you!

PR #626, opened 2021-06-14
Updates to #625

lucianoiam:
Most importantly, 5407232 can be done without connecting twice to each Ardour signal. This now means a minor patch to the original (pre #625)`ArdourSurface::ArdourFeedback` class. Just overload `event_loop()` for that surface component and return the "helper UI" loop when the main surface loop is unsuitable due to the server component periodically blocking on it. Code is easier to follow and also marginally more efficient.

Also include some minor forgotten updates like updating the surface author email also for the JS file headers.

### DIFF
--- a/libs/surfaces/websockets/component.h
+++ b/libs/surfaces/websockets/component.h
@@ -53,7 +53,7 @@ public:
 	}
 
 	BasicUI&                     basic_ui () const;
-	PBD::EventLoop*              event_loop () const;
+	virtual PBD::EventLoop*      event_loop () const;
 	Glib::RefPtr<Glib::MainLoop> main_loop () const;
 	ARDOUR::Session&             session () const;
 	ArdourMixer&                 mixer () const;
--- a/libs/surfaces/websockets/feedback.cc
+++ b/libs/surfaces/websockets/feedback.cc
@@ -100,8 +100,11 @@ struct PluginParamValueObserver {
 };
 
 FeedbackHelperUI::FeedbackHelperUI()
-	: AbstractUI<BaseUI::BaseRequestObject> ("feedback_helper")
+	: AbstractUI<BaseUI::BaseRequestObject> ("WS_FeedbackHelperUI")
 {
+	char name[16];
+ 	snprintf (name, sizeof(name), "WS-%p", (void*)pthread_self ());
+ 	pthread_set_name (name);
 	set_event_loop_for_thread (this);
 }
 
--- a/libs/surfaces/websockets/feedback.cc
+++ b/libs/surfaces/websockets/feedback.cc
@@ -102,10 +102,7 @@ struct PluginParamValueObserver {
 FeedbackHelperUI::FeedbackHelperUI()
 	: AbstractUI<BaseUI::BaseRequestObject> ("feedback_helper")
 {
-	pthread_set_name ("test_ui_thread"); // FIXME - needed?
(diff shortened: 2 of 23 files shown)


lucianoiam:
There is a pending update... I do not know if Ardour official builds have LWS_WITH_GLIB enabled for libwebsockets; probably not, based on what I saw when trying ardour+libwebsockets from the stable Ubuntu 20.04 repos.

If that changes someday there is a possibility the surface will stop working for users that do not have LWS_WITH_GLIB in their libwebsockets, because there is no way to check for that during runtime. However there is a solution that I think it was not available by the time I developed the surface https://github.com/warmcat/libwebsockets/issues/2322#issuecomment-860576043 . Will have a look.

lucianoiam:
Just to clarify, the situation was exactly the same before #625 it is not that the PR broke something; the source is now better commented and the surface should work on some more configurations. Just let's keep an eye on LWS_WITH_GLIB when building the official packages.

x42:
>  I do not know if Ardour official builds have LWS_WITH_GLIB enabled for libwebsockets;

see https://github.com/Ardour/ardour/pull/625/#issuecomment-860261985
Binaries from ardour.org come with libwebsockets (currently 4.0.15) that features GLIB event loop integration.
You can check for yourself https://nightly.ardour.org/list.php

Self-builds also worked on debian here, even before PR #625.

x42:
PS. users report build failures on older systems:

```
../libs/surfaces/websockets/server.cc: In member function ‘virtual int ArdourSurface::WebsocketsServer::start()’:
../libs/surfaces/websockets/server.cc:150:13: error: ‘struct lws_context_creation_info’ has no member named ‘foreign_loops’
   _lws_info.foreign_loops = 0;
             ^~~~~~~~~~~~~
```

lucianoiam:
Sorry I didn't realize that from your comments in #625 . I now better understand why distro builds aren't supported...

lucianoiam:
> PS. users report build failures on older systems:
> 
> ```
> ../libs/surfaces/websockets/server.cc: In member function ‘virtual int ArdourSurface::WebsocketsServer::start()’:
> ../libs/surfaces/websockets/server.cc:150:13: error: ‘struct lws_context_creation_info’ has no member named ‘foreign_loops’
>    _lws_info.foreign_loops = 0;
>              ^~~~~~~~~~~~~
> ```

Is that new? that part of the code has been always guarded by #ifdef LWS_WITH_GLIB

lucianoiam:
sorr

> PS. users report build failures on older systems (that is current git with only #625)
> 
> ```
> ../libs/surfaces/websockets/server.cc: In member function ‘virtual int ArdourSurface::WebsocketsServer::start()’:
> ../libs/surfaces/websockets/server.cc:150:13: error: ‘struct lws_context_creation_info’ has no member named ‘foreign_loops’
>    _lws_info.foreign_loops = 0;
>              ^~~~~~~~~~~~~
> ```

Removed that line, it is not necessary.

x42:
> Is that new?

Yes, since #625, users still using libwebsockets 2.x are affected.

x42:
Above you said

>There is a pending update...

So this PR is not ready to be merged, yet?

lucianoiam:
I don't plan to make further updates, it can be merged. For official Ardour LWS_WITH_GLIB is guaranteed to work, no need to workaround anything.

x42:
Done. Thanks.

PR #628, opened 2021-08-01
Remove two obsolete prototypes from LV2Plugin

atsampson:
`save_preset` and `remove_preset` were removed in 40c162d609, but the prototypes were left behind.

(Spotted while debugging another problem that turned out not to be Ardour's fault!)


x42:
Rebased and merged as 6.8-272 d9aaae3614
Thanks!

PR #629, opened 2021-08-03
Create ko.po

MarongHappy:
Korean translate.


pauldavisthefirst:
Thanks for this! I've manually rebased and merged this. Please be aware that a full translation requires more than just gtk2_ardour/po. libs/ardour/po and libs/gtkmm2ext/po also require translation.

PR #630, opened 2021-08-03
Update about.cc

MarongHappy:
Korean translate.


x42:
Rebased and merged as 6.8-274 738d36b5b7 -- Thank You!

PR #634, opened 2021-08-12
German translation

edogawa23:
unfinished, but please include this before the nutempo merge...


x42:
squashed into a single comment and merged as 6.8-291 945c8f2880

PR #636, opened 2021-08-14
Update ko.po

MarongHappy:
Ardour 6.9 Update.


x42:
Sorry for the late merge. Since 6.9 is the end of Ardour 6 there was no rush. There will also be a "call for translation" before 7.0 will be released.

Anyway, rebased and merged as 7.0-pre0-2489 d5df5847e1
Thank you

PR #638, opened 2021-08-19
Fix MIDI send latency compensation

marcan:
Send delays for MIDI were not working, resulting in a track setup like this:

* MIDI send -> non-latent bus
* MIDI send -> latent bus

...ending up with the non-latent bus playing earlier than it should.

This also fixes the send meters for MIDI sends, which suffered from the same problem.


x42:
Rebased and merged as  7.0-pre0-867-g9b99633abe
Thank you!

PR #639, opened 2021-08-28
Use HTTPS instead of HTTP

BrunoVernay:
The tutorial is a redirect to a http link, but still. it does not hurt.
The "pingback"  should not be very risky, but better safe than sorry.
Only the local video-server stays in plain HTTP.

### DIFF
--- a/libs/ardour/ardour/rc_configuration_vars.h
+++ b/libs/ardour/ardour/rc_configuration_vars.h
@@ -260,13 +260,13 @@ CONFIG_VARIABLE (DenormalModel, denormal_model, "denormal-model", DenormalFTZDAZ
 
 /* web addresses used in the program */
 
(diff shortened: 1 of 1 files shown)


abulvenz:
Now that you want to change it:
https://ardour.org/donate redirects permanently to https://community.ardour.org/donate

x42:
At least for URLs that ardour loads itself (pingback) we cannot rely on SSL.
Ardour binries ship libcurl and NSS, and then there is custom code to lookup certs which are installed system-wide (not bundled).

Depending on the distro, this may or may not work.

BrunoVernay:
OK: I keep HTTP for Ardour's internaly handled HTTP requests.
I integrated the permanent redirect.

pauldavisthefirst:
I've manually updated the donate URL.

PR #640, opened 2021-08-29
Make BBT_Time::print_padded const

ngeiswei:
As this method does not modify BBT_Time it can be const.  This allows such
method to be called over a const BBT_Time instance.


pauldavisthefirst:

PR #641, opened 2021-09-02
Create add_cdmarker.lua

ElizabethHarmon:
Allows for creation of CD markers using a keyboard shortcut (or one of the action buttons at the top right). Useful for while transport is rolling (as an alternative to TAB for generic marker) and for efficient and accurate placement at playhead without having to right-click then left-click on the ruler.


x42:
Nice and simple! Thanks!
rebased and merged as 7.0-pre0-903 08e22fdb4b

PR #642, opened 2021-09-04
Cache the absence of ffmpeg_harvid

marcan:
Without this, absence of the transcode tool causes it to be searched for
for every entry in the file picker.


x42:
This is intentional, nag the user and pick up the tools when a user installs them while Ardour is running.

marcan:
If it's intentional it should be checked once per file picker open or something like that. Right now it gets checked for *every single file* in the picker, and that check involves stat()ing every single binary on my system. Even with everything hot in cache and an NVMe SSD, that still takes on the order of 30-60 seconds to browse certain directories with the file picker, every time.

x42:
Rebased and merged as 7.0-pre0-917 d4b0090850

PR #643, opened 2021-09-04
Allow track views to be more than 1000 pixels tall

marcan:
People have 4K displays these days; having a single maximized track is
not unusual.


x42:
rebased and merged as 7.0-pre0-916 a33261df3d

PR #644, opened 2021-09-04
Fix spurious MIDI region multi-edit

marcan:
Editor::set_selected_midi_region_view relies on the MIDI region
selection existing in order to trigger deselection of the notes
contained within, prior to making a new selection.

MidiRegionView::add_to_selection was clearing the region selection
itself prior to invoking set_selected_midi_region_view, which
resulted in "ghost" selected notes being left behind. Upon switching
edit modes, this actually allowed the user to spuriously and
inadvertently edit multiple MIDI regions at once.

Fix this by just not clearing the selection and letting
Editor::set_selected_midi_region_view do it.

Repro steps:

Starting in Grab mode with everything deselected, in a project with two or more
distinct MIDI regions:

- Enter Edit mode
- Drag to select notes in the 1st region
  * Notes in the 1st region are selected
- Drag to select notes in the 2nd region
  * Notes in the 2nd region are selected
  * The 1st region *itself* is deselected, but its notes remain highlighted
  * At this point edits behave as intended, and only affect the 2nd region
- Click the Edit mode button again
  * Now *both* regions are selected
- Make an edit, e.g. hit up to transpose
  * Selected notes in *both* regions are now affected!


pauldavisthefirst:
Fixed via other changes. Thanks, though!

PR #645, opened 2021-10-18
Fix midi maps

ptokarski:
Master bus reference was incorrect (lowercased), so connections between control surface and DAW controls didn't work for master bus. This pull request fixes all the predefined configurations. Tested for nanoKONTROL2 with Master preset.


luzpaz:
Is this PR/Issue still relevant ?

x42:
Since 716501defad both "Master" and "master" are recognized and it should also work if the master-bus name itself is translated.

PR #646, opened 2021-10-20
Fix mime package install location

h3xx:
`<prefix>/share/mime/applications` gets clobbered with data from `<prefix>/share/mime/packages/*.xml` every time "update-mime-database" is run.


pauldavisthefirst:

PR #627, opened 2021-07-10
Fix various typos

luzpaz:
Found via `codespell -i 3 -w -S *.po,./share/patchfiles,./libs  -L ba,buss,busses,doubleclick,hsi,ontop,seh,sur,ue`

PR #648, opened 2021-10-30
Typo in code and comments paramater

BrunoVernay:
paramater / parameter
In C code, Lua code and comments.  Searched through the whole repo, but not tested ...
It is a tiny subset of #627 !  Eventually, I could slowly distill #627 in separate PR as they are merged. If code is impacted I could build the thing but I guess some kind of automation rebuilds Ardour for each commit

### DIFF
--- a/gtk2_ardour/plugin_eq_gui.cc
+++ b/gtk2_ardour/plugin_eq_gui.cc
@@ -407,7 +407,7 @@ PluginEqGui::run_impulse_analysis ()
 	/* run at most at session's block size chunks.
 	 *
 	 * This is important since VSTs may call audioMasterGetBlockSize
-	 * or access various other /real/ session paramaters using the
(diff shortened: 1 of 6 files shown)


x42:
`libs/appleutility` is unmodified upstream code from Apple -- best leave that alone (or file an issue upstream)

Also we generally prefer to split changes to `lib/*` and `gtk2_ardour/*` into separate commits.

BrunoVernay:
3 commits
- gtk2_ardour/ (Comments)
- libs/ (Code!)
- Lua (Comments)

I removed Apple stuff.

x42:
Thanks! rebased and merged as 7.0-pre0-1145-g157af3ac4f

BrunoVernay:
On a Sunday? Works of passion ...

PR #649, opened 2021-12-01
fix the '-c' (or '--name') option that changes the JACK client name

Houston4444:
All is in the title. This patch fixes the ardour7 `c` (or `--name`) argument behavior.
See `ardour7 --help`.


Houston4444:
Already done in #652

PR #652, opened 2021-12-02
Nsm switch new2 (rebased)

Houston4444:
This PR superseeds #650 , it enables NSM :switch: capability which allows to open an other session without closing ardour.

this is the same as #651 , with master updated because of conflicts.

Of course, the JACK client is totally removed and recreated with the new client name.

I added the private bool _nsm_first_session_opened to ARDOUR_UI, to prevent to re-scan plugins if the session to open is not the first one.

I added a bool 'force_unload' optional argument to unload_session(), to prevent warning message in cases of changes in the session while switching from NSM. At NSM switch, the ardour session is saved, and then, at a moment we can't determinate (but very few time after*), the other session is loaded.

It seems to work very well, I have not encountered any problems.

### DIFF
--- a/.directory
+++ b/.directory
@@ -1,4 +1,0 @@
-[Dolphin]
-Timestamp=2021,12,2,14,28,56
-Version=4
-ViewMode=1
--- a/.directory
+++ b/.directory
@@ -1,0 +1,4 @@
+[Dolphin]
+Timestamp=2021,12,2,14,28,56
+Version=4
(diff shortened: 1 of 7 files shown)


x42:
Rebased, squashed and merged as 7.0-pre0-1289  692aa7bbb8004 and 36a2fb1ba8.
Thank you!

Houston4444:
Yes ! thanks !

PR #656, opened 2022-01-01
Allow latency measurements even if inverted wiring is detected

lack:
Inverted wiring is important to call out when detected, but the measured
latency is still valid and should be applicable.

### DIFF
--- a/gtk2_ardour/engine_dialog.cc
+++ b/gtk2_ardour/engine_dialog.cc
@@ -3056,9 +3056,10 @@ EngineControl::check_audio_latency_measurement ()
 	}
 
 	if (mtdm->inv ()) {
+                // Even though the signal is inverted, the latency is measured
+                // So warn, but allow
 		strcat (buf, " ");
 		strcat (buf, _("(inverted - bad wiring)"));
-		solid = false;
 	}
 
 	lm_results.set_markup (string_compose (results_markup, buf));


x42:
> latency is still valid and should be applicable.

Sadly this is not always true, in some cases it can off by a significant amount of time (multiples of the 6 sine periods used for measurement).

Also signals are out of phase, so that's a problem as well when bouncing.  In that case a warning may indeed be fine.
but I'm still torn if this is a good idea.

Have you tested if swapping the polarity in the cable fixes the issue on your end? Or record a signal and check if flipping polarity in ardour passes a NULL test?

lack:
> > latency is still valid and should be applicable.
> 
> Sadly this is not always true, in some cases it can off by a significant amount of time (multiples of the 6 sine periods used for measurement).

I suppose that could be true... How high would the offset have to be?  In my case, the measured latency seems perfect, in that when I record a second track while listening to the first track in my headphones, the second recording ends up perfectly aligned to the first. Or close enough to my untrained ears that I don't notice an issue... And it's definitely more right than using a latency of zero!

> Also signals are out of phase, so that's a problem as well when bouncing.  In that case a warning may indeed be fine.
> but I'm still torn if this is a good idea.

Do you have time to help me understand what the potential issue is exactly with there being an inverted signal? In my use of Ardour (fairly typical multi-track recording where I wear headphones and do a pretty-okay job of isolating any bleed from my headphones to my mic) I haven't noticed any issues, though I'm really not an expert in this space, just a hobbyist.

Would this inversion mean that I may actually get some of the bleed from my headphones automatically/accidentally canceled out in the 2nd track?

> Have you tested if swapping the polarity in the cable fixes the issue on your end?

Yes, I seem to recall that if I rig an inverted cable with 2 cables and some alligator clips, and perform the test again, I get the same latency without the inversion warning.  It's been a while (I've been running happily with this patch and just using my inverted measurement) but I can try this again to be sure I'm remembering correctly.

My setup is pretty straightforward. It's hard for me to envision how there even could be an inversion unless it's something being done intentionally by the hardware for some reason. I have a single USB audio device (a Soundcraft Notepad 12fx), which appears in Linux as a single USB device with 4 inputs and 4 outputs.  I'm using it for input and for output in ardour.  For my latency measurements I've cross-connected the balanced mono aux output (which I understand is a sum of the first 2 of the 4 USB output channels) to the 1st USB input channel using a single balanced cable.

I think I'll also try redoing the measurement tomorrow using the left main output instead of the AUX output to make sure I've got the least complicated signal path possible.

> Or record a signal and check if flipping polarity in ardour passes a NULL test?

Can you tell me more about what you mean by a NULL test? How would I do this?

lack:
> I think I'll also try redoing the measurement tomorrow using the left main output instead of the AUX output to make sure I've got the least complicated signal path possible.

Confirmed.  Balanced cable from L main output to input 1 has about the same latency measurement, but still detected as an inverted signal.

lack:
So I tried a second experiment: I first recorded a nearly-sine-wave pattern using a synth I have hanging around (recorded on USB input channel 3), then I cross-connected the same Main L to Input 1 and re-recorded the exact same signal.  The result doesn't look inverted to me!

image

Given that this is the exact same wiring that leads the latency calculation to say my signal is inverted, maybe something else is going on?

x42:
> then I cross-connected the same Main L to Input 1 and re-recorded the exact same signal.

Thanks. This is precisely what I was interested it when I asked about a NULL test:

If you play back that session and flip the polarity (phase invert -- button at the top on each strip in Ardour's mixer), the signals should cancel out. Not exactly silence due to noise and input gain, but likely close.


So the mystery is why the analyzer detects inverted wiring even if you flip the polarity of a balanced cable...

lack:
> > then I cross-connected the same Main L to Input 1 and re-recorded the exact same signal.
> 
> Thanks. This is precisely what I was interested it when I asked about a NULL test:
> 
> If you play back that session and flip the polarity (phase invert -- button at the top on each strip in Ardour's mixer), the signals should cancel out. Not exactly silence due to noise and input gain, but likely close.

Confirmed.  When I invert the 2nd recording and play back, it's significantly quieter.

> So the mystery is why the analyzer detects inverted wiring even if you flip the polarity of a balanced cable...

Not quite - Maybe I wasn't clear.  When I use a normal balanced cable to do the latency test, I get the inverted wiring warning.  When I invert the polarity of the balanced cable I get no inverted wiring warning.

But doesn't my NULL test show that the wiring is not in fact inverted?

x42:
Sorry, I misunderstood.

Yes, your test show that the wiring is correct, and the latency-detection incorrectly assumes that it is inverted.

x42:
Thanks for doing the extensive testing!

Rebased and merged as 7.0-pre0-1662 79a365edce
(I also took the liberty to change the comment slightly)

PR #657, opened 2022-01-02
AVX/SSE2 optimized routines

ashafq:


x42:
Rebased and merged as 7.0-pre0-1652 b90bfa2597
Thanks a lot!

PR #658, opened 2022-01-09
Upgrade waf to 2.0.23 and enable 'clang_compilation_database'

rtbo:
This updates waf to the latest release version and enables the generation of `build/compile_commands.json` which is a compilation database recognized by some IDE/editors to provide compiler quality linting.
To name a few:
 - Visual Studio Code
 - VIM (with ALE plugin)

Developers using VS Code can then create the file `.vscode/c_cpp_properties.json` to refer to the compilation database.
E.g.:
```json
{
    "configurations": [
        {
            "name": "Linux",
            "compilerPath": "/usr/bin/gcc",
            "compileCommands": "${workspaceFolder}/build/compile_commands.json",
            "intelliSenseMode": "linux-gcc-x64"
        }
    ],
    "version": 4
}
```


x42:
Thanks. I've split this into two commits, separating the actual waf binary update.
Rebased and merged as b8e1cd53ce81 and 7.0-pre0-1704 2139b0391c

I'm thinking of making this optional though `build/compile_commands.json` is 12 MB large.
Then again that's small compared to ~2GB of object files.

rtbo:
@x42 I found out that it can be made optional without effort by loading `clang_configuration_database` in `options` rather than `configure`
The code doc is as follow:
```
Usage:

	Load this tool in `options` to be able to generate database
	by request in command-line and before build:

	$ waf clangdb

	def options(opt):
		opt.load('clang_compilation_database')

	Otherwise, load only in `configure` to generate it always before build.

	def configure(conf):
		conf.load('compiler_cxx')
		...
		conf.load('clang_compilation_database')
```

I've also seen that there is a bug in waf that lead to incorrect args in `compile_commands.json` which lead to these errors (for each `_` gettext macro)

image

Not a big deal, but it should be easy enough to fix in waf. If I can get this done and merged, I'll eventually submit a new PR in ardour.

x42:
I don't see these gettext errors here, but I ran into an issue when cross-compiling (see commit message of af69061644a969012aee35c89bbf3e6f1b0995ce).

rtbo:
@x42 I would really suggest to enable `clang_compilation_database` in `options` rather in `configure`.
(I found out after I posted the PR)
I just created #661

PR #659, opened 2022-01-09
websockets provide BBT

rtbo:
Add `transport_bbt` to the websockets protocol to get the Bar/Beat/Ticks view in a remote UI


x42:
Nice!
Rebased and merged as 7.0-pre0-2488 93987de09f

PR #660, opened 2022-01-09
fix websockets toggle_roll

rtbo:
Also fix `BasicUI::toggle_roll` which is supposed to do the same as spacebar but doesn't.
I simply added the missing parameter as for `ArdourUI::toggle_roll`.
Comments seem to suggest that `BasicUI::toggle_roll` must do the same as `ArdourUI::toggle_roll`.


x42:
Rebased and merged as 7.0-pre0-2490 a77ac7107f
Thank you!

PR #661, opened 2022-01-14
compile_commands.json is generated upon request

rtbo:
`compile_commands.json` not generated by default but upon explicit request.
Enable new waf command: `./waf clangdb`

```
$ ./waf --help
...
Main commands (example: ./waf build -j4)
  build    : executes the build
  clangdb  : generates compile_commands.json by request
  clean    : cleans the project
  configure: configures the project
...
```
```
$ ./waf clangdb
Git version: 7.0-pre0-1751-g25bd59958c

Writing revision info to libs/ardour/revision.cc using 7.0-pre0-1751-g25bd59958c, 2022-01-13
Build commands will be stored in build/compile_commands.json
'clangdb' finished successfully (0.547s)
```

### DIFF
--- a/wscript
+++ b/wscript
@@ -796,6 +796,7 @@ def create_resource_file(icon):
 def options(opt):
     opt.load('compiler_c')
     opt.load('compiler_cxx')
+    opt.load('clang_compilation_database')
     autowaf.set_options(opt, debug_by_default=True)
     opt.add_option('--program-name', type='string', action='store', default='Ardour', dest='program_name',
                     help='The user-visible name of the program being built')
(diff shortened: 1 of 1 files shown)


x42:
Alas, that has the same issue with cross compiling and in `options` it's too early to check for that

x42:
to cross-compile we run
```
./waf configure --dist-target=mingw  --windows-vst --ptformat --with-backends=jack,portaudio,dummy --optimize --cxx11
./waf
```

and during compilation that then produces
```
[...]
[  42/1159] Compiling libs/pbd/cartesian.cc
[  43/1159] Compiling libs/pbd/windows_mmcss.cc
[  44/1159] Compiling libs/pbd/windows_mmcss.cc
[  45/1159] Compiling libs/pbd/windows_timer_utils.cc
[  46/1159] Compiling libs/pbd/windows_special_dirs.cc
[  47/1159] Compiling libs/pbd/windows_special_dirs.cc
[  48/1159] Compiling libs/pbd/windows_timer_utils.cc
[...]
```
So some files are compiled twice, which results in duplicate symbols as mentioned in af69061644a969012aee35c89bbf3e6f1b0995ce.

I don't know if this is a waf bug, or a clang_compilation_database issue. It doesn't happen when compiling natively. 
In those cases `obj.source += ..` is not evaluated twice.

x42:
This is obsolete since 9886d6c19edbf6a and af69061644a96901.

Now explicitly enable it:   `./waf configure -compile-database`

PR #664, opened 2022-01-26
Fix typos in gtk2_ardour/ directory

luzpaz:
This commit only includes typos in the `gtk2_ardour/` directory very easier review. 

Found via `codespell -q 3 -S *.po,./share/patchfiles,./libs -L ba,buss,busses,doubleclick,hsi,ontop,ro,seh,siz,sur,te,trough,ue`


x42:
Thanks for cleaning up after us!

PR #666, opened 2022-01-29
Added MIDI MAP for Alesis Q49 V2 master keyboard

tassyv:


x42:
Thank you!
Rebased and merged as 7.0-pre0-2487 c9a1dfb928

PR #667, opened 2022-01-29
Contributing example LUA scripts

tassyv:


x42:
Nice collection of scripts! Thanks for sharing them.

I'm not sure if we need a collection of "add plugin XYZ" example scripts. In particular if those plugins are not cross-platform nor bundled with Ardour.

I'm tempted to only merge a subset:

* `_connect_MIDI_keyboard_control.lua` is nice. There is no similar example yet
* `_create_from_template.lua` - good example for the `new_route_from_template` API
* `add_audio_track.lua` - this should probably be a built-in feature!
* `add_MIDI_monitor.lua` is potentally useful, and the plugin comes with Ardour.
                                                                               
I'm unsure about `_connect_Hydrogen_MIDI.lua` -- that requires JACK so it won't be useful for the vast majority of Ardour users, but it can come in handy as example script.

what do you think?

tassyv:
Sounds good to me Robin

Note that I tried to name the scripts with the heading _ as suggested for scripts that would be useful only as examples in the git repo (and not shipped)

There is one thing with which I personally struggled when writing my LUA scripts, is to be able to insert a plugin just before the fader. That is visible in the _add_x42_compressor.lua (which also highlights picking mono/stereo version of the plugin).
If you think that aspect would be valuable, I could modify the script to insert one of the standard plugins shipped with Ardour instead ?

x42:
Good point. The examples are indeed useful since they differ in details.
I pulled them all into Ardour 7.0-pre0-1992 dab85fd309  Thanks!

PR #668, opened 2022-01-31
Qmdsp kiss

kiilerix:
Ardour might not love downstream packaging, but simplifying Fedora packaging seems like a lesser evil. I can imagine other distros might have similar challenges.


x42:
Rebased and merged as 7.0-pre0-2485  6f04dfc774

PR #669, opened 2022-02-02
Update dark-ardour.colors - add contrast

cooltehno:
Hello Paul, Robin!
I couldn't express for a long time: what exactly was not satisfied for my test with the default "dark". The most boring thing in the default theme for me  - is the combination of "pale"-green &"pale"-gray & "pale"-red. Of course I understand my subjectivity.))) So, I absolutely don't insist about this PR. I'm going to make some more cascade of tweaks (besides this one) through _separate_ PR(s)  - so you could ponder decisions in a comfortable way.

dark_02_02_2022

### DIFF
--- a/gtk2_ardour/themes/dark-ardour.colors
+++ b/gtk2_ardour/themes/dark-ardour.colors
@@ -1,22 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Ardour theme-name="dark">
   <Colors>
-    <Color name="alert:blue" value="2E41E6ff"/>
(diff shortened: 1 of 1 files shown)


x42:
I'm not a fan of this. I really like the dark color scheme  with the #3d gray. It  is well balanced and should not be darker or have more contrast (which is harder on the eye long term).

cooltehno:
No problem!  :) Just in case, I've upload the gif with gray (them:bg) back to default and lighter (closer back to default)  to widget:green darker (midi   trackheader):
gray_back

PR #670, opened 2022-02-02
Update dark-ardour.colors - outline of selected midi notes

cooltehno:
This PR changes color of the outline of the selected midi notes from the red (which is almost not noticeable especially in big amount of velocity) -> to the brightest white.

midi_note_selected_outline

### DIFF
--- a/gtk2_ardour/themes/dark-ardour.colors
+++ b/gtk2_ardour/themes/dark-ardour.colors
@@ -229,7 +229,7 @@
     <ColorAlias name="midi note mid" alias="alert:yellow"/>
     <ColorAlias name="midi note min" alias="alert:orange"/>
     <ColorAlias name="midi note selected" alias="widget:ruddy"/>
-    <ColorAlias name="midi note selected outline" alias="alert:red"/>
+    <ColorAlias name="midi note selected outline" alias="neutral:foregroundest"/>
     <ColorAlias name="midi note velocity text" alias="theme:contrasting"/>
     <ColorAlias name="midi patch change fill" alias="theme:contrasting selection"/>
     <ColorAlias name="midi patch change outline" alias="neutral:foreground"/>


x42:
So the issue is that with orange notes, the red selection outline is too low contrast?

A red border is used to indicate selection throughout the program. So think it is preferable to change the color schema of MIDI velocity and keep selected outline consistent.

Does that make sense to you?

cooltehno:
Yes, for my test red is too low contrast. But I see another tangible compatibility problem: when I select the notes - the background of region become red (very difficult to recognize selected notes):
red_active_region

pauldavisthefirst:
We could bring the fill feature back. It left in part because we added the velocity display bar, but I agree that the selected-note status is not sufficiently obvious at present.

pauldavisthefirst:
not sure what do about this. the goal is excellent, the result is good, but conflicts with the common selection color everywhere else. The problem is that when we have two levels of things selected (region and notes), there's bound to be some conflict with the colors .... so, I am not sure what to do.

cooltehno:
Hello, Paul! Glad to know you are interested in this PR again! :) I have an idea, but as usually I don't know the volume of affords....:))) If the fill of the selected notes feature could be resurrected, may be the general fill of the note could be dark colored and the interior velocity line - "inverted" light.
selected_notes_fill_190822

Also as additional contrast of red may be increase the editable region transparency.

cooltehno:
As a compensation of the reducing red fill of the editable region - may be to make a thick outline:
selected_notes_fill_190822_2

cooltehno:
or may be the original PR with white borders - as a temporary solution

cooltehno:
@pauldavisthefirst Paul, look!
What if to use a velocity line, as an additional agent for red selection enhancing? Just let the velocity lines in selected notes become red:
red_line_velocity

@x42 Robin, also the "**red selection throughout the program**" is super-satisfied!

pauldavisthefirst:
now implemented in git master branch (outline color used for velocity bar). Looks good!

cooltehno:
Tested! SUPERR!! Thanks Paul!!

As usual (...aaa ...desires have no limits), in the best traditions of a pretentious bastard I've noticed a little problem in high velocity level: red line is very similar to the orange fill of notes (~from 100 to 127). If steel the fill of the selected notes could be dark - this could be perfect contrast to the "general red"(also no need to make the dark fill to be changed by velocity - when we change  velocity - changes only the line length) .
red+orange

But anyway, Big Thanks! :+1:

cooltehno:
If the fill of the selected notes could be constant dark - this looks more intelligible than gradient changing from light to dark (I don't know which variant is easier to implement :))
gradient_vs_constant

luzpaz:
@cooltehno maybe open a 2nd feature request for this ?

luzpaz:
bump

PR #672, opened 2022-02-02
Update dark-ardour.colors - tooltip bg

cooltehno:
This PR changes the background of tool-tip notifications from current **black&white** (Robin's commit) (I think it makes some mess in grid and ruler's areas, which are already black&white) **->** to the theme's general **green**. The old one (pale-blue) also looks too boring/// Again it's all on my subjective taste.))
tooltip_bg


cooltehno:
I've just looked at Mixbus (with its general light-brown yellowish widgets) and tested the 'theme:contrasting alt' yellowish color for tooltip background (also not bad on my taste)
tooltip_contrasting_alt
:

x42:
That's why I've changed it to white on black -- consistent with tooltips on other apps here.

PR #673, opened 2022-02-02
Update dark-ardour.colors - loop rectangle&markers

cooltehno:
This PR changes location loop markers from green (which is the same of "Start&End" markers) -> to alert blue. Also the loop rectangle changes from total ubiquitous green (especially when we have a lot of automation lines opened which are all green) to calm blue.
loop
loop automation


x42:
Nice!

merged as 7.0-pre0-204 bf2ff23e44

PR #674, opened 2022-02-02
Update dark-ardour.colors - secondary clock

cooltehno:
The PR changes the text color of the secondary clock from green -> to white. This unloads the greenish value and adds some difference from the main clock.
secondary_clock


pauldavisthefirst:
I feel that this change makes the secondary clock too prominent/noticeable. I like the idea of it having a different color, if that color was in the same "standout" range as the primary clock.

cooltehno:
may be:
Screenshot from 2022-08-19 12-46-54
I didn't make a new RP and made just a picture (if the idea is right - I'll make a new PR). Here I use the color from the standart Ardour's palette: **"widget:green"**

pauldavisthefirst:
Robin and I both feel that both clocks are of equal importance and should thus have the same color. Displaying (e.g.) seconds and BBT times does not rank the clocks significance. If the user doesn't want the secondary clock, they can hide it.

PR #677, opened 2022-02-09
Reanable solo for all previous soloed tracks

dehnhardt:
I have spent a lot of time mixing in the last few days and kept missing the possibility to reactivate solo mode for the same tracks after deactivating several soloed tracks via the flashing solo button in the transport or monitor area.
This should only affect the solo mode. No other mixer parameters as I want to adjust them, in solo mode, then listen to the full mix, switch to solo again...
With this change, when you have put some channel strips to solo mode and decativate them with the solo button in the transport or monitor area, these channels are stored.
When pressing the solo button in the transport or monitor area again,all channel strips are set to solo mode again.

### DIFF
--- a/libs/ardour/session.cc
+++ b/libs/ardour/session.cc
@@ -6348,6 +6348,24 @@ Session::listen_position_changed ()
 }
 
 void
+Session::add_solo_history (const PBD::ID &routeId)
+{
+	_solo_history.emplace_back (routeId);
+}
+
+void Session::restore_solo_history ()
+{
+	for (PBD::ID routeId : _solo_history) {
+		boost::shared_ptr<Stripable> route = stripable_by_id( routeId );
+		if (route && route->solo_control()){
+			set_control (route->solo_control(), 1.0, Controllable::NoGroup);
+			continue;
+		}
+	}
+	_solo_history.clear ();
+}
+
+void
 Session::solo_control_mode_changed ()
 {
 	if (soloing() || listening()) {
@@ -7448,7 +7466,6 @@ Session::cancel_all_solo ()
 
 	get_stripables (sl);
 
-	set_controls (stripable_list_to_control_list (sl, &Stripable::solo_control), 0.0, Controllable::NoGroup);
 	clear_all_solo_state (routes.reader());
 
 	_engine.monitor_port().clear_ports (false);
--- a/gtk2_ardour/ardour_ui3.cc
+++ b/gtk2_ardour/ardour_ui3.cc
@@ -62,7 +62,11 @@ void
 ARDOUR_UI::cancel_solo ()
 {
 	if (_session) {
-		_session->cancel_all_solo ();
+		if( _session->soloing() )
+			_session->cancel_all_solo ();
+		else{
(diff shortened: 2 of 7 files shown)


libs/ardour/session.cc
```diff
@@ -7347,7 +7364,7 @@ Session::cancel_all_solo ()
 
 	get_stripables (sl);
 
-	set_controls (stripable_list_to_control_list (sl, &Stripable::solo_control), 0.0, Controllable::NoGroup);
+	// set_controls (stripable_list_to_control_list (sl, &Stripable::solo_control), 0.0, Controllable::NoGroup);
```
dehnhardt, line 7367: There is probably still work to be done here. It looked like this line and the next one would do the same - but I'm not sure. When this line is executed, it seems that solo mode is already disabled in all tracks clear_all-solo_state then does nothing - so my mechanism would not work like that.

pauldavisthefirst:
Do you have any plans to finsih this off?

dehnhardt:
> Do you have any plans to finsih this off?

I had a conversation about this with Robin and he said, that this might be a too small solution and that mixer scenes or a general undo would be better. Now that mixer scenes arrived in Mixbus I find,  that this does not solve the problem I tried to solve. (AFAIK as I'm on holiday and had only a small look into the mixer scenes) 
So I'm still willing to continue this if there is a chance that it's accepted and merged. Did you had a look into my explanation an questions?

dehnhardt:
I added handling of VCAs. So now everything seems to work as expected.

x42:
This approach has several issues.

First of all the user-interface: The rude solo button should never restore a solo state.

Next, any changes to connections, VCA assignments, adding/removing sends or any other operation which affects implicit up/downstream solo must invalidate the remembered solo state. -- Also change preferences regarding exclusive solo should do so

Also keep in mind that you can A/PFL not only routes, but also Input ports.

As opposed to Mute, Solo is not an automatable control. My gut feeling is that we should also not allow to fake automation using some volatile state.

dehnhardt:
Thank you for your comments, Robin. 
I've never seen the solo button as you described it. For me it was allways a simple way to disable my - sometimes complex - solo listening settings. And I  often unconsciously pressed this button again to switch the solo back on - which of course never worked. That's how I got the idea in the first place. Of course I see your other points as wall, so: Since this idea is not feasible in a meaningful way now, but I still think that it makes sense to have this functionality, here are a few questions about your suggestion:
The basic idea is to temporarily turn off all solo states while a button is pressed - right?
Would you consider it useful to make the momentary all mute off switch via any enabled the solo button of a channel - e.g. by pressing it with [control] [middle button] of the mouse, or would this also be OK via the rude solo button in your opinion?
The implementation would then iterate over all routes, switch off the solo-state if necessary and pass the temporarily changed route objects in a list to the SoloMuteRelease class and restore the original state after releasing the corresponding button via the release method of the SoloMuteRelease class.
If so, I should be able to doe that.

dehnhardt:
Mixer scenes makes this more or less superflous. I will close it..

PR #678, opened 2022-02-13
Address some gtkmm deprecation warnings

kiilerix:
Some small steps to help Ardour stay current.

The individual changesets are quite trivial and quite reviewable.


kiilerix:
I rebased and dropped a changeset (`hide_all()` cleanup) that las mentioned on IRC seemed problematic.

x42:
Wow. Thanks a lot!
I've resolved conflicts, rebased and merged as  7.0-pre0-2516 5bf088e8d1

For future reference, we try to avoid changing files in `lib/*` and `gtk2_ardour/*` in the same commit (first update the library,  then the API user). But for issue at hand that's not a big deal.

PR #680, opened 2022-03-04
iCon controllers

CalebJPotter:
Implements support for iCon Platform M+ controller and iCon QCon ProG2 controller.

Many of the buttons do not map to these controllers well by default using the Mackie Control device profile. These changes remap the buttons using some existing button handlers, and several new ones.
Some functions added are specific to the QCon ProG2 button labeling, and some are more general.
There is also a fix for a bug where controller strips could get stuck if tracks were deleted.


pauldavisthefirst:
Great PR! Thanks very much for working on this. Not 100% sure when I will merge it right now, but looks very well thought out.

pauldavisthefirst:
manually rebased and merged. Thanks!

PR #682, opened 2022-03-24
Fix memory leak in ctorNilPtrPlacementProxy

Teemperor:
The missing `delete` here is I think clearly visible. The other functions above also allocate on the stack, so I just did the same in this function.

### DIFF
--- a/libs/lua/LuaBridge/detail/Namespace.h
+++ b/libs/lua/LuaBridge/detail/Namespace.h
@@ -485,8 +485,8 @@ private:
     template <class T>
     static int ctorNilPtrPlacementProxy (lua_State* L)
     {
-      const T* newobject = new T ();
-      Stack<T>::push (L, *newobject);
+      const T newobject;
+      Stack<T>::push (L, newobject);
       return 1;
     }
 


x42:
This c'tor is used for weak and shared-pointers.

The memory is free()ed when the shared pointer reference-count reaches zero.

Teemperor:
That's true for the memory managed by those pointers, but IIUC the leak here is the weak_ptr/shared_ptr itself as `T` is ``boost::shared_ptr<ARDOUR::PluginInfo>`.

So the code is:
```
template <class T>
    static int ctorNilPtrPlacementProxy (lua_State* L)
    {
      const T* newobject = new T ();
      Stack<T>::push (L, *newobject);
      return 1;
    }
```

T is `boost::shared_ptr<ARDOUR::PluginInfo>`
```
$ c++filt _ZN9luabridge9Namespace9ClassBase24ctorNilPtrPlacementProxyIN5boost10shared_ptrIN6ARDOUR10PluginInfoEEEEEiP9lua_State
int luabridge::Namespace::ClassBase::ctorNilPtrPlacementProxy<boost::shared_ptr<ARDOUR::PluginInfo> >(lua_State*)
```

If we throw that as dummy code into godbolt with ASAN it confirms the leak:
https://godbolt.org/z/31571c4Eo

x42:
The example code misses a whole lot of what happens You have to implement `do_something` and also run the c'tor from a function call.

The c'tor effectively ends up in  `libs/lua/LuaBridge/detail/Userdata.h:668
With the patch applied this then calls 
```
luabridge::UserdataValue<boost::shared_ptr<ARDOUR::PluginInfo> >::push<boost::shared_ptr<ARDOUR::PluginInfo> > (L=0x555555c17718, u=...) at ../libs/lua/LuaBridge/detail/Userdata.h:402
```

```
    /**
      Push T via copy construction from U.
    */
    template <class U>
    static inline void push (lua_State* const L, U const& u)
    {
      new (place (L)) U (u);
    }
```

The lifetime of the object is now managed by the Lua interpreter and no longer by C++, this will not work for shared or weak pointers. Lua will garbage collect the shared_ptr itself, regardless if some C++ code holds a reference to it.

Teemperor:
> The lifetime of the object is now managed by the Lua interpreter and no longer by C++, this will not work for shared or weak pointers. Lua will garbage collect the shared_ptr itself, regardless if some C++ code holds a reference to it.

The leak is not with the shared_ptr object in lua (which is indeed deallocated by lua and I guess you call the dtor manually somewhere). The code is however creating that shared_ptr by invoking the copy constructor on second shared_ptr (this is the properly managed copy in lua), but the originally allocated shared_ptr is never deleted.

This code effectively does:
```cpp
      // in ctorNilPtrPlacementProxy<T> with T = boost::shared_ptr
      // 1: allocates sizeof(shared_ptr) bytes
      const shared_ptr* newobject = new shared_ptr ();
      // now in: Stack< boost::shared_ptr>::push (L, *newobject);
      // 2: shared_ptr stored in lua's managed memory. 1's contents are copied.
      new (place (L)) shared_ptr (*newobject); 
```

Allocation 2 is fine, but no one is deleting allocation 1. It's only created and then passed to the copy constructor of `boost::shared_ptr`

Maybe it's better to see without the C++ metaprogramming going on. I ran the (unmodified) `luabindings.cc` through Clang. After optimizations you get the LLVM intermediate representation from the snippet below. You can see the call to `new` in the line starting with `%2` (which is the virtual register where the pointer is stored). You can find all uses by just Ctrl+F `%2` (hint: there are no uses of the pointer). From the compiler's perspective the result of `new` is completely unused (the values stored by the default constructor are already optimized away as they just modify leaking memory).

```llvm
; Function Attrs: sspstrong uwtable
define linkonce_odr noundef i32 @_ZN9luabridge9Namespace9ClassBase24ctorNilPtrPlacementProxyIN5boost10shared_ptrIN6ARDOUR10PluginInfoEEEEEiP9lua_State(%struct.lua_State* noundef %0) #4 comdat align 2 personality i32 (...)* @__gxx_personality_v0 !dbg !276317 {
  call void @llvm.dbg.value(metadata %struct.lua_State* %0, metadata !276320, metadata !DIExpression()), !dbg !276322
  ; NOTE: This is the result to the call of `new` vvvvvvvvvvvvvvv
  %2 = tail call noalias noundef nonnull dereferenceable(16) i8* @_Znwm(i64 noundef 16) #42, !dbg !276323, !heapallocsite !21103
  %3 = tail call i8* @lua_newuserdata(%struct.lua_State* noundef %0, i64 noundef 32), !dbg !276362
  %4 = bitcast i8* %3 to i32 (...)***, !dbg !276368
  store i32 (...)** bitcast (i8** getelementptr inbounds ({ [4 x i8*] }, { [4 x i8*] }* @_ZTVN9luabridge13UserdataValueIN5boost10shared_ptrIN6ARDOUR10PluginInfoEEEEE, i64 0, inrange i32 0, i64 2) to i32 (...)**), i32 (...)*** %4, align 8, !dbg !276368, !tbaa !82665
  %5 = getelementptr inbounds i8, i8* %3, i64 16, !dbg !276375
  %6 = getelementptr inbounds i8, i8* %3, i64 8, !dbg !276376
  %7 = bitcast i8* %6 to i8**, !dbg !276376
  store i8* %5, i8** %7, align 8, !dbg !276377, !tbaa !210183
  %8 = tail call i32 @lua_rawgetp(%struct.lua_State* noundef %0, i32 noundef -1001000, i8* noundef nonnull @_ZZN9luabridge9ClassInfoIN5boost10shared_ptrIN6ARDOUR10PluginInfoEEEE11getClassKeyEvE5value), !dbg !276378
  %9 = tail call i32 @lua_setmetatable(%struct.lua_State* noundef %0, i32 noundef -2), !dbg !276379
  %10 = bitcast i8* %6 to i8**, !dbg !276382
  %11 = load i8*, i8** %10, align 8, !dbg !276382, !tbaa !210183
  call void @llvm.memset.p0i8.i64(i8* noundef nonnull align 8 dereferenceable(16) %11, i8 0, i64 16, i1 false), !dbg !276387
  ret i32 1, !dbg !276388
}
```

x42:
Nice bit of detective work!  I see now what you meant. Sorry for being thick.

rebased and merged as 7.0-pre0-2482 3dfcbf17b2e6db1b21011f5ea2e74e5b480dc51c

PR #683, opened 2022-03-31
wscript: fix Apple-specific flag being added on non-Apple platforms

a1batross:
Although GCC and Clang completely ignore this flag on non-Apple platforms, some "GCC-compatible" compilers may not, like EDG frontend based compilers.


x42:
Nice find.
Manually rebased and merged as 7.0-pre0-2481 f26257d6e4

PR #684, opened 2022-04-05
update german translation

edogawa23:


x42:
Great! Thank you for your efforts!
Rebased and merged as 7.0-pre0-2480 77fe1c6d32

PR #686, opened 2022-04-08
adds ability to bank over in mackie sends subview

CalebJPotter:
The cursor_left and cursor_right mackie control buttons will now move the
sends subview left and right like it does in the plugin subview. Previously,
if you had more than 8 sends (which is almost always the case for Mixbus),
then sends 9+ were unreachable on an 8 channel controller.

### DIFF
--- a/libs/surfaces/mackie/subview.cc
+++ b/libs/surfaces/mackie/subview.cc
@@ -732,17 +732,17 @@ bool SendsSubview::handle_cursor_left_press()
 bool SendsSubview::handle_cursor_right_press()
 {
 	uint32_t num_sends = 0;
-	if(_subview_stripable->send_name(0).size() > 0) {
+	if (_subview_stripable->send_name(0).size() > 0) {
 		bool more_sends = true;
-		while(more_sends) {
-			if(_subview_stripable->send_name(num_sends).size() > 0) {
+		while (more_sends) {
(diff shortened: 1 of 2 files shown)


libs/surfaces/mackie/subview.cc
```diff
@@ -717,6 +718,23 @@ void SendsSubview::handle_vselect_event(uint32_t global_strip_position)
...
+	if (_current_bank >= 1)
+	{
+		_current_bank -= 1;
+	}
+	mcp().redisplay_subview_mode();
+
+	return true;
+}
+
+bool SendsSubview::handle_cursor_right_press()
```
x42, line 732: Does this allow to bank beyond the end of available tracks?

CalebJPotter: Yes, you can keep banking right. That's how it is for the plugins subview, so I made it that way here too. I can put in a check to prevent that if you think it would be better.

pauldavisthefirst: I think the plugin subview is wrong in this respect, and so ideally both would stop when they would otherwise bank beyond "the end".

CalebJPotter:
Is this how you want it, with the brackets on the same lame as the conditional preceding it?

pauldavisthefirst:
manually rebased and closed. thanks!

PR #687, opened 2022-04-08
Fix source typos (heirarchy->hierarchy)

luzpaz:
May introduce breaking API changes. Please review carefully.


x42:
Thanks, impressive how many typos we had :)

Since it's a static function, the API change is of no concern.
Rebased and merged as 7.0-pre0-2516 5bf088e8d1

PR #688, opened 2022-04-08
Fix pointer typo (callbck->callback)

luzpaz:


x42:
Rebased and merged as 7.0-pre0-2519 58e5ee6051

PR #689, opened 2022-04-08
Fix pointer typo (scrollin->scrolling)

luzpaz:


x42:
This should perhaps be `scrollwin` (scroll window).
We also also often use  in/out suffixes to indicate nesting of widgets.

In either case "scrolling" as verb makes no sense for an item.

gtk2_ardour/luawindow.cc
```diff
@@ -133,9 +133,9 @@ LuaWindow::LuaWindow ()
...
 	// layout
 
-	Gtk::ScrolledWindow *scrollin = manage (new Gtk::ScrolledWindow);
-	scrollin->set_policy (Gtk::POLICY_AUTOMATIC, Gtk::POLICY_AUTOMATIC);
-	scrollin->add (entry);
+	Gtk::ScrolledWindow *scrolling = manage (new Gtk::ScrolledWindow);
+	scrolling->set_policy (Gtk::POLICY_AUTOMATIC, Gtk::POLICY_AUTOMATIC);
+	scrolling->add (entry);
 	scrollout.set_policy (Gtk::POLICY_NEVER, Gtk::POLICY_ALWAYS);
 	scrollout.add (outtext);
```
luzpaz, line 140: Duh, just realized these lines give it better context scroll**in** and scroll**out**  
I guess I'll just close this ticket then?

x42:
Ideally we'd use `scroll_window_inner` `scroll_window_outer` -- but that's more to type and also needs more space.

Variable names are a bit of a compromise. Some functions take many arguments and if each variable name is long, it's hard to read. Many variables are also just acronyms, single letters or sometimes even not English.

Function/method names are in general supposed to English, although some of the use British English (e.g. analy**s**er vs. analy**z**er,  col**o**r vs. col**ou**r, etc) and we don't mind.

Anyway I'll close this now. Thanks for you effort though, other changes look good!

PR #663, opened 2022-01-25
waf / wscript cleanup

kiilerix:
This adds consistency but shouldn't cause any functional change.

See the individual changesets and commit messages for clearest review, and cherry-pick any good parts.

PR #691, opened 2022-04-09
Trivial waf changes

kiilerix:
To help land #663, let's start with the most trivial parts.

See and review (and cherry-pick) the individual changes.

Quite clearly no functional changes. 

The main value from these changes is consistency to make it easier to make further changes without being bothered with trivial diffs.


x42:
Thanks a lot for doing that! 
This is indeed a lot more readable, and easy to discern as NO-OP changes.

Merged as 7.0-pre0-2528 705de9389b

PR #692, opened 2022-04-09
Static waf changes

kiilerix:
The goal of this PR is to start cleaning up the build files, making them more maintainable.

This is mainly removing unused stuff - very few + lines.

The history is clean - see and review (and cherry-pick) the individual changes.

pyflakes will quite reliably catch many potential problems in such cleanup. Just running any build command will also cover most of these changes. As additional testing, I suggest running a build and verifying that rare build targets such as i18n updates and release scripts work as usual. I'm less familiar with how this usually works and there could be changes or use cases I failed to notice.


pauldavisthefirst:
manually rebased & merged

PR #694, opened 2022-04-12
Mackie plugin subview - prevent infinite banking right

CalebJPotter:
Uses separate functions for PluginSelect subview and PluginEdit subview to prevent infinite banking right.


pauldavisthefirst:
manually rebased and committed. thanks!

PR #695, opened 2022-04-16
Update dark-ardour.colors - patch selector knobs unnamed >> lighter

cooltehno:
This PR changes unnamed **knobs** fill in the midi **patch selector** from almost indistinguishable dark to the intelligible light.
unnamed_patch_fill


x42:
For the main theme, we should only use color to indicate a **significant** difference that is not otherwise obvious without color.

cooltehno:
I suppose I had overloaded the illustration with my GIF enthusiasm// So here's a simple comparative picture (left side - current, right-PR). I mean in current stage unused knobs are the same as used:
patch_selector_knobe_unused

cooltehno:
@x42  "some that do have MIDNAM files have valid responses to unnamed ones" A... didn't know. Now problem)) You're master! ))

pauldavisthefirst:
For the reasons outlined by x42, I think we don't want this change.

PR #697, opened 2022-04-20
US2400: maps track parameters and sends to vpots

CalebJPotter:
Pots 0-4 now display the 5 track parameters from the Mackie track subview
when a track is selected and pots 8-23 will display up to the first 16 sends.


pauldavisthefirst:
manually rebased and merged. Thanks Caleb! Sorry that you won't be around anymore, really sorry.

PR #698, opened 2022-04-20
Disable sandbox in lua SessionInit scripts.

sebrautu:
Original discussion : https://discourse.ardour.org/t/no-standard-io-library-in-a-sessioninit-lua-script/107159/2


x42:
Rebased and merged as 7.0-pre0-2574 76a0b3c70e
Thanks!

PR #701, opened 2022-05-09
Fix source comment typos in `libs/ardour`

luzpaz:
Found via `codespell`

### DIFF
--- a/libs/ardour/plugin.cc
+++ b/libs/ardour/plugin.cc
@@ -592,5 +592,5 @@ PluginInfo::is_utility () const
 bool
 PluginInfo::is_analyzer () const
 {
-	return (category == "Analyser" || category == "Anaylsis" || category == "Analyzer");
+	return (category == "Analyser" || category == "Analysis" || category == "Analyzer");
 }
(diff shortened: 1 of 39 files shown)


libs/ardour/plugin.cc
```diff
@@ -592,5 +592,5 @@ PluginInfo::is_utility () const
 bool
 PluginInfo::is_analyzer () const
 {
-	return (category == "Analyser" || category == "Anaylsis" || category == "Analyzer");
+	return (category == "Analyser" || category == "Analysis" || category == "Analyzer");
```
luzpaz, line 595: Bug fix?

x42: Maybe. LV2 and VST3 uses URIs/IDs to identify categories, this way there is no localized text needed.
Perhaps we should make those strings translatable, but I don't know how VST2 handles this.

x42:
Rebased and merged as 7.0-pre0-2691 36ae8e7022
Thank you for this contribution!

libs/ardour/plugin.cc
luzpaz: Pleease advise on how to proceed?

luzpaz:
The rest of the typos were merged in 1e640563d6664fd926092cda9579465960e331b3

PR #704, opened 2022-05-12
Export dialog: update state after "Select All" and "Deselect All"

kiilerix:
The state was inconsistent after using the "All" buttons, causing
confusing behaviour. Most visibly, the message "No timespan has been
selected!" message and the disabling of the "Export" button were not
cleared.

Fixed by making
ExportTimespanSelector::set_selection_state_of_all_timespans call
update_timespans and CriticalSelectionChanged , like for example
ExportTimespanSelector::update_range_name and
ExportTimespanSelectorMultiple::update_selection do.


x42:
Nice find and fix. merged as 7.0-pre0-2705 2a2b2c8b36

PR #705, opened 2022-05-13
Remember last route type

itmuckel:
In the Add Tracks/Busses dialog the last chosen route type will be remembered now. It works only for the builtin types.
Also it works only for ardour as it looked to me as if there were some default selections made for Mixbus.

The other two commits remove some unused imports and unimplemented methods and can be dropped if you don't want them.

### DIFF
--- a/gtk2_ardour/add_route_dialog.cc
+++ b/gtk2_ardour/add_route_dialog.cc
@@ -361,6 +361,10 @@ AddRouteDialog::on_response (int r)
 	UIConfiguration::instance().set_show_on_cue_page (show_on_cue_page());
 	UIConfiguration::instance().set_insert_at_position ((int) insert_at());
 
+	/* save last route type */
+	auto* node = _session->extra_xml ("AddRouteDialog", true);
+	node->set_property (X_("LastRouteType"), enum_2_string (type_wanted ()));
+
 	reset_name_edited ();
 	/* Don't call ArdourDialog::on_response() because that will
 	   automatically hide the dialog.
@@ -567,7 +571,7 @@ AddRouteDialog::get_template_path ()
 AddRouteDialog::TypeWanted
 AddRouteDialog::type_wanted()
 {
-	if (trk_template_chooser.get_selection()->count_selected_rows() != 1) {
+	if (trk_template_chooser.get_selection ()->count_selected_rows () != 1) {
 		return AudioTrack;
 	}
 	TreeIter iter = trk_template_chooser.get_selection ()->get_selected ();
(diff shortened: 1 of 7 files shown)


gtk2_ardour/add_route_dialog.cc
```diff
@@ -598,11 +591,31 @@ AddRouteDialog::type_wanted()
 	} else if (str == _("Foldback Busses")) {
 		return FoldbackBus;
 	} else {
-		assert (0);
```
itmuckel, line 601: I ran into problems when selecting a non-builtin-type and trying to save/load it. I defaulted to AudioTrack, but maybe this is not what we want.

gtk2_ardour/add_route_dialog.cc
```diff
@@ -598,11 +591,31 @@ AddRouteDialog::type_wanted()
...
+			return _("Audio Tracks");
+		case VCAMaster:
+			return _("VCA Masters");
+		case FoldbackBus:
+			return _("Foldback Busses");
+		default:
+			return _("Audio Tracks");
+			break;
+	}
+}
```
x42, line 617: These should probably not be translatable. In general the state should not contain translatable strings. We use `X_("..")` to indicate that.

x42:
The mixbus special case can go away. We don't ship the "Create Audio Tracks Interactively" script anymore.

gtk2_ardour/add_route_dialog.h
```diff
@@ -59,7 +59,7 @@ class AddRouteDialog : public ArdourDialog
 {
 public:
 	AddRouteDialog ();
-	~AddRouteDialog ();
+	~AddRouteDialog () override;
```
x42, line 62: Is there a reason to keep this? The d'tor does nothing, so we should just remove it. Fewer LOC is better :)

x42:
All in all. This is a very really nice cleanup and well done feature.
I don't yet see an easy way to make the state untranslatable. 
Unless you have a good idea, I'm tempted to just merge it and sort this out in a follow up commit.

pauldavisthefirst:
The usual way to put enum types into untranslatable state strings is to register the enum in the appropriate enums.cc file, then use enum_2_string()

gtk2_ardour/add_route_dialog.h
```diff
@@ -47,6 +47,7 @@
 #include "ardour/plugin.h"
 #include "ardour/types.h"
 #include "ardour/template_utils.h"
+#include "ardour/route_group.h"
```
itmuckel, line 50: I needed this import otherwise it wouldn't compile. Don't know why.

x42:
Rebased and merged as 7.0-pre0-2714 2a4ebd17d6
Thank you very much!

PR #706, opened 2022-05-14
Muted regions now have '🔇' instead of '!'

itmuckel:
The second commit removes unused includes, replaces `0` with `nullptr` and replaces most iterators with range based fors (where possible).


x42:
Rebased and merged as 7.0-pre0-2737 498a3d90c5

PR #707, opened 2022-05-17
Fix ControlList::_x_scale to avoid ratio overflow when adjusting fade length

kiilerix:
Region fades would sometimes get in a mode with weird behaviour. They would be drawn in 2d with crossing lines, mainly moving back and forth horizontally - not as a function of time. It would sound as weird as it looked. The fade would sometimes jump around when resizing. It could be worked around by resetting the fade shape. It turned out the problem could be reproduced by making minute long fades.

This change fixes or works around the problem.

Back story:

timepos_t (in temporal/timeline.h) uses 62 bit for integer value and the max value is thus 4611686018427387904 ~ 5e18.
timepos_t counts superclocks, where superclock_ticks_per_second is 56448000 ~ 6e7. It can thus store up to 8e10 seconds - thousands of years.

ratio_t (in temporal/types.h) can represent fractions as 64 bit (signed) numerator and denominator.
timepos_t avoids floating point operations, but has operator* with ratio_t. To avoid crazy loss of precision it will multiply the superclock count with the numerator before dividing with the denominator.

Audio region fade in and out uses a number of increasing timepos_t values (in a ControlList) up to the length of fade. When dragging to resize, these values are (in_x_scale) multiplied with the ratio_t of the new and old fade length.
The problem is that the 62 bits will overflow if using fades more than sqrt(5e18) ~ 2e9 superclock ticks ~ 38 seconds. It will overflow into the "beat" flag and (at 58 seconds) into the sign bit. The timepos_t values in the fade will thus jump and can be negative or change to count beats.

To work around that problem, this changeset just use floating point values for scaling the timepos_t values. All scaled values are stored as integer anyway, so it should not make any actual difference for this use case. There might however be other uses of ControlList where it matters.

As an implementation detail of this "workaround" of using double, it could perhaps also be nice to implement timepos_t operator* (or operator*=) for double. But I'm not sure we want floating point support in timepos_t.

An alternative (and better) solution would be to convince the fraction multiplication to use 128 bits. It is essential to avoid overflow - mainly in static analysis, alternatively as runtime checks or asserts. It is fair enough that data types can overflow, but very unfortunate when it happens inside utility functions - such as scaling with ratios.

Perspective / further work / ideas:

It would be nice if timepos_t had detection of overflow - at least in debug mode.

Also, it would be nice to have the type system enforce that these timepos_t values are positive and count superclock ticks - not beats. It backfires a bit to use timepos_t for plain unsigned "superclock tick count". It doesn't seem like much would be lost by using plain uint64_t of superclock ticks in this place.

The .ardour file format doesn't store explicitly what shape and fade length was used. It just stores the generated timepos_t values of fades, and not as "nice" XML, but in a multiline string that has to be parsed. It would be nice to store the source information instead of the generated values. That would make it possible to (re)generate the fade curve.

I have seen .ardour files with jumping and negative time values. It would be nice to detect and recover from such bad timepos_t values. Especially to make sure that the series is strictly monotonically increasing.

Right now, the timepos_t values are scaled up and down when resizing, expecting to keep the same shape. That means that once an error has been introduced, it will never go away (until the shape is reset). That fragility will be less of a problem once the root cause is fixed. It would probably be cleaner to avoid the scaling problem completely by creating a fade curve of the right size instead of scaling the old one. Computing the curves is not instantaneous. That could be mitigated by storing the curve values precomputed.

Some code snippets that demonstrate some confusing aspects when trying to pinpoint the problem::

        std::cout << "Finding the limits of int_div_round, around sqrt(2**63) = 3037000499.976" << "\n";
        int64_t i64, i64x;
        i64 = 3037000499;
        std::cout << "Playing around just below the signed overflow limit for 64 bit: " << i64 << "\n";
        std::cout << "inline: " << (i64 * i64) / i64 << "\n";
        // prints: 3037000499
        std::cout << "int_div_round: " << int_div_round (i64 * i64, i64) << "\n";
        // prints: 3037000499

        i64 = 3037000500;
        i64x = i64 * i64;
        std::cout << "Playing around just above the signed overflow limit for 64 bit: " << i64 << "\n";
        std::cout << "checking inline overflow: " << (i64 * i64) << "\n";
        // prints: -9223372036709301616 - which is correct signed value of 0x8000000008abc290
        std::cout << "checking variable overflow: " << i64x << "\n";
        // prints: -9223372036709301616 - which is correct signed value of 0x8000000008abc290
        std::cout << "inline multiplication: " << (i64 * i64) / i64 << "\n";
        // prints: 3037000500 - which is surprisingly correct
        // this is actually incorrect in signed 64 bit - the compiler must be optimizing
        std::cout << "variable multiplication: " << i64x / i64 << "\n";
        // prints: -3037000499 - which seems wrong but is correct in signed 64 bit
        std::cout << "int_div_round: " << int_div_round (i64 * i64, i64) << "\n";
        // prints: 3037000499 - which shows surprising results from rounding when sign flips

        i64 = 3037000501;
        i64x = i64 * i64;
        std::cout << "Playing around further above the signed overflow limit for 64 bit: " << i64 << "\n";
        std::cout << "inline multiplication: " << (i64 * i64) / i64 << "\n";
        // prints: 3037000501 - which is surprisingly correct
        std::cout << "variable multiplication: " << i64x / i64 << "\n";
        // prints: -3037000496 - which seems wrong but is correct in signed 64 bit
        std::cout << "int_div_round: " << int_div_round (i64 * i64, i64) << "\n";
        // prints: -3037000497 - which shows other surprising results from rounding when sign flips

I guess there will be discussion of different ways to solve this problem. Please consider picking the first changesets first.

### DIFF
--- a/gtk2_ardour/editor_drag.cc
+++ b/gtk2_ardour/editor_drag.cc
@@ -277,8 +277,8 @@ Drag::Drag (Editor* e, ArdourCanvas::Item* i, Temporal::TimeDomain td, bool trac
 	, _constraint_pressed (false)
 	, _grab_button (-1)
 {
-
 }
+
 timepos_t
 Drag::pixel_to_time (double x) const
 {
--- a/libs/evoral/ControlList.cc
+++ b/libs/evoral/ControlList.cc
@@ -359,8 +359,9 @@ ControlList::list_merge (ControlList const& other, boost::function<double(double
 void
 ControlList::_x_scale (ratio_t const & factor)
 {
+	double double_factor = (double)factor;
 	for (iterator i = _events.begin(); i != _events.end(); ++i) {
-		(*i)->when = (*i)->when.operator* (factor);
+		(*i)->when = timepos_t::from_superclock ((*i)->when.val() * double_factor + 0.5);
 	}
 
 	mark_dirty ();
--- a/libs/temporal/tempo.cc
+++ b/libs/temporal/tempo.cc
@@ -2466,7 +2466,7 @@ TempoMap::set_state (XMLNode const & node, int version)
 
 	/* global map properties */
 
-	/* XXX this should probably be at the global level in the session file because it affects a lot more than just the tempo map, potentially */
+	/* XXX this should probably be at the global level in the session file because it is the basic time unit in session files and affects a lot more than just the tempo map */
 	superclock_t sc;
 	if (node.get_property (X_("superclocks-per-second"), sc)) {
 		set_superclock_ticks_per_second (sc);
--- a/libs/temporal/temporal/types.h
+++ b/libs/temporal/temporal/types.h
@@ -89,7 +89,7 @@ class _ratio_t {
 
 	int64_t operator* (int64_t v) const { return int_div_round (v * _numerator, _denominator); }
 
-                                        private:
+  private:
 	T _numerator;
 	T _denominator;
 };


pauldavisthefirst:
Very nice analysis, thanks very much for putting in the work to understand that.

I would probably not apply 6860199 because it's not really accurate. if you work mostly with MIDI, you will see more timepos_t that use Beats than ones using superclock_t.

Just to clarify: you are doing the multiplication as doubles in order to get 80 bits of range, rather that just 64 bits (available by using .val()) ?

We definitely do not want to make "double math" part of the timepos_t API. The whole design of timepos_t is intended to avoid every using floating point for time representation.

kiilerix:
> Very nice analysis, thanks very much for putting in the work to understand that.

Thanks. Please let me know where the analysis isn't spot on. I'm aware of being blinded by deep dives ;-)

> I would probably not apply [6860199](https://github.com/Ardour/ardour/commit/68601993e4e4424e06b01e5fb600d27400ae2634) because it's not really accurate. if you work mostly with MIDI, you will see more timepos_t that use Beats than ones using superclock_t.

Still, if there are *any* regions, they will use unsigned superclock count in several places in the `.ardour` file ( `grep '\<a[0-9]\{6,\}'` ). These values are pretty useless without a mapping to seconds. The global _superclock_ticks_per_second (as set from TempoMap `superclocks-per-second`) is thus very crucial. 

I see the superclock use as very "actual" - not at all "potential". It might indeed have potential to be used even more, but it is already very fundamental.

> Just to clarify: you are doing the multiplication as doubles in order to get 80 bits of range, rather that just 64 bits (available by using .val()) ?

I'm doing it as one (of several) ways to avoid overflow. Not so much 80 bits range, but more the 15 bit exponent. As the fade time can be scaled down to 64 and all the 32 snapshot values have to fit in that range, we really can't care about precision.

(But still, chained scaling seems very fragile. It would be better to recompute at each resize.)

I am aware that the code path I touch might be used for other purposes where my chosen solution is bad. I have however not investigated that.

> We definitely do not want to make "double math" part of the timepos_t API. The whole design of timepos_t is intended to avoid every using floating point for time representation.

Ok. But that makes it essential that the ratio multiplication uses 62+64=126 bit precision internally?

pauldavisthefirst:
manually rebased and merged. thanks very much!

kiilerix:
Thanks ... and thanks for making a better solution in 873625f77fb61e14ab2dca05c657906fdd8f6d09 .

pauldavisthefirst:
as you may notice, i've just pushed a new implementation for us on most platforms where __int128 is available. i've also removed other instances that i was aware of that compute v * (n/d). if you know of any more that I didn't address, please let me know.

kiilerix:
Thanks - that looks very nice. It clarifies several things I stumbled upon.

It could be nice to be explicit about which "supported" platforms/compilers does not have __int128 and when we expect to fall back to boost. I assume 32 bit Raspberry Pi? And Windows / VS? Where a _mul128 implementation perhaps would please a big part of the paying users? 

My remaining open questions/concerns:

The use of signed types when we know/assume values are unsigned - both in the case of "fade length" and in the general case of timepos_t. And the use of timepos_t when we know/assume it is sample/superclock based, not beats.

More specifically: The region fade times are not absolute but stored relative to clip start/end and should thus not use timepos_t (through ControlEvent / ControlList / AutomationList) but timecnt_t (or plain uint64_t)?

And I'm puzzled by the fade data model of storing low level computed values instead of the high level source data. But changing that seems to go through lots of non-trivial layers and infrastructure ;-)

pauldavisthefirst:
1. we are dropping support for all 32 bit platforms anyway. std::atomic<int64_t> on such platforms uses a mutex, which is not viable for performance reasons.
2. we use signed types to allow for out-of-bounds values
3. timepos_t is used to represent distance from a known, implicit origin (e.g. on the timeline, or inside a source). timecnt_t is used to represent distance where the origin needs to be made explicit.
4. I don't know what your final point/question refers to.

PR #709, opened 2022-05-28
fix boundary check

dancesWithBugs:
fixes a couple of boundary checks to prevent off-by-one errors


x42:
Nice find. Thanks!
manually rebased and merged as 7.0-pre0-2861 e505407b76

PR #712, opened 2022-06-11
Random minor changes

kiilerix:
Please see the individual changesets and pick and choose the good ones ... or give feedback on which changesetes should be changed or dropped.

### DIFF
--- a/gtk2_ardour/artest
+++ b/gtk2_ardour/artest
@@ -15,14 +15,14 @@ run_tests () {
 	RUN=run-tests
 	if test -n "$2"; then
 		RUN=$2
-		echo "Running test $2 of $1..."
+		echo "Running test $2 of libs/$1..."
 	else
-		echo "Running tests for $1..."
+		echo "Running tests for libs/$1..."
 	fi
 	echo "-------------------------------------------"
 	echo ""
-  $3 $LIBS_DIR/$1/$RUN || ALLGOOD=no
-  echo ""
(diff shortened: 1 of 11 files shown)


x42:
Thanks! Most of the changes are nice, but the whole queue is a bit much

* the translation will be updated by `waf i18n` before a call for translation will be issued
* ptformat should remain disabled by default
* we need to  fix the unit tests (before v7)

I guess we can cherry-pick individual commits.

kiilerix:
Thanks.
A question and a comment:
* Why should ptformat remain disabled by default?
* As far as I know, this fixes the unit tests. With this change, gtk2_ardour/artest covering libs/ is passing for me. (But it is annoyingly noisy ;-) )

x42:
> Why should ptformat remain disabled by default?

Because the feature is still in beta and also potentially problematic  for some GNU/Linux distros.

kiilerix:
> > Why should ptformat remain disabled by default?
> 
> Because the feature is still in beta and also potentially problematic for some GNU/Linux distros.

I think it would be nice to clarify the status and implications before someone enable the feature with wrong expectations. I suggest marking it "experimental" if that is what it is.
But why could it be problematic for some distros? Because of interoperability with proprietary software? Or fair use of trademarks when explaining the interoperability? Neither of these seem to be a relevant concern for this feature.

pauldavisthefirst:
all commits  manually cherry-picked into master. thanks!

PR #713, opened 2022-06-19
Superclock init

kiilerix:
The purpose of this PR is to make some clean-up around superclock, making the code more maintainable. IMO.

It will establish some invariants and a more clean pattern for how superclock and tempomap are initialized, thus mitigating the risc and maintenance cost of these global variables. It makes sure that new sessions never accidentally use default values. The default values will only be used when explicitly requested.

It fixes some problems around DEBUG_EARLY_SCTS_USE, which has the purpose of detecting use of the global _superclock_ticks_per_second variable before it has been set.
1: The code structure had changed, so it wouldn't compile if the check was enabled.
2: Fixing that, it would fail at runtime because the superclock actually was used before being set, thus relying on the global default.

Tempomap is another global variable / singleton. It was initialized before superclock was set, and would use superclock incorrectly. (That caused some bugs that has been fixed.)

Some comments to the changes in this PR:

The default superclock value of 56448000 is no longer defined in 2 places. It is however now also set explicitly in the test runner, which doesn't use the full session initialization code. (But it could perhaps be beneficial to use another superclock value there, to help make sure the variable really still can be changed. It wouldn't be a problem if they came out of sync.)

The scts_set flag is no longer used. Instead, we initialize the superclock to the obviously invalid value 0, and use that to detect if it has been set.

The setter for the global variable most_recent_engine_sample_rate was named set_sample_rate, but is very different from Session::set_sample_rate and AudioEngine::set_sample_rate / AudioBackend::set_sample_rate . It is renamed to make it more clear what it does.

The hardcoded TempoMap defaults of 120/4 are removed from TempoMap::init - it is already set in Session::create, which also is the single point where we will set the superclock for new sessions. Existing sessions are handled differently ... and AFAICS they are also handled correctly. And if not, we will find out and can make an explicit fix. 

One concern remains: What values should be used when loading a session that for some reason doesn't contain tempomap and superclock? I do think it is crucial that this fall-back handling is explicit. My refactorings might create new problems in this area, but they might also make existing problems more clear so they can be fixed.

1. TempoMap is mandatory, handled by TempoMap::set_state. Sessions before 6.0 are handled by set_state_3x, but that code path doesn't set superclock explicitly ... which seems crucial, as the global default changed in March this year.

2. In TempoMap::set_state , 46117303da7 made superclocks-per-second seems optional. I guess it should bail out explicitly if it is missing.

3. In TempoMap::set_state , Tempos and Meters also seem optional, but since the maps are cleared before trying to read, they are essentially mandatory, and it should bail out if they are missing.

Also, slightly related: Session::set_state makes sample-rate optional in session files, but it seems like a crucial synchronization with the audio engine thus is skipped. Are sessions that predates sample-rate (before 2.3 - bd3b9d763b0 in 2008) really still supported? Or should sample-rate be considered mandatory now? 

This is a follow-up to https://github.com/Ardour/ardour/pull/711 trying to bring remaining good parts forward.
See the individual commit messages.

### DIFF
--- a/libs/ardour/session_state.cc
+++ b/libs/ardour/session_state.cc
@@ -650,6 +650,8 @@ Session::create (const string& session_template, BusProfile const * bus_profile,
 		}
 
 	} else {
+		set_superclock_ticks_per_second (56448000); /* 2^10 * 3^2 * 5^3 * 7^2 */
+
 		(void) TempoMap::write_copy(); /* we are going to throw away the return value and replace the map entirely */
 		TempoMap::WritableSharedPtr new_map (new TempoMap (Tempo (120, 4), Meter (4, 4)));
 		TempoMap::update (new_map);;
--- a/libs/ardour/session_state.cc
+++ b/libs/ardour/session_state.cc
@@ -1718,7 +1718,7 @@ Session::set_state (const XMLNode& node, int version)
 				if (AudioEngine::instance()->running ()) {
 					error << _("Session: Load aborted due to sample-rate mismatch") << endmsg;
 				} else {
-					error << _("Session: Load aborted since engine if offline") << endmsg;
+					error << _("Session: Load aborted since engine is offline") << endmsg;
 				}
 				ret = -2;
 				goto out;
--- a/gtk2_ardour/editor_audio_import.cc
+++ b/gtk2_ardour/editor_audio_import.cc
@@ -284,7 +284,7 @@ Editor::import_smf_tempo_map (Evoral::SMF const & smf, timepos_t const & pos)
 		return;
 	}
 
-	/* we have to create this in order to start the update process, but
+	/* we have to create this in order to get a write lock, but
 	   we're going to throw it away by creating our own new map and
 	   populating it. This will go out of scope when we return from this
 	   method.
--- a/libs/ardour/audioengine.cc
+++ b/libs/ardour/audioengine.cc
@@ -198,7 +198,7 @@ AudioEngine::sample_rate_change (pframes_t nframes)
 	}
 
 	SampleRateChanged (nframes); /* EMIT SIGNAL */
-	Temporal::set_sample_rate (nframes);
+	Temporal::set_most_recent_engine_sample_rate (nframes);
 
 #ifdef SILENCE_AFTER_SECONDS
 	_silence_countdown = nframes * SILENCE_AFTER_SECONDS;
--- a/libs/ardour/test/testrunner.cc
+++ b/libs/ardour/test/testrunner.cc
@@ -10,6 +10,7 @@
 #include <cppunit/BriefTestProgressListener.h>
 
 #include "pbd/debug.h"
+#include "temporal/tempo.h"
 #include "ardour/ardour.h"
 #include "test_ui.h"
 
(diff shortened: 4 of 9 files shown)


libs/ardour/session_state.cc
```diff
@@ -650,6 +650,8 @@ Session::create (const string& session_template, BusProfile const * bus_profile,
 		}
 
 	} else {
+		set_superclock_ticks_per_second (56448000); /* 2^10 * 3^2 * 5^3 * 7^2 */
```
x42, line 653: Old sessions (v6 or earlier) may not have a super-clock as part of the session-state.
Superclock needs to be set in that case

x42:
All this relates to a discussion that we postponed a year a year ago: Why is superclock/sec even a variable?
Compare to MPEG (which puts 300000 in the spec). I think it should e a constant, which also allows for better optimized code.

pauldavisthefirst:
It's a variable because the value has already been changed once. At one point it was a const expression.

libs/ardour/session_state.cc
kiilerix: A very high level comment:
I haven't been able to spot a single point where I can say that HERE all global(ish) variables have been set. There is not even a single file that takes care of all cases. New sessions and existing sessions are handled in very different ways. Admittedly, it works, and all reported bugs have been fixed. (Even though I think some bugs still are lurking if creating "new" from an existing session, and some global values are inherited.)
These changes are trying to clean it up a bit, but without proposing architectural changes that I wouldn't dare to propose.

This function is for creating *new* sessions. And this chunk is in the code path is for *not* using a template. This is the point where the global tempomap is initialized, and it seems like it should be an invariant that right before doing that, the superclock should be (re)initialized. The comment about old sessions thus doesn't seem trivially applicable. Please elaborate if it still applies. In either case, it seems like supportive evidence that these global variables (and their initialization if we can't get rid of them) could benefit from a cleanup.

kiilerix:
Regarding superclock as variable:

I do get that some "events" happen in "sample time", and some events happen in "beat time", and that in order to handle both correctly at runtime, we need a clock that is higher resolution than sample time.

But it seems like everything in the file format really could&should stay in their native time format: samples or beats. The superclock could be seen as a runtime concept to make these work together. That would allow the runtime superclock to be a constant, with the optimization benefits that might give. (Going back and forth between these units will be lossy. But that is inherent: In order to stay precise and flexible, you have to remain in one domain.)

I can imagine that the runtime superclock could be fixed as compile time constant, and sample-time units could be converted from file-superclocks to runtime-superclocks when sessions are loaded. In the future, sessions could perhaps use the sample rate as superclock.

(I assume that my point of view in some ways by design and definition is wrong, but I hope my point of view can help influence and clarify something, like point out problems and areas that could benefit clarification. And I do think that this PR could clean things up a bit and enable future clean-up in the future.)

pauldavisthefirst:
superclock provides us with sample-rate independence. that's the primary reason for its existence.

x42:
Ideally we would also quantize BPM, so that each beat's tick can be lossless converted to sample and back.
There is some initial work for this using `superclocks_per_note_type` to calculate tempo, but that is still incomplete.

kiilerix:
Anyway, the purpose of this PR is not to change what superclock is - just to make it more clear how it is handled.

libs/ardour/session_state.cc
x42: > This function is for creating new sessions.

You can create new sessions from templates

kiilerix: Right. This PR has a problem there ... but the problem is even more general.

New sessions based on templates are handled by copying the template to the session file, then calling Session::load_state on it as if it was any other existing session. 

But I realize that this PR fails to handle setting superclock if loading an "old" v6 session that doesn't have superclocks-per-second. 46117303da7 not only fixed potentially uninitialized variables; it also fixed a real problem of loading sessions without superclock, falling back to the existing global default.

That would need fixing.

kiilerix:
Replaced by https://github.com/Ardour/ardour/pull/735

PR #715, opened 2022-07-01
Add missing comma to luasession/wscript

luzpaz:
"Omitting a comma between strings causes implicit concatenation which is confusing in a list."


x42:
Thanks. merged as  7.0-pre0-3121 54a4002b2b
interesting that this hasn't cause any side-effetcs so far

PR #723, opened 2022-09-10
README Update

chewitt1:
Renamed/Deleted README to/created README.md with added program info.

### DIFF
--- a/README
+++ b/README
@@ -1,5 +1,0 @@
-Please see the Ardour web site at https://ardour.org/ for all documentation..
-
-For information on building ardour: 
-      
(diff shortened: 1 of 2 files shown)


x42:
I agree that the current readme is a bit sparse.

Why mix HTML, markdown and plain-text (links) in the same file? Ideally it would all be markdown.

The "Resources" section is currently redundant (just a single link that is also present elsewhere).
Perhaps all links should rather be consolidated there instead of being spread out? and a link to the forum (discourse.ardour.org) should added.   "Technologies & Languages" should go away

PR #726, opened 2022-10-03
Adding Catalan translation

xavivars:


pauldavisthefirst:
This is great and could be in time for the 7.0 release. Do you plan to do the translations for the libraries, in particular libardour?

Thank you very much for the effort and work.

xavivars:
Hi! 

I'm actually acting as a proxy for the actual translator of the file. I'm part of Softcatalà, a non-profit that (amongst other things) translates a bunch of open-source into Catalan. We're in charge of software like Mozilla Firefox & Thunderbird, LibreOffice or Gnome. 

One of our volunteers wanted to translate Ardour, so I generated the PO and handled it to him so he could do the translation. Our plan is to completely translate Ardour, but it will probably require a little bit of time. 

I saw some closed tasks from a few years ago where you (as a dev team) didn't want to use online translation tools for Ardour. But in some cases (Catalan is an example) it's pretty hard to find someone who is both tech-savy enough to generate the PO files, and submit PRs, and also with enough time and interested enough in specific software to make the translation. That's why, in this case, I'm doing the former and Albert Pla is doing the latter. So knowing that you don't want to use Transifex or something like that, would it be possible to at least provide the POT file directly so having a build environment is not a requirement?

Thanks!

prokoudine:
@xavivars This is an interesting request as I don't think any CI has been set up to do this sort of thing. I _think_ it's entirely possible to setup a GH action to run the update script and store a POT file as an artifact for downloading. Might need someone with more experience than me, or just wait till I have the time and mental capacity to look into this.

Meanwhile, should I merge this translation now?

xavivars:
If you could merge it, I would really appreciate it, so we have a starting point. I'll try to generate the remaining po files so we can have Ardour fully translated into Catalan

prokoudine:
Merged yesterday, I think you can close it unless you want to use this PR for more updates

PR #727, opened 2022-10-03
Toggle ruler fixes

kiilerix:
The goal of this PR is to fix some small and less small issues, and clean up some annoyances:
- There were no menu option to disable the new Cue Marker ruler.
- The existing menus used naming different from the actual rulers, and they were shown in a different order.

Also, the related code were in random order, making it harder to figure out exactly what as going on maintain consistency. Clean that up to make it more maintainable.

### DIFF
--- a/gtk2_ardour/ardour.menus.in
+++ b/gtk2_ardour/ardour.menus.in
@@ -528,6 +528,7 @@
         <menuitem action="toggle-loop-punch-ruler"/>
         <menuitem action="toggle-cd-marker-ruler"/>
         <menuitem action="toggle-marker-ruler"/>
+        <menuitem action="toggle-cue-marker-ruler"/>
         <separator/>
         <menuitem action="toggle-video-ruler"/>
       </menu>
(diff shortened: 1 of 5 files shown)


kiilerix:
```
Oct 03 21:17:12 <las>   kiilerix: i'd like to commit that PR for 7.0 but the string freeze is a quandry
Oct 03 21:17:19 <las>   on the one hand, we're supposed to be in one
Oct 03 21:17:26 <las>   on the other hand, very few translation updates
Oct 03 21:17:29 <kiilerix>      but there are no new strings
Oct 03 21:18:06 <las>   ah, i was reading it incorrectly
Oct 03 21:18:54 <kiilerix>      but where is RulerMenuPopup used?
Oct 03 21:19:12 <kiilerix>      (talking about https://github.com/Ardour/ardour/pull/727 )
Oct 03 21:19:30 <las>   kiilerix: it may not be anymore :)
Oct 03 21:20:52 <las>   kiilerix: we use a dialog now with check boxes
Oct 03 21:21:02 <las>   kiilerix: years ago it was a menu
Oct 03 21:21:12 <kiilerix>      yes - I touch that one too
Oct 03 21:21:19 <las>   kiilerix: but they are awkward to use when modifying multiple items
Oct 03 21:21:25 <las>   kiilerix: we could just remove it
```
(The `RulerMenuPopup` cleanup can follow later.)

x42:
Rebased and merged as 7.0-101 34527525be
Thank you!

PR #728, opened 2022-10-04
Automation line fixes

kiilerix:
The goal of this PR is to fix a couple of issues related to editing of automation lines:
- a bug preventing dragging some automation lines
- a usuability issue making dragging of automation lines hard - apparently harder than intended

One question to the existing code remains: Why do we need polyline clamping to upper bound but not to lower bound?

### DIFF
--- a/libs/canvas/poly_line.cc
+++ b/libs/canvas/poly_line.cc
@@ -153,6 +153,9 @@ PolyLine::covers (Duple const& point) const
 	Points::size_type i;
 	Points::size_type j;
 
+	double squared_threshold = _threshold + _outline_width;
+	squared_threshold *= squared_threshold;
+
 	/* repeat for each line segment */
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
From a code perspective I like it, but clamp2() should be static and abs() should likely be fabs() (the C++ standard does not seem to provide a templated version that handles any type)

Will need to defer trying it out till after 7.0

kiilerix:
Thanks for the comments - fixed that.

But I'm still not sure if we need clamping of upper bound at all, when we apparently are fine without clamping to the lower bound?

Ok to postpone to 7.0, but note that it fixes an easy-to-reproduce UI bug.

x42:
Rebased and merged as 7.0-28-g0b27d6e652 Thank you!

PR #730, opened 2022-10-05
Https

kiilerix:
Avoid MITM and other problems from using plain http.

The build scripts would perhaps also be important to update.

### DIFF
--- a/README-GITHUB.txt
+++ b/README-GITHUB.txt
@@ -9,4 +9,4 @@ on `git describe` (a full git clone is needed for this) or alternatively a
 version-file in the release-dist. Github cannot provide the latter and also
 does not allow to inject a version-file during export archiving.
 
-For further information see http://tracker.ardour.org/view.php?id=7328
+For further information see https://tracker.ardour.org/view.php?id=7328
--- a/TRANSLATORS
+++ b/TRANSLATORS
@@ -89,7 +89,7 @@ master branch. That should generate a file called something like
 
 3. Run 'bzip2 0001-Update-German-translation.patch' to compress the file.
 
-4. File a new bug report at http://tracker.ardour.org and attach the resulted 
+4. File a new bug report at https://tracker.ardour.org and attach the resulted
 compressed file and tell either rgareus (Robin Gareus) or las (Paul Davis) on 
 the project's IRC channel (#ardour at irc.freenode.net) about your patch.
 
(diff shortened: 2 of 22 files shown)


kiilerix:
```
Oct 05 23:10:09 <las>   kiilerix: we have http -> https redirection turned on in nginx
Oct 05 23:10:16 <las>   kiilerix: no need for it in every URL
Oct 05 23:10:46 <x42>   except ping-back still uses http (since we cannot rely on certs with ardour binaries)
Oct 05 23:13:28 <kiilerix>      las, but server side redirection to https doesn't help much; in the attack scenario where someone can intercept and redirect connections, they can modify the http request to redirect to some rogue site or proxy the https away
Oct 05 23:13:38 <las>   kiilerix: fair enough
```
I maintain that everybody should use https everywhere (with few exceptions like localhost or very trusted private networks or where other higher level encryption is used). There might be places we the change isn't trivial, but that shouldn't prevent us from doing the right thing in other places.

x42:
The problem is that we cannot rely that our official binaries can find certs on the target system. (some distributions have non-standard setup).

There is nothing to hide (no credentials, etc) on requests that Ardour makes and there is an additional benefit that our server load is reduced if no additional encryption is performed.

x42:
I have no strong opinion about the README URLs, but all source-code URLs (incl PBD::open) should remain opt-in.

kiilerix:
To the best of my knowledge, this PR is only proposing change of URLs that users will open in their web browser. If a users browser can't handle https, they really shouldn't try to connect to the internet. I thus don't see the relevance of the objection.

Every single unencrypted remote URL that Ardour uses / refers to is a a liability and tech debt. I guess it eventually will end up as a CVE and embarrassment. It will serve you (and us, your users) well to try to minimize that debt. This PR is one small step, that makes the remaining problem a bit more clear.

But regarding the **different** problem of changing **other** URLs to https:
I sympathize with the difficulty of integrating with system security infrastructure in a cross-platform binary. But it is something others have done before. And while not perfect, I think it will be a fine solution to fall back to bundled root certificates. And for connections back to your own servers, you could even "hardcode" the fingerprint.

x42:
Point taken. rebased and merged as 7.0-96 74584db195 
I have split the last commit into smaller chunks, mainly because we have the general rule to not change library and UI code in the same commit.

PR #731, opened 2022-10-06
Nice labels for export analysis spectrogram

kiilerix:
The goal of this PR is to make the time labels on the export analysis spectrogram use nice round 1/2/5 units so it is easier to interpolate the location of events found by inspection or listening.

Before, the X-axis labels would be placed in nice even positions, but
with whatever odd time corresponded to that. That made it harder than
necessary to read the graph and approximate when things happened.

Instead, round the interval down to nearest power of ten ... and if
suitable, scale that up with a factor of 2 or 5. That will (with the
necessary handling of how seconds/minutes/hours relate) make sure that
the time labels are nice with a minimal amount of non-zero digits. That
makes it easy to do math and interpolate when reading the graph.

After a couple of refactorings, the actual change is simple.


x42:
Rebased and merged as 7.0-13 18527e4056
Thank you!

PR #732, opened 2022-10-07
Finding ffmpeg

kiilerix:
Nice to have for 7.0 ...

### DIFF
--- a/libs/ardour/export_format_specification.cc
+++ b/libs/ardour/export_format_specification.cc
@@ -446,10 +446,10 @@ ExportFormatSpecification::set_state (const XMLNode & root)
 		 * hardcode them here for the time being.
 		 */
 		if (format_id() == F_FFMPEG) {
-			_codec_quality = -2; // ExportFormatOggVorbis::default_codec_quality();
+			_codec_quality = -2;
 		}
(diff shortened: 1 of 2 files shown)


x42:
We cannot rely that a distro provided ffmpeg has a compatible commandline interfaces.

x42:
I have manually applied the first part. Thanks!

kiilerix:
> We cannot rely that a distro provided ffmpeg has a compatible commandline interfaces.

You still have the harvid versions as primary choice. That is what you and all users of your binary will rely on. This is just offering a fallback.

Also, on Windows it already searches much wider for system installed ffmpeg. This is just offering pretty much the same ting on Linux/Mac. Why is this change different?

x42:
ffmpeg does not have a stable CLI interface, and it breaks regularly. Many versions do not work and we should not even try.  In addition to that, codecs and video-filters (required to add black lead in/out) are compile time options.

On Windows, the binaries we ship are named ffmpeg.exe and placed in a dedicated folder. However Ardour also needs to find the binaries if a user installs harvid or xjadeo independently from Ardour. A registry key is used for this.

Unlike Unices there is no global search path for binaries, that could lead to conflicting installations.

PR #733, opened 2022-10-09
Fix disabling l10n under Linux

eudoxos:
Should fix [#8654](https://tracker.ardour.org/view.php?id=8654) (at least it does for me).


pauldavisthefirst:

PR #734, opened 2022-10-13
Fixes around export, ending up with libsndfile MP3 support

kiilerix:
The goal of this PR is to make mp3 export available without external harvid_ffmpeg ... for example when using Fedora system package. Libsndfile introduced mp3 support in version 1.1.0 - that should "of course" be used when available. It seems like you are happy with ffmpeg mp3 support, but using libsndfile seems like a good, simple, and reliable alternative.

(The problem with the format bitmasks was a tough one blocking getting mp3 working correctly. I'm  happy I nailed it ;-) )

One question: is sample rate conversion really relevant for mp3. I guess it "always" will be better to use session sample rate and leave it to the encoder to do whatever it takes to reduce bandwidth.

The main (pre-existing) TODO is a more generic way to handle quality control in ExportGraphBuilder::Encoder::init_writer .

(Also, I wonder why Vorbis quality UI enums go from 0 to 10 when vorbis_encode_init_vbr (like libsndfile) goes from (approximately) 0 to 1. It seems more intuitive to explain it as 0 to 100%. And perhaps use approximated size reduction instead of linear mapping.

### DIFF
--- a/gtk2_ardour/export_format_dialog.cc
+++ b/gtk2_ardour/export_format_dialog.cc
@@ -1113,6 +1113,7 @@ ExportFormatDialog::change_encoding_options (ExportFormatPtr ptr)
 	boost::shared_ptr<ARDOUR::ExportFormatOggVorbis> ogg_ptr;
 	boost::shared_ptr<ARDOUR::ExportFormatFLAC>      flac_ptr;
 	boost::shared_ptr<ARDOUR::ExportFormatBWF>       bwf_ptr;
+	boost::shared_ptr<ARDOUR::ExportFormatMPEG>      mpeg_ptr;
 	boost::shared_ptr<ARDOUR::ExportFormatFFMPEG>    ffmpeg_ptr;
 
 	if ((linear_ptr = boost::dynamic_pointer_cast<ExportFormatLinear> (ptr))) {
@@ -1123,6 +1124,8 @@ ExportFormatDialog::change_encoding_options (ExportFormatPtr ptr)
 		show_flac_enconding_options (flac_ptr);
 	} else if ((bwf_ptr = boost::dynamic_pointer_cast<ExportFormatBWF> (ptr))) {
 		show_bwf_enconding_options (bwf_ptr);
+	} else if ((mpeg_ptr = boost::dynamic_pointer_cast<ExportFormatMPEG> (ptr))) {
+		show_mpeg_enconding_options (mpeg_ptr);
 	} else if ((ffmpeg_ptr = boost::dynamic_pointer_cast<ExportFormatFFMPEG> (ptr))) {
 		show_ffmpeg_enconding_options (ffmpeg_ptr);
 	} else {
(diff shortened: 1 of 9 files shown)


kiilerix:
```
Oct 06 21:51:56 <kiilerix>      x42, I was surprised it didn't use libsndfile mp3 encoding now when that is available ... but I doubt that is a 7.0 change
Oct 06 21:52:31 <las>   kiilerix: sometime during 7.x
```

x42:
We cannot update to a new libsndfile (which supports mp3) until https://github.com/libsndfile/libsndfile/issues/374 is fixed.

kiilerix:
Correct, this doesn't have any benefit for **your** binaries at the moment.

This will however have benefit for users who doesn't use your binaries but build locally or use distro builds. They/we already use a more recent version of libsndfile. Yes, it is "unsupported" by you, and there might be problems with zoom recorders, but that is an existing and unrelated problem.

And as you told me in another PR: The existing ffmpeg mp3 functionaliy is somewhat fragile and cannot in general use the system ffmpeg. You rejected changes to make it work without your speciel harvid_ffmpeg. It thus seems like the best solution and the future to export mp3 with libsndfile like all other formats. You can (and will) keep using ffmpeg, and this export format will not be supported and will not show up in your builds.

And even if you personally doesn't benefit from it right now, it also doesn't harm to be prepared for the future.

Meanwhile: Please consider the early changes in this PR. I do think they provide value by their own. Can we land them, and reduce the problem to the part where we disagree?

x42:
To avoid confusion, we should only offer this option, if ardour-video-tools (ffmpeg_harvid) is not available at runtime.
It seems that libsndfile does not offer explicit bitrate control (constant/variable kbit/sec) so ffmpeg is preferable.

Also, is there a check if libsndfile is build with liblame support?

kiilerix:
Ok. I have (reluctantly) changed to only check for libsndfile mp3 if ffmpeg isn't found. Having one format depend on another might be equally confusing. I'm not fond of it.

And when libsndfile mp3 becomes universally available, and users just want to use Ardour without studying the past, I think it will be the ffmpeg variant that will seem like the outlier that cause confusion. At that point I think they will have to live side by side. (But I see how ffmpeg export still could provide value - for example if it could parse what formats were supported and make them **all** available in the UI.)

Yes, libsndfile just exposes one quality control. I agree that could seem like a disadvantage. But I guess a lossy compression algorithm could expose and infinite number of tweakable parameters. A big part of their value is that they just provide something that is a fair trade-off. Without (necessarily) going to the Apple-extreme, simplicity also has value. Offering more options doesn't necessarily make it better or preferable. And the list of options for ffmpeg mp3 could seem quite confusing and user un-friendly. Few users would be able to make a qualified choise between VBR and CBR ... if they know what it means.

I'm sure there could be a check for liblame support. But what problem would that solve? Isn't the runtime detection for actual mp3 support sufficient (and preferable to a configure check)?

x42:
Rebased and merged. The last commit was split into two (libs and GUI)

PR #735, opened 2022-10-15
Superclock init 2

kiilerix:
The goal of this PR is to clean things up around temporal initialization of global variables:
- Ensure a simple invariant by making sure the global variables only are initialized in one place.
- Avoid repeating the default superclock rate and tempo 120 in several places. Make them live in enums.cc Temporal::reset and nowhere else.
- Ensure that the global variables always are initialized when instantiating a session, without having to consider each session creation code path separately. It seems like there is an existing problem where some code paths doesn't initialize the global, and thus have an issue with superclock rates unintentionally being inherited when creating a new session from a running session.
- Avoid TempoMap initialization in Session::create , which doesn't seem to be the most obvious place to do it. That function is mainly preparing the session file in the file system, and not symmetric to how other tempomap initialization happens later when reading session files. 
- Make DEBUG_EARLY_SCTS_USE work again, and avoid dead, broken code.


pauldavisthefirst:
manually rebased & merged.

PR #737, opened 2022-10-15
Calm grey tooltip - Update dark-ardour.colors

cooltehno:
(Next try of PR)
This PR propose more calm grey background&black text instead off too much contrasting current black background&white text for tool-tip:
calm_tooltip
Current tool-tip colors are very strain on the eyes IMO.
@x42 Also this offer keeps the neutrality of colors..


x42:
This is too low contrast, similar issue as #738

PR #738, opened 2022-10-16
"Cardinal-like" tooltip - Update dark-ardour.colors

cooltehno:
This is an alternative (inverted) PR to the previous - grey background&light grey text. An idea was inspired from the Cardinal GUI:
grey+light-grey-tooltip

A Cardinal's example:
cardinal_example


x42:
deja-vu? We have discussed this before. The default theme tooltip should use high-contrast white on black (like all desktop envs do).

x42:
The tooltip should also be clearly separated, a gray background like most of the default dark background does no help.

cooltehno:
Thanks for response, @x42!
It's my subjective view, I feel too much dark when go through the GUI.. Mixbus's tool-tip is looking more calm to my perception.
Anyway thanks! :)

PR #741, opened 2022-10-20
JACKAudioBackend::join_process_threads: Don’t pass NULL for jack_client_stop_thread

bluebear94:
This is needed for PipeWire’s JACK frontend, which implements this function and expects the first argument to be non-null.

Open question: What should we do when `_priv_jack` happens to be null?

### DIFF
--- a/libs/backends/jack/jack_audiobackend.cc
+++ b/libs/backends/jack/jack_audiobackend.cc
@@ -876,6 +876,8 @@ JACKAudioBackend::create_process_thread (boost::function<void()> f)
 int
 JACKAudioBackend::join_process_threads ()
 {
+	GET_PRIVATE_JACK_POINTER_RET (_priv_jack, -1);
+
 	int ret = 0;
 
 	for (std::vector<jack_native_thread_t>::const_iterator i = _jack_threads.begin ();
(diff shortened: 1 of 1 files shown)


kiilerix:
I can confirm that this fixes a number of
```'client != NULL' failed at ../pipewire-jack/src/pipewire-jack.c:6188 jack_client_stop_thread()```
stderr messages at shutdown. I don't know if it was correlated with some actual problems.

It seems to me like "_priv_jack happens to be null" handling is fine as it is in this PR, returning -1 instead of letting _priv_jack be null. But it should perhaps write a cerr message first, just as it does when the actual stop/join fails.

It might also be a problem that _priv_jack is retrived and checked, even if it ends up not being used. The code should perhaps be refactored to use more structured preprocessing, with each of the two branches containing a full for loop ... and only one of them getting _priv_jack.

I guess the comment "jack_client is not used by JACK2's implementation" was about pass NULL to jack_client_stop_thread. That comment might need to be updated, now when the main point is that "jack_client is sued by some JACK implementations, like PipeWire".

x42:
Rebased and merged as 7.0-84 79f55742d6
Thank you!

PR #743, opened 2022-10-22
Fedora ca - proposed fix for #9005

kiilerix:
Note: Since I don't have the official Ardour debian build environment, this hasn't really been tested. Just like the existing fix attempts not had been tested on Fedora. I can test on Fedora 36/37 when the code has been built upstream.

The existing commits for https://tracker.ardour.org/view.php?id=9005 didn't solve the problem and didn't seem to move in the right direction. Plus they introduced an insecure fallback. I thus start by reverting them.

The main problem fixed here is that the check failed because setting capath to /non-existant made all checks fail. Experiments on Fedora 36 confirms that making capath point to /etc/ssl/certs works fine. I thus suggest using the default on all platforms.

The check for ca-bundle.crt with FILE_TEST_IS_DIR kind of worked before. That was because the tests specified with the bitfield are ORed. The code seemed to assume they were ANDed. That is fixed too.

I think it would be a very healthy fallback to use bundled root certs, and the comment about that is thus augmented to suggest using the mozilla approach.

### DIFF
--- a/gtk2_ardour/ardour_http.cc
+++ b/gtk2_ardour/ardour_http.cc
@@ -55,18 +55,13 @@ const char* HttpGet::ca_info = NULL;
 void
 HttpGet::ca_setopt (CURL* c)
 {
-	if (ca_info && strlen (ca_info) > 0) {
+	if (ca_info) {
 		curl_easy_setopt (c, CURLOPT_CAINFO, ca_info);
 	}
 	if (ca_path) {
 		curl_easy_setopt (c, CURLOPT_CAPATH, ca_path);
 	}
-
-	if (ca_info && strlen (ca_info) == 0) {
-		/* not hat for you */
-		curl_easy_setopt (c, CURLOPT_SSL_VERIFYPEER, 0);
-		curl_easy_setopt (c, CURLOPT_SSL_VERIFYHOST, 0);
-	} else if (ca_info || ca_path) {
+	if (ca_info || ca_path) {
 		curl_easy_setopt (c, CURLOPT_SSL_VERIFYPEER, 1);
 	}
 }
@@ -99,9 +94,8 @@ HttpGet::setup_certificate_paths ()
 	else if (Glib::file_test ("/etc/pki/tls/cert.pem", Glib::FILE_TEST_EXISTS|Glib::FILE_TEST_IS_REGULAR)) {
 		// GNU/TLS can keep extra stuff here
 		ca_info = "/etc/pki/tls/cert.pem";
-	} else {
-		ca_info = ""; // disable cert check
 	}
+	// else NULL: use default (currently) "/etc/ssl/certs/ca-certificates.crt" if it exists
 
 	if (Glib::file_test ("/etc/pki/tls/certs/ca-bundle.crt", Glib::FILE_TEST_EXISTS|Glib::FILE_TEST_IS_DIR)) {
 		// we're on RHEL // https://bugzilla.redhat.com/show_bug.cgi?id=1053882
@@ -111,7 +105,7 @@ HttpGet::setup_certificate_paths ()
 		// Debian and derivs + OpenSuSe
 		ca_path = "/etc/ssl/certs";
 	} else {
-		ca_path = "/nonexistent_path";
+		ca_path = "/nonexistent_path"; // don't try -- just in case:
 	}
 
 	/* If we don't set anything defaults are used. at the time of writing we compile bundled curl on debian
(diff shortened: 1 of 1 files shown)


x42:
> but worst case the directory exists and doesn't prevent ca_info from working. https://bugzilla.redhat.com/show_bug.cgi?id=1053882

Except it did just do that. https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/

Can we somehow figure out if the binary runs on a new enough RHEL system that has a compatible `/etc/ssl/certs`?

Our binary packages are announced to work on "Any GNU/Linux since since circa 2016 (glibc ≥ 3.4.21)."

kiilerix:
> > but worst case the directory exists and doesn't prevent ca_info from working. https://bugzilla.redhat.com/show_bug.cgi?id=1053882
> 
> Except it did just do that. https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/

Sorry, I don't see where that blog post says that ca_path=/etc/ssl/certs prevents a correct ca_info bundle from working on RH systems. Can you spell it out for me?

(Is it the part with "I forget the details, but I believe SSL_CTX_set_default_verify_paths() can fail, but it can also 'succeed' but result in an empty set of trusted certificates"? That does indeed sound scary and non-actionable. And hopefully fixed in the openssl you ship. And it seems unlikely that passing a ca_path where it finds no certificates is worse than passing a non-existent directory. If there should be a problem like that, it would be better to hit it loud and clear so it can be fixed.)

The blog post says that ca_info and ca_path are two different ways to achieve the same thing. Providing both just provides some fake redundancy. That is why RH systems traditionally have been happy with just the ca_info bundle.

(It recommends **either** using /etc/pki/tls/certs/ca-bundle.crt as ca_info **or** /etc/ssl/certs as ca_path. We could try to move in that direction, but that might be tricky when curl seems to add a layer of defaults too. I think it remains fine to pass both, as long as ca_path directory exists (though possibly empty).)

The blog post also (outdatedly) says that RH /etc/ssl/certs doesn't contain certs in the c_rehash format. Openssl would thus not pick up any certs when ca_path points there. But that is fine, as long as the cert is found in the ca_info bundle.

And to clarify: the blogpost is outdated, and Fedora systems now generally have hashed certs in /etc/ssl/certs (where Ardour currently doesn't look). Mr Happy Assassin is the lead of RH's Fedora QA team, and he helped push https://bugzilla.redhat.com/show_bug.cgi?id=1053882 forward and make Fedora /etc/ssl/certs compatible with Debian. With Fedora 37 coming out in a few days, Fedora 35 will be desupported and "all" Fedora systems will have a /etc/ssl/certs that can be used as ca_path **instead** of using the ca_info bundle. But: that is redundant and will still not work reliably on RHEL for a decade.

> Can we somehow figure out if the binary runs on a new enough RHEL system that has a compatible `/etc/ssl/certs`?

I'm pretty sure the 7.0 or current nightly binaries with /nonexistent_path doesn't work on latest RHEL; the binaries will not even look at /etc/ssl/certs .

But I'm also pretty sure that no RHEL systems have Debian-compatible /etc/ssl/certs yet. Not even CentOS Stream. There are no new enough RHEL systems.

But I will be happy to test your builds on Fedora 36 and 37.

> Our binary packages are announced to work on "Any GNU/Linux since since circa 2016 (glibc ≥ 3.4.21)."

I wonder: Do you have some old systems around that you use for verifying this intent? Like an Ubuntu LTS and RHEL / Centos7?

x42:
The 2015 blog post notes that 

>Basically, don't use RHEL/Fedora's /etc/ssl/certs, it's a trap. Don't use /etc/ssl/certs/ca-bundle.crt, even though it does probably exist on most current Fedora/RHEL installs; 

Anyway I concede, worst case we can bump the min required version for Fedora.

Rebased and merged as 7.0-89 fac5228e50

Thanks you!

kiilerix:
> The 2015 blog post notes that
> 
> > Basically, don't use RHEL/Fedora's /etc/ssl/certs, it's a trap. Don't use /etc/ssl/certs/ca-bundle.crt, even though it does probably exist on most current Fedora/RHEL installs;

Yes, the question is what he meant with "don't use". I assume it means "don't rely on it" ... which we still don't. And since we use /etc/pki/tls/certs/ca-bundle.crt and not /etc/ssl/certs/ca-certificates.crt , we should be fine.

I assume this came in too late for the 2022-10-23 nightlies. I will verify tomorrow.

PR #748, opened 2022-11-02
PulseAudio overflow fix ... and cleanup

kiilerix:
The PulseAudio backend used to be great for my usecase of editing/mixing live recordings on my desktop pc. But it has been unusable for me for a while, with lots of xruns. It seems to be a regression in PipeWire, but Ardour seems to be pushing it unnecessarily hard into a territory of undefined behaviour. With a simple and reasonable fix it works again.

After the fix comes some minor but nice cleanup and improvements from diving into the PA backend.


x42:
Rebased and marged as7.0-197 ef5b29c6dd

PR #749, opened 2022-11-03
Export fixes proposed for 7.1

kiilerix:
It would be nice to have the exception handling in 7.1, to be prepared for new formats (such as mp3) in user's custom formats.

The rest is quite trivial.


x42:
Rebased and merged as 7.1-8 3fb90523cc
Great work, as always!

PR #750, opened 2022-11-03
Additional minor PulseAudio tweaks proposed for 7.1

kiilerix:
It would be nice to use correct naming of PulseAudio. And since the backend has been broken for some users for a while anyway, now (7.1) would be a good time to fix it.

The rest is quite trivial.

### DIFF
--- a/doc/source_tree_layout.txt
+++ b/doc/source_tree_layout.txt
@@ -114,7 +114,7 @@ libs/
  - libs/backends/
    (uses libpdb, ardouralsautil | appleutility, implements libardour interface)
    Interaction with Operating System's Audio/MIDI API:
-   ALSA, CoreAudio, JACK, PortAudio/ASIO, Pulseaudio
+   ALSA, CoreAudio, JACK, PortAudio/ASIO, PulseAudio
 
  - libs/surfaces/
    (uses libpbd, libevoral, libtimecode, libardour ;; libgtkmm2ext, libcanvas)
--- a/libs/ardour/audioengine.cc
+++ b/libs/ardour/audioengine.cc
@@ -1191,7 +1191,7 @@ AudioEngine::client_real_time_priority ()
 	if (!_backend->is_realtime ()) {
 		/* this is only an issue with the Dummy backend.
 		 * - with JACK, we require rt permissions.
-		 * - with ALSA/Pulseaudio this can only happen if rt permissions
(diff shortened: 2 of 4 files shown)


x42:
I guess it's fine to merge.

Worst case Ardour does not correctly remember the most recently used backend.
Autostart may fail, or some popup message "Settings have changed.. you may lose some i/o connections." could show.

Perhaps we also need to clean out old `<EngineStates/>`?

x42 (commented):
(re https://github.com/Ardour/ardour/pull/750/commits/db4249c1bf758ede901057c4d2cd6d00a2d3568a start stream uncorked)

I think there was an issue why this is required.

Best guess, on slower machines it can take a while between start() and until the main process thread comes around.  Start/Stop  Ardour's PA backend  interrupted other pulse applications.

kiilerix:
> I guess it's fine to merge.
> 
> Worst case Ardour does not correctly remember the most recently used backend. Autostart may fail, or some popup message "Settings have changed.. you may lose some i/o connections." could show.

If the chosen backend isn't found, it shows the Setup dialog with JACK selected. Yes, after choosing backend it warns about losing settings. But it seems to preserve settings much better than when actually switching between JACK and PA ;-)

> Perhaps we also need to clean out old `<EngineStates/>`?

I only have one PA entry in my user `config`, so I guess they are cleaned automatically (as if only known backends are written?)


> (re [db4249c](https://github.com/Ardour/ardour/commit/db4249c1bf758ede901057c4d2cd6d00a2d3568a) start stream uncorked)
> 
> I think there was an issue why this is required.
> 
> Best guess, on slower machines it can take a while between start() and until the main process thread comes around. Start/Stop Ardour's PA backend interrupted other pulse applications.

I don't know how to try to verify it. But adding 10s sleep in the place of the uncork does no harm to existing playback streams.

It seems like an important role of an audio engine to keep running while individual sources have their own underruns. I'm thus hesitant to let that unconfirmed hypothesis influence decisions. We will be in a better place if we either can forget about it or get some hard knowledge.

Also, I could see a point in uncorking *after* filling the buffer. But the existing code is uncorking with an empty buffer and waiting for playback to start when prebuf is reached. That seems to be pretty much the same as starting uncorked.

x42:
Rebased and merged as 7.1-14 4958c0d666

PR #751, opened 2022-11-04
Various Minor Fixes

kiilerix:


x42:
Merged, except for the VST2/3 Path changes and the unconditional error log

PR #753, opened 2022-11-04
Plugin locations for linux 64 bit and "ptformat" clarification

kiilerix:
Follow-up to #751 , leaving out the parts with objections.


x42:
PTformat is not clarified in  7.2-217 bb029560d1 (without using any Trademarked names)

kiilerix:
> PTformat is not clarified in 7.2-217 [bb02956](https://github.com/Ardour/ardour/commit/bb029560d1c6daffe030fe356f696578ee5297db) (without using any Trademarked names)

I guess you meant to say "is **now**"?

I agree that makes it better than before. Either way, I think the configure text also should match the menu text "Import PT session"?

But mainly, I don't get the concern about trademarked names as long as they are used fairly. I cannot imagine any problem claiming that you can read files created by a named trademarked product, as long as you don't claim to be the trademarked product. For example, LibreOffice has no problem saying "Microsoft Word" or "Apple Keynote" in the file open dialog. An Ardour already has fair use of the trademarks such as Apple and Linux. Can you elaborate on the concern?

pauldavisthefirst:
the first two (useful) commits are already in master,

PR #754, opened 2022-11-09
C++17 fixes

gonsolo:


gonsolo:
> I have addressed the first part. instead of a `register int` it should really be a `size_t`.

Thanks.

> As for the iterator, we will have to check for backwards compatibility (with gcc 6.3).

That should work: https://www.godbolt.org/z/rcnb1b7K4

x42:
Rebased and merged as 7.1-165 1cf6886775
Thank you!

PR #757, opened 2022-11-16
Fix missing glib include path.

gonsolo:


libs/surfaces/midi_surface/wscript
```diff
@@ -10,7 +10,8 @@ def options(opt):
     autowaf.set_options(opt)
 
 def configure(conf):
-    pass
+    autowaf.check_pkg(conf, 'glib-2.0', uselib_store='GLIB', atleast_version='2.73', mandatory=True)
+    autowaf.check_pkg(conf, 'glibmm-2.4', uselib_store='GLIBMM', atleast_version='2.32.0', mandatory=True)
```
kiilerix, line 14: These are already checked in the top level wscript and should thus not be necessary here. Duplicate and slightly conflicting definitions of the uselib flags could potentially be confusing. But I hope to clean that up after other wscript cleanup has landed.

kiilerix:
I wonder why glib include path is missing for you - it works fine here.

libs/surfaces/midi_surface/wscript
gonsolo: Ok, I didn't see this. I can clean this up, too.

x42:
I'm curious, does this change actually fix the problem?
Do you use a custom waf?

As @kiilerix said, this change should be a NO-OP at best.

gonsolo:
glib.h is at /usr/include/glib-2.0/glib.h under Kinetix, but GLIB is not in obj.uselib. The GLIB there should be enough then.

x42:
Nevermind.  Missed that `obj.uselib` change. That line is correct to meniton GLIB[mm]

kiilerix:
My waf is rusty, but I would assume that midi_surface would inherit this from libevoral, which could explain why it works here ... but not why it doesn't work for gonsolo.

gonsolo:
1. I added a second commit which remove the check for glib and glibmm in libs/surfaces/midi_surface/wscript. It still works.
2. libs/evoral is independent of libs/surfaces, isn't it?
3. I believe there are superfluous checks for glib and glibmm in libs/evoral/wscript.

gonsolo:
Removing the superfluous lines in libs/evoral/wscript still builds for me.

kiilerix:
>     2. libs/evoral is independent of libs/surfaces, isn't it?

yes, libs/evoral is independent of libs/surfaces, but midi_surface depends on evoral. I'm thus confused why this change is needed on your system.

>     3. I believe there are superfluous checks for glib and glibmm in libs/evoral/wscript.

Yes, there are many such superfluous checks. I have PRs pending for cleaning up wscript files, but I don't think I have added that kind of cleanup to the queue yet ;-)

x42:
Rebased and merged as 7.1-108 c3d4453bdf
Thank you!

PR #762, opened 2022-11-20
Temporal cleanup

kiilerix:
The superbeat change needs special attention.


libs/ardour/luabindings.cc
```diff
@@ -684,7 +684,7 @@ LuaBindings::common (lua_State* L)
 #endif
 		.addFunction ("magnitude", &Temporal::timecnt_t::magnitude)
 		.addFunction ("position", &Temporal::timecnt_t::position)
-		.addFunction ("origin", &Temporal::timecnt_t::origin)
+		.addFunction ("origin", &Temporal::timecnt_t::position)  // backwards compatibility
```
x42, line 687: There is no need to keep Lua API backwards compatible. For better or worse it follows libardour's API and things break on occasion.
There is also currently no script in share/scripts/*.lua that uses that API.

kiilerix: That does indeed sound "worse" ;-)
Breaking a public API unnecessarily is so much against my beliefs ...

pauldavisthefirst:
I am sorry that I left this for so long. I just tried manually resolving the conflicts, and I did not understand enough about the intent of the sequence of commits in this branch to do so successfully. 

If there's any chance of you addressing the conflicts, I would be happy to merge this before 8.0 (next week).

libs/temporal/temporal/timeline.h
```diff
@@ -40,8 +40,19 @@ class timecnt_t;
...
+/* A timepos_t designates an absolute position on the global timeline. It is
+ * measured since time zero, and it can thus only be positive. It is in one of
+ * two TimeDomains: AudioTime (wall time measured using superclock,
+ * proportional to sample count) or BeatTime (counting musical ticks, which are
+ * subdivided quarternotes or Beats, independent of tempo). Conversion between
+ * these time domains will use the global TempoMap.
+ *
+ * timepos_t is also sometimes used for storing time lengths not relative to time
+ * zero. In that case it is important to avoid using the TempoMap. Such
+ * durations can also be negative.
```
kiilerix, line 52: Note: This "extended use" of timepos_t has confused me a lot. It seems like it would benefit a lot from  having a separate type for position-less deltas.

If my understanding is wrong, then we need a clarification even more.

pauldavisthefirst: I think the final part of the comment is wrong at this point.

As your intuition suggests, it caused a LOT of problems and I believe we've eliminated all the places where we tried to do this. The comment can be edited.

kiilerix: Ok, I will take a closer look at the current situation. Can you give a hint how such position-less durations are handled now? Do they just always have a position?

Meanwhile, it would be great if you can cherry-pick the parts you like.

pauldavisthefirst: We now try to never use such position-less durations, but always use timecnt_t

Finding problem cases is non-trivial, AFAIK.

pauldavisthefirst:
manually rebased & merged.

I subsequently added a comment to TempoPoint::quarters_at_superclock() to try to clarify why it is so complex (by showing what it would look like with double math.

PR #763, opened 2022-11-23
Silence all warnings and switch on Werror

gonsolo:
There are lots of warnings generated by deprecated gtk header files as well as others.
Add a flag to waf to disable all those warnings, making it possible to switch on Werror.
After that warnings can be switched on on a case-by-case basis and fixed accordingly.
Also, switching on Werror surprisingly makes mountpoint.cc fail. Fix this.


kiilerix:
But also: Different compilers emit warnings in (more or less slightly) different situations. It is annoying to develop on one platform where everything is fine, and then see it fail on other compilers because strict Werror. Werror is manageable with mandatory validation on all platforms before landing on the main branch, but I think it can be annoying with the current development process.

So in my opinion: Compiler warnings should be considered tech debt that MUST be fixed with high priority, but not necessarily blockers for landing changes. Developers working on code will often see compiler output for the files they change, and will generally make sure their change is warning free.

x42:
Which compiler are you using, and what OS?

Here on debian GNU/Linux there are no compiler warnings (when using `--strict`), except for the few explicit '#warning' (gcc 10.2.1)

x42:
`-Werror` introduces a toolchain dependency, and is hence not an option.

gonsolo:
> Which compiler are you using, and what OS?

g++ (Ubuntu 12.2.0-3ubuntu1) 12.2.0 (Kinetic)

I saw lots of warnings regarding gtk/glib deprecation (GTimeVal, GTypeDebugFlags) and was just curious what else would be necessary to get a clean build. Switching those warning off revealed the following (amongst others):

* mountpoint.cc (fixed in the commit, I don't know why that is running through without -Werror)
* declarations in pca.c
* use-after-free in smf_save.c

> `-Werror` introduces a toolchain dependency, and is hence not an option.

I can comment it out. This way it can be switched on from time to time to see what's going on. Also it is possible to remove the flags in quiet-build one-by-one and see whether the specific warning can be fixed. -Werror is especially nice for this since it stop compiling immediately.

But of course it's totally up to the main developers.

I'm curious why you don't see those warnings on Debian.

gonsolo:
BTW: I'm configuring with
`CXX='ccache g++' CC='ccache gcc' ./waf configure --optimize --compile-database`
and build with
`./waf`

gonsolo:
Since not everybody seeing the same warnings and the general disagreement on Werror I'm closing this. I'll keep it in my branch though.

x42:
Could it be that some the Werror patch prevents detection of `getmntent`?

Can you check `build/libs/ardour/libardour-config.h` if it has `#define HAVE_GETMNTENT 1`. on GNU/Linux it should definitely be available.

FWIW, I use `./waf configure --strict --ptformat --cxx11 --use-lld --backtrace --libjack=weak --maschine --compile-database`

libs/qm-dsp/maths/pca/pca.c
```diff
@@ -27,6 +27,11 @@
 
 #define SIGN(a, b) ( (b) < 0 ? -fabs(a) : fabs(a) )
 
+void covcol(double** data, int n, int m, double** symmat);
```
x42, line 30: It makes more sense to instead declare those methods as static instead.
```diff
diff --git a/libs/qm-dsp/maths/pca/pca.c b/libs/qm-dsp/maths/pca/pca.c
index 1a7bfdd549..0db27e152b 100644
--- a/libs/qm-dsp/maths/pca/pca.c
+++ b/libs/qm-dsp/maths/pca/pca.c
@@ -30,7 +30,7 @@
 /**  Variance-covariance matrix: creation  *****************************/
 
 /* Create m * m covariance matrix from given n * m data matrix. */
-void covcol(double** data, int n, int m, double** symmat)
+static void covcol(double** data, int n, int m, double** symmat)
 {
 double *mean;
 int i, j, j1, j2;
@@ -89,7 +89,7 @@ return;
 
 /**  Error handler  **************************************************/
 
-void erhand(char* err_msg)
+static void erhand(char* err_msg)
 {
     fprintf(stderr,"Run-time error:\n");
     fprintf(stderr,"%s\n", err_msg);
@@ -106,7 +106,7 @@ Ref: Smith et al., Matrix Eigensystem Routines -- EISPACK Guide
 Springer-Verlag, 1976, pp. 489-494.
 W H Press et al., Numerical Recipes in C, Cambridge U P,
 1988, pp. 373-374.  */
-void tred2(double** a, int n, double* d, double* e)
+static void tred2(double** a, int n, double* d, double* e)
 {
        int l, k, j, i;
        double scale, hh, h, g, f;
@@ -184,7 +184,7 @@ void tred2(double** a, int n, double* d, double* e)
 
 /**  Tridiagonal QL algorithm -- Implicit  **********************/
 
-void tqli(double* d, double* e, int n, double** z)
+static void tqli(double* d, double* e, int n, double** z)
 {
        int m, l, iter, i, k;
        double s, r, p, g, f, dd, c, b;
```

However it should probably go upstream: https://github.com/c4dm/qm-dsp/ -- then again we're behind upstream.

gonsolo:
> Can you check `build/libs/ardour/libardour-config.h` if it has `#define HAVE_GETMNTENT 1`. on GNU/Linux it should definitely be available.

It's not set. I'll investigate.

libs/qm-dsp/maths/pca/pca.c
gonsolo: Reported here; https://github.com/c4dm/qm-dsp/issues/7. But the last commit to the project is from 2020.

gonsolo: Also, STATLIB doesn't seem to be maintained.

gonsolo:
> Could it be that some the Werror patch prevents detection of `getmntent`?

Indeed:
```
error: cast from pointer to integer of different size [-Werror=pointer-to-int-cast]
             return (int) getmntent(0);
                    ^
```

libs/qm-dsp/maths/pca/pca.c
gonsolo: > It makes more sense to instead declare those methods as static instead.
> 
> ```diff
> diff --git a/libs/qm-dsp/maths/pca/pca.c b/libs/qm-dsp/maths/pca/pca.c
> index 1a7bfdd549..0db27e152b 100644
> --- a/libs/qm-dsp/maths/pca/pca.c
> +++ b/libs/qm-dsp/maths/pca/pca.c
> @@ -30,7 +30,7 @@
>  /**  Variance-covariance matrix: creation  *****************************/
>  
>  /* Create m * m covariance matrix from given n * m data matrix. */
> -void covcol(double** data, int n, int m, double** symmat)
> +static void covcol(double** data, int n, int m, double** symmat)
>  {
>  double *mean;
>  int i, j, j1, j2;
> @@ -89,7 +89,7 @@ return;
>  
>  /**  Error handler  **************************************************/
>  
> -void erhand(char* err_msg)
> +static void erhand(char* err_msg)
>  {
>      fprintf(stderr,"Run-time error:\n");
>      fprintf(stderr,"%s\n", err_msg);
> @@ -106,7 +106,7 @@ Ref: Smith et al., Matrix Eigensystem Routines -- EISPACK Guide
>  Springer-Verlag, 1976, pp. 489-494.
>  W H Press et al., Numerical Recipes in C, Cambridge U P,
>  1988, pp. 373-374.  */
> -void tred2(double** a, int n, double* d, double* e)
> +static void tred2(double** a, int n, double* d, double* e)
>  {
>         int l, k, j, i;
>         double scale, hh, h, g, f;
> @@ -184,7 +184,7 @@ void tred2(double** a, int n, double* d, double* e)
>  
>  /**  Tridiagonal QL algorithm -- Implicit  **********************/
>  
> -void tqli(double* d, double* e, int n, double** z)
> +static void tqli(double* d, double* e, int n, double** z)
>  {
>         int m, l, iter, i, k;
>         double s, r, p, g, f, dd, c, b;
> ```
> 
> However it should probably go upstream: https://github.com/c4dm/qm-dsp/ -- then again we're behind upstream.

Updated.

libs/pbd/wscript
```diff
@@ -121,7 +121,7 @@ def configure(conf):
                 define_name='HAVE_POSIX_MEMALIGN', execute = False, mandatory=False)
     conf.check_cc(
             msg="Checking for function 'getmntent' in mntent.h",
-            fragment = "#include <mntent.h>\n int main(void) { return (int)getmntent(0); }\n",
+            fragment = "#include <mntent.h>\n int main(void) { getmntent(0); }\n",
```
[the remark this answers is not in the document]
gonsolo: Nope:
`If control flows off the end of the compound-statement of main, the effect is equivalent to a return with operand 0`
C++ working draft 6.9.3.1-5 or ISO n3690 3.6.1-5.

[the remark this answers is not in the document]
x42: That draft is newer than the compilers we support (gcc 5.2 or later, official binaries still use gcc 6)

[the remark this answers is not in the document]
gonsolo: Yes but the second (n3690) is from 2013, gcc 5.2 is from 2015, supports it (https://godbolt.org/z/a5vhW6PWx), the wording has been in the standard basically forever, gcc 4.1.2 (from 2007) supports it (https://godbolt.org/z/3fxEov9a4), clang 3.0 from 2021 supports it, it's in the C++98 standard (https://open-std.org/JTC1/SC22/WG21/docs/wp/pdf/nov97-2/body.pdf, page 43) from (ahem) 1998, and I bet you can't find a c++ compiler that *doesn't* support it.

I rest my case. :)

wscript
```diff
@@ -71,6 +71,16 @@ compiler_flags_dictionaries= {
         'generic-arm64' : '',
         # All flags required to get basic warnings to be generated by the compiler
         'basic-warnings' : [ '-Wall', '-Wpointer-arith', '-Wcast-qual', '-Wcast-align', '-Wno-unused-parameter' ],
+        # All flags for a quiet build
+        'quiet-build' : [ '-Wno-deprecated-declarations', '-Wno-unused-value', '-Wno-unused-result',
```
x42, line 75: Wow, I really think a quiet build should just set`Wno-deprecated-declarations` and `Wno-unused-but-set-variable'`, and the rest should be fixed.  Perhaps `Wno-unknown-pragmas` can remain as well.

x42: how about quiet and really-quiet?

On macOS there are many deprecated warnings (since openGL is deprecated, but it still works), and in some cases we do have "unused values" e.g.  table  `++row` counters. which are fine. So we can safely hide those.

gonsolo: I just had lots of gtk warnings about their own deprecations and was trying to get an overview about what else is warned about. The nice thing about the quiet-build approach is that you can remove one warning at a time and fix it because it's only a few cases per warning (instead of hundreds of errors flying by). The nice thing about Werror would be that no fixed warning would ever be introduced again by new code.

PR #765, opened 2022-12-12
update german translation

edogawa23:


x42:
Danke! rebased and merged as 7.1-309 d92342fc90

PR #767, opened 2022-12-28
Update French translation

trebmuh:
This includes:
- a full translation of the preferences window
- a few things here and there


x42:
Thank you! Rebased and squashed into 7.2-72 84640d4f7d

PR #768, opened 2022-12-28
Make 2 strings translatable

trebmuh:
If I'm not wrong, those strings weren't translatable and are now.
Please, double-check as I'm not a coding person.


x42:
I've fixed the mismatch bracket and merged this as 7.2-73 4cb67f83a7

PR #769, opened 2022-12-29
Group system ports by common prefix.

mkruisselbrink:
Extra "other" ("External") ports were already being grouped by their common prefix into bundles to better display ports coming from different jack clients. This commit factors out that logic into a separate method to also apply this logic to extra "system" ("Hardware") ports. This way hardware ports from different devices/clients (for example when using pipewire as jack backend) are grouped by device rather than all being listed as one bundle.


x42:
Thank you!
Rebased and merged as 7.2-71 7a4ddf5cce

PR #770, opened 2022-12-29
Use pretty names of ports not just for mono bundles, but also stereo.

mkruisselbrink:
Session::setup_bundles() creates mono and stereo bundles from hardware inputs and outputs. For mono bundles the name of the bundle was based on the pretty name of the port (if the port has a pretty name), however stereo bundles always used the indices of ports to make the name. When using pipewire (or otherwise having multiple jack clients exposing physical ports) the indices are even less meaningful than otherwise (as different devices could appear in arbitrary order), so also using pretty names for stereo bundles makes the UI less confusing in places where these bundle names are used (for example the menu when clicking on an IOButton).


x42:
Rebased and merged as 7.2-69 ef253d3fc
Thank you!

PR #771, opened 2022-12-29
Fix AutomationListPropertyTest.

mkruisselbrink:
These tests use reference files that were generated with a particular value for superclock_ticks_per_second. The default for that has changed since the last time the reference files were updated though, causing tests to fail. Rather than updating the reference files for the new default value, this makes the test not depend on the default value by hardcoding the value that was used to generate the reference files.


x42:
Thank you.

Squashed and rebased as 7.2-216 185fe497c6

PR #772, opened 2022-12-30
Re-enable lib/ardour/test/session_test.

mkruisselbrink:
This test seems to be compiling and passing just fine (when run in isolation), so turning it back on seems like a good idea. To make it pass when run as part of the full ardour test suite, this does remove the WebSockets control surface from the control surfaces test though, as that control surface messes up the event loop of the main thread, which would otherwise cause use-after-free crashes in the session test.


x42:
Thanks! Rebased and merged as 7.2-75 bf7f6386d5

PR #773, opened 2023-01-03
Fix build on FreeBSD/powerpc64*

pkubaj:
Only glibc has __ppc_get_timebase() function. On FreeBSD use the same assembly call that __ppc_get_timebase() actually executes.

This probably should be extended to musl and OpenBSD, but I have no way of checking that.


x42:
Thank you!
Rebased and merged as 7.2-207-g387cac0ec5

PR #774, opened 2023-01-06
Tempo dialog : added an estimation of accuracy to the Tap Tempo function

MrHeadwar:


x42:
Can you elaborate on the statistics?

When I send perfectly timed "taps", the delta BPM is very large for a long time. The accuracy does not converge to 100% even if the timing is perfect.

tempo

I think you want to use running variance (or stddev) instead.

MrHeadwar:
Thanks for reviewing my PR. Here is the underlying reasoning (feel free to disapprove ! ;) ) 
- The accuracy estimation seems important to me as it tells the user when he has a "reasonable" approximation and can stop tapping. It is not supposed to reach 100% but to show an error < 1 BPM. And its an "imperfect" human measure so it shouldn't reach 100%.
- On the estimation itself : we can do it "statistically" with a pure variance calculation. That was my first go, and yielded statistically correct numbers but they had no real-world meaning (and were not converging that much either).
- In the real world, when someone taps a constant tempo, he listens to the music and taps as he feels the beat. As the music continues, the errors on each beat don't add up but compensate one another. We can assume the user at least types the right number of beats, and so the error is only at the first and last beat. I took a physicist approach and postulated that the error is <= 1/2 beat at the beginning and end, so 1 beat total. 
- hence my calculation, that is (Calculated value  - Calculated value with the same formula for 1 more beat)/Calculated value.
Maybe the progress bar should change color (eg. to green) when the BPM error is less than 1% ?

MrHeadwar:
About the second point above and to correct myself : the statistical variance *gives* an information which is the accuracy of the user typing. That is a fun statistics in itself but doesn't help in the context. ;)

luzpaz:
Any progress here ?

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):

sdevisch (approved):
Approved.

PR #776, opened 2023-01-12
Fix minor errors in temporal/wscript

mkruisselbrink:


x42:
Rebased and merged as 7.2-101 b0f5fea53a
Thank you!

PR #778, opened 2023-01-18
Update clear_gray-ardour.colors - MIDI note colors more difference with background & slected notes white contour

cooltehno:
This PR changes the lower boundary of the velocity color diapason of MIDI notes to more dark - this guarantees that any value of velocity will not produce the note's color similar to the background (as it is in the current version). We have a sacrifice of the width of color range (between the 1 and 127 velocity notes) in this case - but we get the clearer visibility of the notes especially in the middle values. Also the selected notes get a white contour instead of red (probably this is the temporary solution, until the selected note fill will take a resurrection).
An illustration:
clear_gray_change_180123
Ardour - is a power!! ///


prokoudine:
Pushed to master as well, thanks!

cooltehno:
Thanks!

PR #780, opened 2023-02-04
Support AVX512F extension for Ardour's mix and peak-detect

ashafq:
1. Optimize Ardour's mix and peak-detect function to use AVX512F instruction set extension.
2. Test and analyzed performance using test fixture program located in [Github Gist](https://gist.github.com/ashafq/50880bbbf7769f4f307e9ea9c7e71cf9)
3. Confirm CPU ID detection from [GCC's cpuid.h](https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/cpuid.h#L96)
4. Add optimized routines conditionally to Ardour's build


x42:
Outstanding work!
Rebased and merged as 7.2-214 17b263dfda

PR #781, opened 2023-02-05
Fix broken build for unit test

ashafq:


x42:
Squashed and rebased as 7.2-226 13562d94f4

PR #782, opened 2023-02-08
Fix CPUID to detect AVX512F

ashafq:
CPUID is part of x86_64 ISA to query CPU features. In order to determine `AVX512F` ISA extension, `EAX` and `ECX` needs to be set to `7` and `0` respectively before invoking `cpuid` instruction. This commit also removes inline assembly for `__cpuid` in favor of using compiler provided intrinsic functions. Both GCC and clang provides `__cpuid` like function via `__cpuid_count` intrinsic.

This commit also creates a portable wrapper over compiler intrinsic functions, __cpuid and __cpuidex. `cpuid` provides base level ISA query and `cpuidex` provides extra extension information like `AVX512F`. These wrappers lean towards MSVC like API.

References:
CPUID Docs: https://en.wikipedia.org/wiki/CPUID

GCC's "docs" on __cpuid_count:
https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/config/i386/cpuid.h

Clang's docs on __cpuid_count:
https://clang.llvm.org/doxygen/cpuid_8h.html

MSVC's docs on __cpuid and __cpuidex:
https://learn.microsoft.com/en-us/cpp/intrinsics/cpuid-cpuidex

### DIFF
--- a/libs/pbd/fpu.cc
+++ b/libs/pbd/fpu.cc
@@ -37,6 +37,8 @@
 #include <intrin.h>
 #elif defined(__GNUC__) || defined(__clang__)
 #include <cpuid.h>
+#else
+#error "Unsupported compiler: need __cpuid and __cpuidex for CPU detection"
 #endif
 
 #include "pbd/compose.h"
--- a/libs/pbd/fpu.cc
+++ b/libs/pbd/fpu.cc
@@ -33,6 +33,12 @@
 #include <asm/hwcap.h>
 #endif
 
+#ifdef PLATFORM_WINDOWS
+#include <intrin.h>
+#elif defined(__GNUC__) || defined(__clang__)
+#include <cpuid.h>
+#endif
+
 #include "pbd/compose.h"
 #include "pbd/fpu.h"
 #include "pbd/error.h"
@@ -48,7 +54,6 @@ FPU* FPU::_instance (0);
 
 #ifdef PLATFORM_WINDOWS
 /* Use MSVC/mingw intrinsic to get CPUID */
-#include <intrin.h>
 static void cpuid(int reg[4], int cpuid_leaf)
 {
 	__cpuid(reg, cpuid_leaf);
@@ -59,7 +64,6 @@ static void cpuidex(int reg[4], int cpuid_leaf, int cpuid_level)
 	__cpuidex(reg, cpuid_leaf, cpuid_level);
 }
 #else
-#include <cpuid.h>
 /* use __cpuid() with the same API to match the MSVC/mingw intrinsic */
 static void cpuid(int reg[4], int cpuid_leaf)
 {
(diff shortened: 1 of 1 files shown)


libs/pbd/fpu.cc
```diff
@@ -316,9 +304,21 @@ FPU::FPU ()
 #endif
 		}
 
+		/* Get CPU extended features: Like AVX512F */
+		if (num_ids > 0) {
```
x42, line 308: This check is redundant. This can only be reached if `num_ids > 0`

libs/pbd/fpu.cc
```diff
@@ -49,35 +45,32 @@ using namespace std;
...
 }
-
-#endif /* !PLATFORM_WINDOWS */
+/* Get extended features of CPUID */
+static void cpuidex(int reg[4], int cpuid_leaf, int cpuid_level)
+{
+	__cpuidex(reg, cpuid_leaf, cpuid_level);
+}
+#else
+#include <cpuid.h>
```
x42, line 62: Do we need  a fallback to the old asm code in case `cpuid.h` is n/a?

libs/pbd/fpu.cc
```diff
@@ -27,10 +27,6 @@
 #include <stdint.h>
 #include <assert.h>
 
-#ifdef PLATFORM_WINDOWS
-#include <intrin.h>
-#endif
-
```
x42, line 33: Best to leave includes at the top.
See item 2 at https://ardour.org/styleguide.html

libs/pbd/fpu.cc
ashafq: Fixed!

libs/pbd/fpu.cc
ashafq: Possible. However, I have tested this with gcc 6.3.0, which worked. Looking at commit logs, `cpuid.h` goes back to the year [2007](https://github.com/gcc-mirror/gcc/commit/b3172cabd267a6e6dd535c565cb7be297907f9e8) on GCC, and [2011](https://github.com/microsoft/clang/commit/b816276fbbf719342540f4db696a908e2fabd6e0) for clang.  I can leave the old inline assembly, but that will require us to maintain it over time. So, I am leaning against having a fallback version at this moment.

x42: indeed new enough.

libs/pbd/fpu.cc
```diff
@@ -27,16 +27,18 @@
...
 /* Needed for ARM NEON detection */
 #include <sys/auxv.h>
 #include <asm/hwcap.h>
 #endif
 
+#ifdef PLATFORM_WINDOWS
+#include <intrin.h>
+#elif defined(__GNUC__) || defined(__clang__)
+#include <cpuid.h>
+#endif
```
x42, line 40: What happens if neither `PLATFORM_WINDOWS`  nor `__GNUC__` nor `__clang__` is defined?

x42: Previously that worked for other compilers (e.g. icc) with custom asm code.

ashafq: Changes are, there are no `cpuid.h` then. Should this be an `#error`?

x42: @pauldavisthefirst any opinion?  Should we just 
```
#error "Unsupported compiler"
```
or fall back to some asm code similar to the one already present (https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=gcc/config/i386/cpuid.h#l221)

x42:
Rebased and merged as 7.2-261 bb31125c94

PR #787, opened 2023-02-17
Fix audio source names when importing files with > 2 and <= 26 channels.

mkruisselbrink:
This code presumably intended to name sources "foo%a", "foo%b" etc, but since it was incorrectly appending the character as an integer sources instead ended up being named "foo%97", "foo%98" etc.

Also changes the branching logic to use this branch upto 26 channels, rather than just upto 25 channels, as that seems to have been the intention.

### DIFF
--- a/libs/ardour/session.cc
+++ b/libs/ardour/session.cc
@@ -4908,9 +4908,9 @@ Session::format_audio_source_name (const string& legalized_base, uint32_t nchan,
 			sstr << "%R";
 		}
 	} else if (nchan > 2) {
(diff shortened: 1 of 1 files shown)


x42:
```
static_cast<char>('a' + 26);
```
is '{' (not 'z')

The  `static_cast<char>` is a good idea, to clarify the intent.

x42:
nevermind, since `chan < nchan`. your code is correct

x42:
Rebased and merged as 7.3-56 8810c36c6e. Thank you!

PR #788, opened 2023-02-20
Fix clang build error

ashafq:
` _mm_prefetch(reinterpret_cast<void const *>(src + 256), _mm_hint(0));` fails to compile on Clang-14 target. Therefore, replacing `_mm_prefetch_()` with clang/gcc's `__builtin_prefetch()`.


x42:
Great! Rebased and merged as 7.3-28 2773199fa3

PR #792, opened 2023-03-20
Catalan localization, on behalf of Albert Homs

xavivars:


pauldavisthefirst:
Yay! Catalan version of Ardour, for spring 2023. How lovely.

Should we expect a .po file for libardour too?

prokoudine:
Merged to master, please close :)

PR #795, opened 2023-04-20
ace-fluidsynth: fix free_path initialization in save

swesterfeld:
I saw that free_path was assigned to NULL during save, but never really initialized, here is a fix.


x42:
Rebased and merged as 7.3-247 55dc3113f3

PR #796, opened 2023-04-26
Add console1 surface

dehnhardt:


pauldavisthefirst:
Holger, this looks great, thanks so much for working on it.

One small issue: were you unaware of the new MIDISurface base class that we're trying to use for (new|old) MIDI control surfaces?

dehnhardt:
I do use it - or do I get you wrong here? class Console1 : public MIDISurface

dehnhardt:
BTW, I will try to add documentation as well, because most of the controllers have different meanins - especially for Mixbus32c and I guess no one would like to figure it all out...

pauldavisthefirst:
Sorry, my mistake. I saw code for the GUI part that replicates some method names that are now private to MIDISurface. I'll keep reviewing it, but it looks great.

pauldavisthefirst:
I note that your code has the same macOS port names for the expected device h/w as the push2:

+	return X_ ("system:midi_capture_1319078870");

this should be replaced by the actual name a Console 1 uses on macOS ...

dehnhardt:
> I note that your code has the same macOS port names for the expected device h/w as the push2:
> 
> * return X_ ("system:midi_capture_1319078870");
> 
> this should be replaced by the actual name a Console 1 uses on macOS ...

I guess I need someone with a Mac to figure that out - right?

dehnhardt:
Thank you Robin for contacting Nathan. Meanwhile someone from the Mixbus forum helped me with the macOS port names.

dehnhardt:
I just added the documentation in the 'manual' repository.

pauldavisthefirst:
manually rebased and merged! Thanks for the excellent work!

pauldavisthefirst:
Commit 29d9eb38ec unneccessarily touches some files outside the real target of the changes. Could you please clean that up and we'll be happy to (re) rebase+merge it.

x42:
Please separate commits to lib and gtk2_ardour into separate commits.
Also your first commit touches unrelated files.

and ideally make sure that every commit compiles, otherwise it breaks `git bisect`.  You can do that your branch with e.g
```
git rebase master --exec ".waf"
```

x42:
It also needs to be updated for recent changes in the control-surface API: 65346496f533951b3d6dd653c6d3a2dde9cd89cc

dehnhardt:
I won't be able to do that before the next weekend. Please be patient:-)
Holger Dehnhardt
***@***.***

06.05.2023 16:59:48 Robin Gareus ***@***.***>:

x42:
Thanks for taking this feedback into consideration.
And no rush, if we stick with a bimonthly release-cycle. Ardour 7.5 isn't due before mid/end July.

dehnhardt:
@x42 and @pauldavisthefirst: I had some time to look into this and found, that I need some help: 
- The only changes in gtk2_ardour are the graphic and the export "$libs/surfaces/console1": To separate this I first need a commit with everything in libs and second a commit with these two changes to gtk2_ardour - right?
- I'm not that firm with git: How should I change the existing commits? git reset --soft 3ddf99df085f81b15597f973dc9ec4fb3034420d and then add the changed files?
- Is it ok to add the .clang-format ti the .gitignore file and if so, should this be a separate commit as well?
- The changes in the  control-surface API are only in the ControlProtocolDescriptor - right?
Thanks for your help!

dehnhardt:
Ok, I think I already have the right changes. Pls. have a look again ;-)

x42:
Very Nice. Thank you!

Rebased and merged as 7.4-63 0d2d30f1a7

PS. I have made (and squashed) two small changes:
* fixed the ControlProtocolDescriptor order: https://github.com/Ardour/ardour/pull/796/files#diff-c0e8a7bdb0448c0a35dc08844180a083588c0933de8d3511394f9a227889ca11R53-R57
 * replaced std::optional (which requires C++17) with boost::optional (which works with C++11) in console.cc https://github.com/Ardour/ardour/pull/796/files#diff-a2807597653135ee3a807dcb1434ecda0c5b3a66fccf2aaba7a0b6c99f5ffbe3R1060

dehnhardt:
Thank you Robin!

PR #797, opened 2023-05-12
Console1: With larger projects, the session was not initialised at this time

dehnhardt:
I worked on a large project and was not able to load it after the first saving. It turned out that  the session was not valid at this point.
So a npe occured.
Still strange, that it was only with large projects...
BTW: Selection still works.


x42:
Rebased and merged as 7.4-82 f958b7b0e0
Thank you!

PR #798, opened 2023-05-17
Console1 plugin interface

dehnhardt:
**This is a draft:**

I added the possibility to control arbitrary plugins. 
The mapping from the plugin controls to the hardware controls is done by a simple file for each plugin which is supported.
These files must reside in the directory "user_config_directory ()/c1mappings/". This directory is created when it does not already exists.
@pauldavisthefirst or @x42: Is it ok to do it that way? And can I add existing plugin configuration files during the installation process?

The files looks like

```
<?xml version="1.0" encoding="UTF-8"?>
<c1plugin-mapping 
        ID="urn:ardour:a-comp" 
        NAME="ACE Compressor">
        <param-mapping id="0">
                <name>Attack</name>
                <mapping shift="false">COMP_ATTACK</mapping>
        </param-mapping>

        <param-mapping id="1">
                <name>Release</name>
                <mapping shift="false">COMP_RELEASE</mapping>
        </param-mapping>

        <param-mapping id="2">
                <name>Knee</name>
                <mapping shift="false"></mapping>
        </param-mapping>

        <param-mapping id="3">
                <name>Ratio</name>
                <mapping shift="false">COMP_RATIO</mapping>
        </param-mapping>

        <param-mapping id="4">
                <name>Threshold</name>
                <mapping shift="false">COMP_THRESH</mapping>
        </param-mapping>

        <param-mapping id="5">
                <name>Makeup</name>
                <mapping shift="false">COMP_PAR</mapping>
        </param-mapping>

        <param-mapping id="6">
                <name>Sidechain</name>
                <mapping shift="false"></mapping>
        </param-mapping>
</c1plugin-mapping>
```
I added a 'shift state' in the plugin mode as well

The handling;
If the "Track Group" button is pressed, the "Plugin Mode" is entered. 
You can select plugins 1-20 for the current selected strippable with the "Select Track" buttons. 
* If a plugin exists at this position in the selected strippable,the LED of the selection blinks 
  * the LED for the selected track stays on as well 
  * if the selected plugin is at the same position as the track (number) I'm still unsure what to do ;-)
  * The plugin can be disabled / enabled by the "Mute" button
  * a configuration is looked up
    *  if a configuration is found, all other mappings are loaded
    * if you press the Track-Button again, the plugin ui is toggled
 * If no plugin exists at the selected position, nothing happens.

### DIFF
--- a/libs/surfaces/console1/console1.cc
+++ b/libs/surfaces/console1/console1.cc
@@ -67,17 +67,17 @@ Console1::~Console1 ()
 
 	tear_down_gui ();
 
-	for (const auto& [_, b] : buttons) {
-		delete b;
+    for( const auto &b : buttons ){
+		delete b.second;
 	}
-	for (const auto& [_, b] : encoders) {
-		delete b;
+    for( const auto &e : encoders ){
+		delete e.second;
 	}
-	for (const auto& [_, b] : meters) {
-		delete b;
+    for( const auto &m : meters ){
+		delete m.second;
 	}
-	for (const auto& [_, b] : multi_buttons) {
-		delete b;
+    for( const auto &mb : multi_buttons ){
+		delete mb.second;
 	}
 
 	/* stop event loop */
--- a/gtk2_ardour/ardour_ui_session.cc
+++ b/gtk2_ardour/ardour_ui_session.cc
@@ -442,7 +442,7 @@ ARDOUR_UI::load_session_stage_two (const std::string& path, const std::string& s
 
 		(void) msg.run ();
 		msg.hide ();
-		delete escaped_error_txt;
+		g_free (escaped_error_txt);
 
 		goto out;
 	}
@@ -473,7 +473,7 @@ ARDOUR_UI::load_session_stage_two (const std::string& path, const std::string& s
 
 		(void) msg.run ();
 		msg.hide ();
-		delete escaped_error_txt;
+		g_free (escaped_error_txt);
 
 		goto out;
 	}
@@ -501,7 +501,7 @@ ARDOUR_UI::load_session_stage_two (const std::string& path, const std::string& s
 
 		(void) msg.run ();
 		msg.hide ();
-		delete escaped_error_txt;
+		g_free (escaped_error_txt);
 
 		goto out;
 	}
@@ -700,7 +700,7 @@ ARDOUR_UI::build_session_stage_two (std::string const& path, std::string const& 
 		msg.set_title (_("Loading Error"));
 		msg.set_position (Gtk::WIN_POS_CENTER);
 		msg.run ();
-		delete escaped_error_txt;
+		g_free (escaped_error_txt);
 		return -1;
 	}
 	catch (Glib::Error const& e) {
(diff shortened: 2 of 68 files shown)


x42:
Please bear with me. I'd like to provide proper feedback, and constructive criticism, which will require some time for this case.

Some initial thoughts:

* Is there a reason why the control list cannot be built dynamically?
* Parameter names are not necessarily constant
* So far Ardour does not parse CSV, for better or worse Ardour stores everything using XML and there is a common code-path for all serialization.

Since the device does not have scribble strips, a *perhaps crazy* idea would be to show that on the screen instead.
When entering plugin mode, show a mockup of the Console 1 with labels on the display in a window to show what each knob does at a given time. That could even include a scrollable list of plugins (and presets) to select.

libs/surfaces/console1/console1.cc
```diff
@@ -219,159 +258,179 @@ Console1::connect_internal_signals ()
 	  console1_connections, MISSING_INVALIDATOR, [] () { DEBUG_TRACE (DEBUG::Console1, "VerticalZoomOut\n"); }, this);
 }
 
+void
+Console1::notify_session_loaded ()
+{
+	DEBUG_TRACE (DEBUG::Console1, "************** Session Loaded() ********************\n");
+	create_strip_inventory ();
+	connect_internal_signals ();
+	/*if (session) {
```
dehnhardt, line 267: I tried a lot to find out if there is already a valid session - but nothings seems to wor so far. Do you have any idea where we have a reliable session?

dehnhardt:
The extension to control arbitrary plugins is working now. However, @x42 idea of adding an image of console1 with the assignmants will take some time. However I would like to merge the current status.

libs/surfaces/console1/console1.cc
pauldavisthefirst: Control protocols are instantiated for a session, they do not exist across sessions. So, if you control protocol has been instantiated, there is a valid session (it was passed into the constructor).

dehnhardt: Thanks Paul, but why does first_selected_stripable() not return a stripable even if one is already selected here? That's what I wanted to achieve here....

pauldavisthefirst: selection probably hasn't happened yet.

dehnhardt: ;-) Ok, but is there any way to get the initial selction? stripable_selection_changed() does not seem to be called initially.
If I manually select another stripable after loading, the method is called.

dehnhardt:
Hi @pauldavisthefirst and @x42: I really would like to have this changes merged. There where several problems with the former code which resulted in some crashes. These are fixed now.
Is there anythig I need to do?
In addition a basic interface for plugins is added. 
This will be extended in the next months and I will try to add X42 suggesten regarding the on screen display.

pauldavisthefirst:
I went to rebase & merge this today, and noticed this:

../libs/surfaces/console1/console1.cc:70:19: warning: structured bindings only available with ‘-std=c++17’ or ‘-std=gnu++17’
   70 |  for (const auto& [_, b] : buttons) {
      |                   ^

We currently have a C++11 code base. If you can fix this, I'll be happy to rebase & merge before 8.0 (next week).

dehnhardt:
Hmm there went something wrong now. I need to undo something, but habǘe to go now.

pauldavisthefirst:
manually rebased & merged. thanks!

dehnhardt:
Many many thanks paul. I'm currently on vacation with my family and have very limited time.

PR #799, opened 2023-05-18
Fix getmntent check with -Werror.

gonsolo:
Fix a warning. The test is just compile tested and the return value is not used anyway.
Also, in my private build I compile with "-Werror" and without this patch this fails.

It also fails to build:
`../libs/pbd/mountpoint.cc:114:10: fatal error: sys/ucred.h: Datei oder Verzeichnis nicht gefunden
  114 | #include <sys/ucred.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
`

### DIFF
--- a/libs/pbd/mountpoint.cc
+++ b/libs/pbd/mountpoint.cc
@@ -110,6 +110,10 @@ mountpoint (string path)
 
 #else // !HAVE_GETMNTENT
 
+#ifdef __linux__
+#error "getmntent is supported on Linux. Is this a configuration error?"
+#endif
+
 #include <sys/param.h>
 #include <sys/ucred.h>
 #include <sys/mount.h>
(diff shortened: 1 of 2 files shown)


filipeborato (approved):

gonsolo:
The test program cannot be run. It always fails with a segmentation fault. A better one would be:
```
FILE *f = setmntent("/", "r");
if (f) { endmntent(f); }

```

x42:
The test is not executed, but it needs to compile without warning if OP wants to use -Werror.

kiilerix:
The current check script will however compile even if mntent.h doesn't contain getmntent. The default declaration of getmntent is fine.

The check could thus be simplified to
`msg="Checking for mntent.h", fragment = "#include <mntent.h>",`
which I assume also works fine for the use case of -Werror.

A more (pointless) thorough check that checks for valid content of mntent.h could perhaps also do
`struct mntent *tmp = getmntent(0);`

gonsolo:
I didn't compile Ardour recently but I would say yes.

sdevisch (approved):

sdevisch (dismissed):

sdevisch (dismissed):

gonsolo:
I still approve this PR, @x42! :wink:

sdevisch (approved):
Approved.

PR #800, opened 2023-06-04
German translation

edogawa23:


x42:
Rebased and merged as 7.4-192 ba019053b6

PR #802, opened 2023-06-07
Update ko.po

MarongHappy:


x42:
The commit was rebased and merged as 7.4-302 52d488e2ce, I'll leave this open for a few days to collect feedback on the Windows issue.

x42:
We "worked around" the issue by disabling translation on Windows by default. A user will have to opt-in now.

PR #806, opened 2023-07-06
Midi map for Devine Versakey

CardamomCake:
Midi map configuration for transport controls for the Devine Versakey midi controller.


x42:
Rebased and merged as 7.5-127 4fc3bd88aa
Thank you!

PR #808, opened 2023-07-14
Add a map for the Donner StarryPad drum pad controller

agraef:
Binding map for the Donner StarryPad, a cheap and cheerful 4x4 drum pad controller.

### DIFF
--- a/share/midi_maps/Donner_StarryPad.map
+++ b/share/midi_maps/Donner_StarryPad.map
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ArdourMIDIBindings version="1.0.0" name="Donner StarryPad">
+
+<!-- Created 2023-07-14 by Albert Gräf. The Donner StarryPad is a cheap and
+     cheerful 4x4 drum pad controller sold by Donner with some transport
+     buttons, two faders (F1, F2), and two encoders (K1, K2), which makes it
+     worthwhile to have it as a generic MIDI surface in Ardour.
+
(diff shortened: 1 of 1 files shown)


agraef:
Here's a picture of the thing, as it's hard to find on Donner's website:

image

I actually have one of these (so I could test the map), they're pretty nice.

agraef:
I forgot to mention this in the map comments, but it's worth noting that the StarryPad has two additional banks of knobs (CC12-15) and faders (CC22-25) accessible with the KNOB BANK and FADER BANK buttons. One might still assign these to, e.g., sends or plugin parameters, but I left them unmapped, so that they can be freely assigned with MIDI learn.

prokoudine:
Hi! Thank you :)

Quick question:

`<Binding channel="1" ctl="20" uri="/route/gain S1"/>` -- this is the F1 fader?

`<Binding channel="1" ctl="28" uri="/route/pandirection S1"/>` -- this is the K1 encoder?

`<Binding channel="1" ctl="21" uri="/bus/gain master"/> ` -- this is the F2 fader?

`<Binding channel="1" ctl="9"  uri="/bus/pandirection master"/>` -- this is the K2 encoder?

agraef:
Yes. The CCs are for the factory setup, at least that's what my unit had as default out of the box. Note that it's possible to change any of these in the StarryPad editor software, in which case you'll have to change the binding map accordingly.

prokoudine:
@agraef Pushed to master, please close the PR as I don't have the privileges :)

PR #809, opened 2023-07-21
Fix #8857: Remove snap from decision making about number of divisions

darless:
In Mantis 8857 states:


pauldavisthefirst:
The problem here is that the author of the current version of ::get_grid_music_divisions (ben loftis) intended to separate snap and grid, so that if snap is disabled, no musical grid is in effect. 

I will ask him to review this, and keep thinking about it myself.

PR #811, opened 2023-07-25
Fix #9421: F7 tied to velocity 96 which is not an option in the velocity drop down

darless:
This fixes [Mantis 9421](https://tracker.ardour.org/view.php?id=9421)

# Issue
Switching velocity in the virtual keyboard via the F(5-8) does not work when trying to go from F7 to F8, once F8 has been selected. Specifically trying to go F5 (32), F6 (64), F7 (96), F8 (127), in that order works. However, going back to F7 (96), and then attempting to go to F8 (127), leaves the actual velocity at 96 instead of 127. The reason for this is in `ArdourDropdown::set_active` which iterates through the menu items and 96 is not an available option.

# Fix
The fix is to add '96' to the menu items.


x42:
You can use the scroll-wheel to get any velocity.

x42:
Rebased and merged as 7.5-243 9f4a0b444e

Thanks!

PR #812, opened 2023-07-27
Windows build tweaks

renatolfc:
This PR contains a set of changes I had to make to the build tooling to enable compilation of Ardour 7.5 for Windows.

Of note, the script was referencing the old wavesaudio backend. I've replaced it with the portaudio backend.

Also, in the cowbuilder environment we were missing a few packages for successful packaging:
 * `ca-certificates`, for the calls to `curl`
 * `rsync` to copy some assets
 * `zip`/`unzip` for file extraction

### DIFF
--- a/tools/x-win/compile.sh
+++ b/tools/x-win/compile.sh
@@ -25,7 +25,7 @@ fi
 
 if test -z "${ARDOURCFG}"; then
 	if test -f ${PREFIX}/include/pa_asio.h; then
-		ARDOURCFG="--windows-vst --with-backends=jack,dummy,wavesaudio"
+		ARDOURCFG="--windows-vst --with-backends=jack,dummy,portaudio"
 	else
 		ARDOURCFG="--windows-vst --with-backends=jack,dummy"
 	fi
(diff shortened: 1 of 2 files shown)


x42:
LZMA works fine for us. Since these scripts are used to create the official binaries it needs to remain enabled.

You can simply set 'QUICKZIP' env variable if needed.

renatolfc:
Sounds good. I’ll update the PR.Apart from that, are the other changes appropriate?On 27 Jul 2023, at 17:52, Robin Gareus ***@***.***> wrote:﻿
LZMA works fine for us. Since these scripts are used to create the official binaries it needs to remain enabled.
You can simply set 'QUICKZIP' env variable if needed.

x42:
I can cherry pick the other two commits. Unless @pauldavisthefirst does it, it will have to wait until early next week.

renatolfc:
I think I can wait. Thanks for reviewing this and maintaining Ardour, @x42.

x42:
Rebased and merged as 7.5-242 f4414c041a

Thank you!

PR #813, opened 2023-08-17
Fix compile error introduced in rev. 7c029f5f.

agraef:
The new call of `PBD::stacktrace()` in rev. 7c029f5f trips a compile error (at least here on Arch Linux with gcc (GCC) 13.2.1 20230801), move the `#include "pbd/stacktrace.h"` in editor.cc up to avoid it.


pauldavisthefirst:

PR #814, opened 2023-09-06
Kurzweil SP4 MIDI name document

infinitekh:
It is a common instrument that is often placed in churches because it is inexpensive.
Information was obtained from the Kurzweil website.


pauldavisthefirst:
manually rebased & merged, thanks!

PR #815, opened 2023-09-08
Keep the search term in the dialog box.

automaciej:
When user wants to keep the dialog box open, it's probably with the intention to edit the search term rather than typing it from scratch again.


x42:
Nice! Thanks.

Rebased and merged as 7.5-494 1c8b91ed86

PR #816, opened 2023-09-09
Many quantization fixes

asahilina:
Fix a bunch of problems in the quantization feature (I think most of these are regressions from nutempo):

- Grid selection was broken due to not actually translating into the beat counts
- Threshold was broken as it was not in terms of ticks, but rather beats (should this be changed to a fraction?)
- Swing was really broken

There's still some weird offset problem with quantizing to septuplets, I'm not sure what's going on there... but at least all the common cases work properly now.


pauldavisthefirst:
Thanks very much for grapping with some of the most arcane code in the whole project. Welcome!

PR #817, opened 2023-09-17
update german translation

edogawa23:


x42:
Rebased and merged as 7.5-581 9aaf5d2c8c

PR #818, opened 2023-09-17
autowaf: fix escaping with Python 3.12

tartina:


kiilerix:
LGTM - very nice to have for Ardour 8, for example for Fedora.

But also, this is in a part of autowaf that Ardour doesn't seem to use. It is unfortunate that Ardour has to maintain it. Ardour should either strip autowaf down to the bare minimum that actually is used, or blindly vendor an upstream version. It would perhaps be better to remove this chunk than to fix it.

x42:
We should just remove `make_simple_dox`. it's not used at all.

kiilerix:
> We should just remove `make_simple_dox`. it's not used at all.

Agreed - that seems like a very good solution too. It would be nice to minimize the surface we have to maintain. Unless we want to rebase to a new upstream version and leave all maintenance to upstream?

pauldavisthefirst:
I removed the offending method from autowaf.py, and upgraded waf to 2.0.26 (from 2.0.23).

PR #822, opened 2023-09-22
Backport VST3 libc++ fix on linux

vimproved:
LLVM libc++ does not have the ext/atomicity.h header. This fix is copied from the upstream vst3_pluginterfaces repo.

I know in previous backports for the vst3 SDK there's been a suggestion of just updating it as a whole, would that be a better solution?


x42:
Rebased and merged as 7.5-639 8aa716d437

Thank you!

vimproved:
Getting a build failure here on libc++ actually, not sure why this wasn't happening before (possibly because I downgraded from 18 to 17). I'll file a new PR with the fix.

PR #823, opened 2023-09-22
Update Polish translation

songo:
This is a major update as previous translation was very dated and about 10% complete. This one is about 60% complete and in a pretty good shape for 8.0. Further updates soon.


x42:
Rebased and merged as 7.5-673 b745a463df
Thank you!

PR #824, opened 2023-09-23
Use <atomic> instead of <stdatomic.h> in libc++ fix

vimproved:
The definitions from <stdatomic.h> conflict with <atomic>, which causes a build failure since <atomic> is included previously.

Follow up to https://github.com/Ardour/ardour/pull/822.

### DIFF
--- a/libs/vst3/pluginterfaces/base/funknown.cpp
+++ b/libs/vst3/pluginterfaces/base/funknown.cpp
@@ -44,9 +44,9 @@
 #endif
 
 #if SMTG_OS_LINUX
-#if !defined (SMTG_USE_STDATOMIC_H)
+#if !defined (SMTG_USE_ATOMIC)
 #if defined (_LIBCPP_VERSION)
-#define SMTG_USE_STDATOMIC_H 1
+#define SMTG_USE_ATOMIC 1
 #else
 #include <ext/atomicity.h>
 #endif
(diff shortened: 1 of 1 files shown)


x42:
Where does this change originate and what problem does it solve? There are no build failures here and nightly builds are also fine?

Looks like with your recent change we follow upstream: https://github.com/steinbergmedia/vst3_pluginterfaces/blob/37de655a51b28a558645621b39d48a00292be5e2/base/funknown.cpp#L38

vimproved:
> Where does this change originate and what problem does it solve? There are no build failures here and nightly builds are also fine?
> 
> Looks like with your recent change we follow upstream: https://github.com/steinbergmedia/vst3_pluginterfaces/blob/37de655a51b28a558645621b39d48a00292be5e2/base/funknown.cpp#L38

The original bug was specific to libc++ on linux, which I don't think is covered by your builds. The original fix was following upstream, however it was still broken since libs/pbd/pbd/atomic.h includes <atomic>, which causes libs/fst/vst3-scanner.cc to fail with `reference to 'memory_order' is ambiguous` since there's both memory_order from stdatomic.h and std::memory_order from atomic.

x42:
Thanks for the explanation. We do use clang/libc++ on macOS but that not affected here.

Rebased and merged as 7.5-665 6e24a409f4

PR #826, opened 2023-09-25
Update ko.po

MarongHappy:


x42:
Rebased and merged as 8.0-rc1-4 f9da85639
Thanks!

prokoudine:
Hi @MarongHappy, could you please guide us through your process of updating translation files and editing them? There seems to be a recurring error: every new ko.po file you submit has duplicated translatable messages, which breaks the translation. Do you update the translation file in some unconventional way? Do you edit it in some unconventional software?

PR #827, opened 2023-09-27
Update Polish translation

songo:
The Polish translation is now about 80% complete and the hard work has been done (yeah... the Preferences ;-) ).


x42:
Rebased and merged as 8.0-rc2-1 3a6a70c240
Thanks!

PR #831, opened 2023-10-09
Fix ardour not loading Moog Subsequence 37 midnam

stiangrindvoll:
Realized ardour was not loading the midnam config as expected.
This ensures it does.


stiangrindvoll:
sorry about the extra pr

PR #832, opened 2023-10-12
Update FR translation for Ardour 8.x versions

jujudusud:
Merge-able before a new release.
I will continue to commit some more translations and typos or ...


prokoudine:
Hi! Do you think you could squash your commits?

jujudusud:
> Hi! Do you think you could squash your commits?

Is it ok like that?

PR #834, opened 2023-10-24
Clean init process to prevent exceptions

dehnhardt:
As suggested by Robin I have moved nearly all initialization to "start_using_device" and remove all connections in "stop_using_device".
In addition I prevent updates when not connected. (And did some proper formatting...)

### DIFF
--- a/libs/surfaces/console1/console1.cc
+++ b/libs/surfaces/console1/console1.cc
@@ -98,8 +98,6 @@ Console1::set_active (bool yn)
 
 	ControlProtocol::set_active (yn);
 
-	DEBUG_TRACE (DEBUG::Console1, string_compose ("Console1::set_active done with yn: '%1'\n", yn));
-
 	return 0;
 }
 
@@ -187,14 +185,28 @@ Console1::begin_using_device ()
 	Glib::RefPtr<Glib::TimeoutSource> periodic_timer = Glib::TimeoutSource::create (100);
 	periodic_connection = periodic_timer->connect (sigc::mem_fun (*this, &Console1::periodic));
 	periodic_timer->attach (main_loop ()->get_context ());
+	connect_session_signals ();
+	connect_internal_signals ();
+	create_strip_inventory ();
 
 	DEBUG_TRACE (DEBUG::Console1, "************** begin_using_device() ********************\n");
 
-	create_strip_inventory ();
-	connect_internal_signals ();
-	stripable_selection_changed ();
-
 	return MIDISurface::begin_using_device ();
+}
+
+int
+Console1::stop_using_device ()
+{
+	DEBUG_TRACE (DEBUG::Console1, "stop_using_device()\n");
+	if (!_in_use)
+		return 0;
+	blink_connection.disconnect ();
+	periodic_connection.disconnect ();
+	stripable_connections.drop_connections ();
+	session_connections.drop_connections ();
+	console1_connections.drop_connections ();
+	MIDISurface::stop_using_device ();
+	return 0;
 }
 
 void
@@ -407,19 +419,6 @@ Console1::setup_controls ()
 	new Meter (this, ControllerID::OUTPUT_METER_R, boost::function<void ()> ([] () {}));
 }
 
-int
-Console1::stop_using_device ()
-{
-	DEBUG_TRACE (DEBUG::Console1, "stop_using_device()\n");
-
-	blink_connection.disconnect ();
-	periodic_connection.disconnect ();
-	stripable_connections.drop_connections ();
-	console1_connections.drop_connections ();
-	MIDISurface::stop_using_device ();
-	return 0;
-}
-
 void
 Console1::handle_midi_controller_message (MIDI::Parser&, MIDI::EventTwoBytes* tb)
 {
@@ -513,6 +512,9 @@ void
 Console1::stripable_selection_changed ()
 {
 	DEBUG_TRACE (DEBUG::Console1, "stripable_selection_changed \n");
+	if (!_in_use)
+		return;
+
 	std::shared_ptr<Stripable> r = ControlProtocol::first_selected_stripable ();
 	if (r) {
 		set_current_stripable (r);
@@ -1194,11 +1196,10 @@ Console1::select_rid_by_index (uint32_t index)
 	std::shared_ptr<Stripable> s = session->get_remote_nth_stripable (rid, PresentationInfo::MixerStripables);
 	if (s) {
 		session->selection ().select_stripable_and_maybe_group (s, true, false, 0);
+	} else {
+		success = false;
 	}
-    else {
-    	success = false;
-    }
-    if( !success ){
+	if (!success) {
 		map_select ();
 	}
 }
--- a/libs/surfaces/console1/wscript
+++ b/libs/surfaces/console1/wscript
@@ -22,7 +22,7 @@ def build(bld):
     obj.includes     = [ '.', './console1']
     obj.name         = 'libardour_console1'
     obj.target       = 'ardour_console1'
-    obj.uselib       = 'SIGCPP XML OSX'
+    obj.uselib       = 'GTKMM SIGCPP XML OSX'
     obj.use          = 'libardour libardour_cp libardour_midisurface libgtkmm2ext libpbd libevoral libtemporal'
     obj.install_path = os.path.join(bld.env['LIBDIR'], 'surfaces')
 


x42:
Rebased and merged as 8.1-21 94fec7fd8a

x42:
Does this work for you? 

We have a report from Harrison that the device does not respond anymore with this patch applied.
Both ports connected, so the surface should be enabled:
image

Any idea what we could check?

dehnhardt:
I can't explain that right now. I developed it with the Mixbus code and then copied it to Ardour and tested it there with multiple connects and disconnects.
Since I'm not at home at the moment, I can't get to the bottom of the problem because I don't have the device here.
If anyone has time, the writeouts with "./ardev -D console1,midisurface" might be helpful. Otherwise, I think it's easiest to revert the commit for now, since I won't get around to fixing it until 2 November.

x42:
This is on macOS, and the user does not compile Mixbus, so it was not trivial to get debug output.

The actual issue: Moving a Fader on Mixbus, updates the console1,  but moving a knob on console 1surface does not update Mixbus. -- This apparently worked last week.


```
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::Console1: stripable_selection_changed 
DEBUG::Console1: set_current_stripable 
DEBUG::Console1: current_stripable found:  
DEBUG::Console1: current_stripable 0 - 1025
DEBUG::Console1: current_stripable: rid 0, bank 0, index 0 
DEBUG::Console1: map_shift()
DEBUG::Console1: map_select())
DEBUG::Console1: map_phase 
DEBUG::Console1: map_recenable()
DEBUG::Console1: map_solo()
DEBUG::Console1: map_gate_scf() - shift: 0
DEBUG::Console1: ****value from comp-type 0
DEBUG::Console1: Console1::map_mute ...
DEBUG::Console1: Console1::map_mute stop blinking
DEBUG::Console1: map_select())
DEBUG::Console1: map_phase 
DEBUG::Console1: map_recenable()
DEBUG::Console1: map_solo()
DEBUG::Console1: map_gate_scf() - shift: 0
DEBUG::Console1: ****value from comp-type 0
DEBUG::Console1: Console1::map_mute ...
DEBUG::Console1: Console1::map_mute stop blinking
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
DEBUG::MIDISurface: something happened on  Console1 in
DEBUG::MIDISurface: data available on Console1 in
```

dehnhardt:
> Moving a Fader on Mixbus, updates the console1, but moving a knob on console 1surface does not update Mixbus. -- This apparently worked last week.

This is even more confusing - I don't see that I have changed anything in retrieveing events from the device.
I'm really sorry for the problems here. 
I fear, reverting the change will be the best for the time being.

x42:
After reverting 94fec7fd8ab59b41c2d22cd819fac520c2416cf8
the user reports "Volume and Pan now works - But none of the other elements work"

There is also no constant stream of "something happened" messages any more, looks like messages are handled again:
image

--

I'll do another build with after reverting 99e2546477f23f4641d85e15054d1e49cfed4ece.
I suppose that might be able to explain things, then again, there are no "

x42:
OK after reverting 99e2546477f23f4641d85e15054d1e49cfed4ece  solo and mute work again.
I don't yet have an explanation..

dehnhardt:
Ok, I'm not sure if your changes were already on my system - even thoug I thought I did a rebase.
Does this mean that for now only pan, volume, solo, mute is working properly or everything? 
If the latter commit was the problem it might be good to reapply my last changes.

x42:
It seems that f34ce17bf0b1e391cc65f4489b30d8b13c6ec0ba prevents any Incoming from being processed.

It seems the issue is caused by calling `MIDISurface::ports_acquire` too early, at a point before BaseUI::run() creates the event loop.

When starting the ctrl surface there is a message
```
GLib-CRITICAL **: 21:51:43.236: g_main_loop_get_context: assertion 'loop != NULL' failed
```
coming from
```#2  0x00007ffff30e468a in g_main_loop_get_context () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007ffff3221119 in Glib::MainLoop::get_context() () at /lib/x86_64-linux-gnu/libglibmm-2.4.so.1
#4  0x00007fffe59296b5 in MIDISurface::ports_acquire() (this=0x61d002135480) at ../libs/ctrl-interface/midi_surface/midi_surface.cc:132
#5  0x00007fffe5927537 in MIDISurface::port_setup() (this=0x61d002135480) at ../libs/ctrl-interface/midi_surface/midi_surface.cc:58
#6  0x00007fffe2b3789d in ArdourSurface::Console1::Console1(ARDOUR::Session&) (this=0x61d002135480, s=...) at ../libs/surfaces/console1/console1.cc:59
#7  0x00007fffe2b349da in new_console1(ARDOUR::Session*) (s=0x6260000d5100) at ../libs/surfaces/console1/console1_interface.cc:33
```

dehnhardt:
Okay, the problem is becoming quite widespread. The error should then occur with many control surfaces, right? I copied the whole port initialization stuff from other implementations.

pauldavisthefirst:
It wasn't copied correctly. 

Other surfaces call (one way or another) BaseUI::run() before doing port setup.

dehnhardt:
@pauldavisthefirst Maybe you can shed some more light on this. I call BaseUI::run() in the 'setActive' method. This is exactly how it is done e.g in faderport8.cc, cc121.cc...
What do I wrong here?

x42:
Faderport8, does not use not use `MIDISurface` as base-class

x42:
I have already cleaned up the code, please pull first before making further changes.

dehnhardt:
Sorry, I did not expect one of tis files being changed outside this PR. I have removed the last commit and rebased. Still, I would like to test it when I'm back home next thursday before reapplying.

libs/surfaces/console1/console1.cc
```diff
@@ -197,6 +204,22 @@ Console1::begin_using_device ()
...
+	blink_connection.disconnect ();
+	periodic_connection.disconnect ();
+	stripable_connections.drop_connections ();
+	session_connections.drop_connections ();
+	console1_connections.drop_connections ();
+	MIDISurface::stop_using_device ();
+	_in_use = false;
+	return 0;
+}
+
```
x42, line 222: Why add this back? instead of using `MIDISurface::stop_using_device`?

libs/surfaces/console1/console1.cc
```diff
@@ -407,19 +430,6 @@ Console1::setup_controls ()
...
-	DEBUG_TRACE (DEBUG::Console1, "stop_using_device()\n");
-
-	blink_connection.disconnect ();
-	periodic_connection.disconnect ();
-	stripable_connections.drop_connections ();
-	console1_connections.drop_connections ();
-	MIDISurface::stop_using_device ();
-	return 0;
-}
-
```
x42, line 422: Why have you remove this?

libs/surfaces/console1/console1.cc
```diff
@@ -512,6 +522,8 @@ Console1::notify_transport_state_changed ()
 void
 Console1::stripable_selection_changed ()
 {
+	if (!_in_use)
+		return;
```
x42, line 526: Why is this required?
It should not be possible to reach here unless the device is in use.
Perhaps just add an `assert(_in_use);`

libs/surfaces/console1/console1.cc
```diff
@@ -175,6 +175,9 @@ Console1::begin_using_device ()
 	  f0 7d 20 00 00 00 01 00 7f 49 6f 6c 73 00 f7
 	*/
 
+	if (_in_use)
+		return 0;
+
```
x42, line 180: Why is this needed? `begin_using_device` cannot be called while the device is already active?
Best compare to `LaunchPadPro::begin_using_device`

x42 (changes requested):

libs/surfaces/console1/wscript
```diff
@@ -22,7 +22,7 @@ def build(bld):
     obj.includes     = [ '.', './console1']
     obj.name         = 'libardour_console1'
     obj.target       = 'ardour_console1'
-    obj.uselib       = 'SIGCPP XML OSX'
+    obj.uselib       = 'GLIBMM GTKMM SIGCPP XML OSX'
```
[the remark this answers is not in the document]
dehnhardt: It doed not compile without: 
In file included from ../libs/surfaces/console1/c1_gui.cc:19:
../libs/surfaces/console1/c1_gui.h:25:10: fatal error: gtkmm/box.h: Datei oder Verzeichnis nicht gefunden
   25 | #include <gtkmm/box.h>
      |          ^~~~~~~~~~~~~

libs/surfaces/console1/console1.cc
dehnhardt: I moved this in my first PR because I found it much clearer to have the two methods start_using_device and stop_using_device directly under each other. 
This makes it easier to keep track of what has to be done when the status of the device changes.
Keep in mind that I don't have the experience and the overview that you do. (Which is pretty obvious right now;-)

libs/surfaces/console1/console1.cc
```diff
@@ -187,6 +190,10 @@ Console1::begin_using_device ()
 	Glib::RefPtr<Glib::TimeoutSource> periodic_timer = Glib::TimeoutSource::create (100);
 	periodic_connection = periodic_timer->connect (sigc::mem_fun (*this, &Console1::periodic));
 	periodic_timer->attach (main_loop ()->get_context ());
+	connect_session_signals ();
+	connect_internal_signals ();
+	create_strip_inventory ();
```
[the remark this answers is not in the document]
dehnhardt: Yes, I've overseen it. Will change it.

libs/surfaces/console1/console1.cc
```diff
@@ -187,6 +190,10 @@ Console1::begin_using_device ()
...
+	_in_use = true;
```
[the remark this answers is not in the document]
dehnhardt: Ok, Then I will remove it.

libs/surfaces/console1/console1.cc
dehnhardt: see above.

libs/surfaces/console1/console1.cc
dehnhardt: This was the only way of not crashing when the connnection was not set or removed. Even after my changes. It might be useless now.

libs/surfaces/console1/console1.cc
dehnhardt: You are right. stop_using_device is called even if there is no connection, start_using_device is only called once. I will delete it.

libs/surfaces/console1/console1.cc
dehnhardt: It still crashes without that check...

libs/surfaces/console1/console1.cc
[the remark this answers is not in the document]
dehnhardt: Removed the ones below.

libs/surfaces/console1/console1.cc
dehnhardt: _in_use = false removed

libs/surfaces/console1/console1.cc
```diff
@@ -187,16 +185,30 @@ Console1::begin_using_device ()
...
-
 	return MIDISurface::begin_using_device ();
 }
 
+int
+Console1::stop_using_device ()
+{
+	DEBUG_TRACE (DEBUG::Console1, "stop_using_device()\n");
+	if (!_in_use)
+		return 0;
```
x42, line 202: This check can go away

libs/surfaces/console1/console1.cc
```diff
@@ -187,16 +185,30 @@ Console1::begin_using_device ()
...
+	blink_connection.disconnect ();
+	periodic_connection.disconnect ();
+	stripable_connections.drop_connections ();
+	session_connections.drop_connections ();
```
x42, line 206: This line can go away
Session connections are unset by `MIDISurface::stop_using_device ();`

x42 (changes requested):

x42:
My `GTKMM` removal from the wscript indeed broke windows builds.
So I've merged this and made the two changes to get them back.

Thanks for your hard work!

dehnhardt:
Many thanks for your ongoing support Robin and Paul!

dehnhardt:
Sorry for commenting here again: Nathan fro Harrison contacted me again and it still seems not to be working on Apple. He tried on the Ardour nightly build 31st of October and had the same problems. From Ardour to Console1 it works, from Console 1 to Ardour it does not work.
I always work with Linux and there it works with: The last Ardour Version, the last GIT version from Ardour, the last Mixbus version and the last GIT version from Mixbus.
I do not have the chance of testing it with an Apple computer but we know that it worked with Mixbus 9.1, becaus Nathan made a video about it...
What can be wrong still?

PR #836, opened 2023-11-03
lppro: Don't unbang trigger until pad is released

DominikM:
Previously, holding a pad on the launchpad pro would trigger the long press method after half a second which would unbang the trigger. This clashed with the Gate launch style for triggering, which resulted in the clip only playing for half a second before stopping regardless of how long the pad is held.

Additionally, there was an extra timer being set for the long press callback.


pauldavisthefirst:
Shouldn't this be conditional on the launch style? As it, it would appear to unbang any trigger on pad release ?

DominikM:
> As it, it would appear to unbang any trigger on pad release ?

That's true, but triggerbox.cc is responsible for choosing what to do with received unbangs. Specifically, at line 899-931, in the case of a one shot trigger, retrigger, and toggle a received unbang does nothing.

pauldavisthefirst:
Manually rebased and merged. Thanks! Welcome to Ardour.

PR #837, opened 2023-11-10
LUA EditorAction script to reverse MIDI notes and regions

ngeiswei:
## Description

LUA EditorAction script to chronologically reverse MIDI notes of selected MIDI regions, alongside MIDI region positions.

* The positions of the MIDI regions are reversed as well, meaning regions at the end appear at the beginning and so on.
* Individual notes are reversed so the ending of a note corresponds to its beginning.  Thus notes should be quantized for this effect to yield good results.
* Note that only MIDI notes are reversed.  Other MIDI events such as CC or SYSEX are left unchanged.

### DIFF
--- a/share/scripts/reverse_midi.lua
+++ b/share/scripts/reverse_midi.lua
@@ -1,11 +1,10 @@
-ardour { ["type"] = "EditorAction", name = "Reverse MIDI events",
+ardour { ["type"] = "EditorAction", name = "Reverse MIDI Events",
 	license     = "MIT",
 	author      = "Nil Geisweiller",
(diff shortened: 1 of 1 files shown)


luzpaz:
Is there a screencast of this in action ?

ngeiswei:
Screencast in action:

https://odysee.com/@ngeiswei:d/Demo-Reverse-MIDI-LUA-Script-for-Ardour-2023-12-09-14-05-53:f

x42:
Brilliant. 
Rebased and merged as 8.1-177 5d73567ce2

PR #839, opened 2023-11-14
Fix error.

gonsolo:


pauldavisthefirst:
no longer needed (see e0d9970056d)

PR #841, opened 2023-11-17
Fix failure to build with libxml2 version 2.12

nphilipp:
This made one argument of xmlStructuredErrorFunc() const, which wasn’t previously so.

### DIFF
--- a/gtk2_ardour/ardour_ui.cc
+++ b/gtk2_ardour/ardour_ui.cc
@@ -255,7 +255,7 @@ libxml_generic_error_func (void* /* parsing_context*/,
 
 static void
 libxml_structured_error_func (void* /* parsing_context*/,
-                              xmlErrorPtr err)
+                              const xmlError *err)
 {
 	string msg;
 
@@ -403,7 +403,11 @@ ARDOUR_UI::ARDOUR_UI (int *argcp, char **argvp[], const char* localedir)
 	/* stop libxml from spewing to stdout/stderr */
 
 	xmlSetGenericErrorFunc (this, libxml_generic_error_func);
-	xmlSetStructuredErrorFunc (this, libxml_structured_error_func);
+
+	/* Cast to xmlStructuredErrorFunc to cope with different constness in different
+	 * versions of libxml2. */
+
+	xmlSetStructuredErrorFunc (this, (xmlStructuredErrorFunc)libxml_structured_error_func);
 
 	/* Set this up early */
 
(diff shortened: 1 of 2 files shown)


x42:
Perhaps a follow up issue to  https://gitlab.gnome.org/GNOME/libxml2/-/issues/526 - though this breaks the API.
It would be unfortunate if we'd need to #ifdef this code depending on libXML version.

x42:
Note that debug versions of Ardour also need further changes. e.g.  libs/pbd/xml++.cc has `xmlErrorPtr xerr = xmlGetLastError (); ` inside a NDEBUG section

nphilipp:
> ```
> ../gtk2_ardour/ardour_ui.cc:406:35: error: invalid conversion from … {aka ‘void (*)(void*, const _xmlError*)’} to … {aka ‘void (*)(void*, _xmlError*)’} [-fpermissive]
> ```

I wonder why it failed to build without the patch then. Maybe I should revert adding the `const` and the call should just cast the function to `xmlStructuredErrorFunc` instead?

> Note that debug versions of Ardour also need further changes. e.g. libs/pbd/xml++.cc has `xmlErrorPtr xerr = xmlGetLastError (); ` inside a NDEBUG section

Mhm, missed that. I guess that here, making it `const xmlError *` should work, shouldn’t it? — I’ll find out, I’m just installing a Debian 11 VM to verify with libxml2 2.9.x 😉 (the Fedora builds cover 2.10 and 2.12).

x42:
Ardour can't be the only project affected by this. Perhaps we should ask libxml devs to provide backwards compatibility?

nphilipp:
> Ardour can't be the only project affected by this. Perhaps we should ask libxml devs to provide backwards compatibility?

They’re kind of in a bind here, declaring that pointer `const` is arguably the right thing – the callback shouldn’t modify what’s passed in – the assumption was probably always there, but not made explicit. Then, libxml2 is a C library and there the difference in constness between the callback and the function signature only yields you a warning, it’s C++ which makes it an error.

I've updated my patch to make that other `xmlError` pointer const (neither gcc nor g++ seem to have a problem with that, regardless of libxml2 version) and make the cast I described above which makes the file compile without warning (on Debian 11 and Fedora 39).

x42:
> They’re kind of in a bind here, declaring that pointer const is arguably the right thing

Yep, the established way for this is to introduce a new function and deprecated the old.

nphilipp:
> > They’re kind of in a bind here, declaring that pointer const is arguably the right thing
> 
> Yep, the established way for this is to introduce a new function and deprecated the old.

I'll leave the can of worms which is that discussion firmly closed. 😅

PR #842, opened 2023-11-20
Fixed a typo

OrakMoya:
Hello,

CS student here trying to get started contributing to open-source software. Honestly don't know where to start, but I spotted this typo while using Ardour so may as well fix it.


x42:
Thanks. We only need the source-code change.  
The .po translation files are automatically updated by `./waf i18n` 

Rebased and merged as 8.1-114 e010cf91b9

PR #843, opened 2023-11-20
Fix GM2 drum kit ProgramChange values

ElectricGeorge:
The GM 2 Percussion Sound Set specifies that PC#1 corresponds with a value of 00h. Before this fix, the Patch Selector UI displays the drum kit names with the wrong program number. This fix makes the names consistent with their PC value.


x42:
Nice find. Thanks!
Rebased and merged as 8.1-176 aa610fbfa5

PR #845, opened 2023-12-13
Fix incorrect variable mistake

ahstro:
Was poking around `mixer_strip.cc`, trying to hide the gain meters and found what I assume was a type/mistake: `hide_button` was used when `width_button` should've been


luzpaz:
Is this still relevant ?

luzpaz:
cc @x42

x42:
Sorry for the delay. Rebased and merged as 8.2-115 2383c9bb37
Thanks for catching this!

PR #847, opened 2023-12-14
Mixer interaction issues

jean-emmanuel:
From PR #844, only fixing a few interaction issues / inconsistencies in the mixer view.

### DIFF
--- a/gtk2_ardour/foldback_strip.cc
+++ b/gtk2_ardour/foldback_strip.cc
@@ -875,6 +875,8 @@ FoldbackStrip::name_button_button_press (GdkEventButton* ev)
 bool
 FoldbackStrip::number_button_press (GdkEventButton* ev)
 {
+	_mixer.select_none();
+	
 	if (ev->type == GDK_2BUTTON_PRESS) {
 		choose_color ();
 		return true;
--- a/gtk2_ardour/foldback_strip.cc
+++ b/gtk2_ardour/foldback_strip.cc
@@ -855,6 +855,8 @@ FoldbackStrip::build_route_select_menu ()
 bool
 FoldbackStrip::name_button_button_press (GdkEventButton* ev)
 {
+	_mixer.select_none();
+
 	if (ev->button == 1 && ev->type == GDK_BUTTON_PRESS) {
 		StripableList            slist;
 		std::shared_ptr<Route> previous = std::shared_ptr<Route> ();
--- a/gtk2_ardour/foldback_strip.cc
+++ b/gtk2_ardour/foldback_strip.cc
@@ -529,7 +529,7 @@ FoldbackStrip::init ()
 	_global_frame.set_name ("MixerStripFrame");
 	add (_global_frame);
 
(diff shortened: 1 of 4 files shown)


gtk2_ardour/vca_master_strip.cc
```diff
@@ -394,10 +394,10 @@ VCAMasterStrip::solo_changed ()
...
 VCAMasterStrip::vertical_button_press (GdkEventButton* ev)
 {
-	if (ev->button == 1 && ev->type == GDK_2BUTTON_PRESS) {
-		start_name_edit ();
-		return false;
-	}
+	// if (ev->button == 1 && ev->type == GDK_2BUTTON_PRESS) {
+	// 	start_name_edit ();
+	// 	return false;
+	// }
```
kiilerix, line 400: *If* this old code should remain as comment, it should have a comment to explain why it is removed yet still around.

(But if there is a comment explaining a non-tricky case, people can look in git history and find the code.)

gtk2_ardour/foldback_strip.cc
```diff
@@ -855,6 +855,8 @@ FoldbackStrip::build_route_select_menu ()
 bool
```
kiilerix, line 855: The hash in the commit message is not found. If it is a reference to a commit that hasn't landed yet, it will probably be rebased and the hash will change. Perhaps better to repeat the relevant parts of the commit message.

x42:
In general it is preferable to act on release (no press). RMB context menus are the only exception.

gtk2_ardour/vca_master_strip.cc
```diff
@@ -119,7 +119,9 @@ VCAMasterStrip::VCAMasterStrip (Session* s, std::shared_ptr<VCA> v)
 	vertical_button.set_angle (90);
 	vertical_button.set_layout_font (UIConfiguration::instance().get_NormalBoldFont());
 	vertical_button.signal_button_press_event().connect (sigc::ptr_fun (&no_propagate), false);
+	vertical_button.signal_button_press_event().connect (sigc::mem_fun (*this, &VCAMasterStrip::vertical_button_press), false);
 	vertical_button.signal_button_release_event().connect (sigc::mem_fun (*this, &VCAMasterStrip::vertical_button_release), false);
+
```
kiilerix, line 124: (this doesn't seem relevant)

gtk2_ardour/foldback_strip.cc
jean-emmanuel: It should be 8c07b1726acb65dbafa329c75eedb5a4d1b1efb8, my mistake.

jean-emmanuel:
> In general it is preferable to act on release (not press). right-mouse-button context menus are the only exception.

@x42 which commit are you referring to exactly ? https://github.com/Ardour/ardour/pull/847/commits/34a0c1ab90c0b96a3110447c920a4fb14e5cf8cc only makes vca label button behave like the rest of the mixer, should it be changed the other way around ?

x42:
Rebased and merged as 8.2-137 d79a852207

jean-emmanuel:
Awesome !! Thank you for fixing the commit reference, I wanted to do it eventually but got busy with other things.

PR #848, opened 2023-12-16
Add more midnam info for Yamaha MX49, MX61, MX88 synths.

ifIMust:
## Summary
Extend the existing Yamaha MX 49/61/88 midnam description with even more information.
Add a control name list, so that automation data will be identified correctly for these models.
Add the missing patch banks and slots, including names for the 128 factory Performance presets, 128 user voices, 8 user drumkits.

## Changes
- Add Performance bank, with patch names for the 128 Performance factory presets. Since user Performances share the same slots, inevitably some names will be wrong; so, the Performance number is also included (e.g. `[088] DownWiDat`), which matches the PERFORMANCE NUMBER display on the MX panel.

- Add/Extend User Voice bank, with generic patch names for all 128 slots (e.g. `User Voice 96`). Even with generic names, the bank is useful: the MX's UI doesn't offer browsing the User bank as its own category, but Ardour's Patch Selector does.

- Add User Drum bank with generic patch names for the 8 User drumkits (e.g. `User Kit 8`)

- Add Control names and information based on product documentation and some discretion. For example, CC 31 (`Amplitude EG Sustain`) is called "Sustain" to match the corresponding knob; "Sustain Pedal" was chosen for CC 64 to differentiate.

- Changed a few names arbitrarily to be more consistent with the literature/presets.

## Validation
Tested with Ardour 8.2.0 Linux and Yamaha MX88.
A MIDI track was created and set to this midnam info. Recorded MIDI from MX88, moved controls around, changed patches from MX panel and from Ardour.
Confirmed that setting a patch from Patch Selector results in the right one being selected on the MX. (Normal Voices, Drum Kits, Performances, boundaries)
Confirmed that patch changes played from MIDI region are handled by the MX as expected.
Confirmed that automation lanes are labeled correctly when recording control data.

Learned that Performance changes at the MX panel do generate Program Change messages, but Voice/Part changes *do not*. This was unexpected and seems to defy configuration.
Confirmed that Performance changes at the MX panel are recorded as patch changes with correct names.

## Remarks
The file is pretty-printed through xmllint. However, I was unsuccessful at validating it, or anything else, against the DTD.

### DIFF
--- a/share/patchfiles/Yamaha_MX-49-61-88.midnam
+++ b/share/patchfiles/Yamaha_MX-49-61-88.midnam
@@ -44,12 +44,27 @@
         <AvailableChannel Channel="15" Available="true"/>
         <AvailableChannel Channel="16" Available="true"/>
       </AvailableForChannels>
+      <UsesControlNameList Name="Controls"/>
+      <PatchBank Name="Performance">
+        <MIDICommands>
+          <ControlChange Control="0" Value="63"/>
+          <ControlChange Control="32" Value="80"/>
+        </MIDICommands>
+        <UsesPatchNameList Name="Performance"/>
+      </PatchBank>
       <PatchBank Name="GM">
         <MIDICommands>
           <ControlChange Control="0" Value="0"/>
           <ControlChange Control="32" Value="0"/>
         </MIDICommands>
         <UsesPatchNameList Name="GM"/>
+      </PatchBank>
+      <PatchBank Name="GM Drum">
+        <MIDICommands>
+          <ControlChange Control="0" Value="127"/>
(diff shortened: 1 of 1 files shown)


ifIMust:
The current version of this file appears to be based  the Yamaha MOTIF XS data for the voices, which is generally the same as the MX. But, the MX has reduced ROM and shortened names for many things.

I want to rewrite the file to align the patch names with the exact (abbreviated) voice names in the MX. I might also clean up/rewrite the more verbose MOTIF XS versions for consistency, and consider including PatchNameLists of both types, allowing the selection of either naming theme.
Closing the PR to do further work.

PR #851, opened 2023-12-21
Build/use libaaf only with dummy backend

nphilipp:
The library is only used by new-aaf-session which requires the dummy backend.


x42:
This will change in the next two weeks before the next release. We first wanted feedback from power users before offering a GUI to import AAF

x42:
Since 8.2-147 1117f8d6cc the dummy backend is no longer optional. It is required for AAF import various session tools etc.

PR #852, opened 2023-12-22
allow to update libaaf from a local directory

agfline:


x42:
PS I think the better option is to not allow to disable the Dummy backend

x42:
Doh. Above comment was meant for #851 

This commit was merged already as 156563a36fc21e018e6cb009a1eec4a3df751e96

PR #854, opened 2023-12-22
Update libaaf to v0.5

agfline:


x42:
Thanks. I've split this into 3 commits (lib and CLI).
I also re-ran clang-format with settings from https://github.com/Ardour/ardour/blob/master/tools/clang-format

PR #859, opened 2024-01-13
Fix Legatize::operator() by not removing 1 tick

ngeiswei:
Please find the following video as complementary explanation

https://odysee.com/@ngeiswei:d/fix-legatize:5


x42:
Nice detective work.
rebased and merged as 8.2-66 dfc1cd94d
Perfect Days

PR #862, opened 2024-01-20
UI tweaks

jean-emmanuel:
I'm back with a few minor ui tweaks! This time I avoided structural changes for real

- removed track header inner shadow in TAV (less fuzzy) 
- themable outline color for buttons, meters & faders
- properly aligned horizontal separator between track headers in TAV (a separator after the last track is still missing though)
- more visible horizontal separator between track headers in TAV (using darker color variant instead of ligher)
- cleaner processor list layout (1px inner stroke removed, more consistent inline display radius and spacing)
- slightly padded tooltips & consistent outline color

That's pretty much it :)


x42:
Looks good. Thanks.

jean-emmanuel:
Yay thank you !

PR #867, opened 2024-02-17
Adding midnam for Yamaha SY85 Synth Factory Banks

Schmitty2005:
Adding midnam for Yamaha SY85 Synth Factory Banks


prokoudine:
Merged, you can close. Thanks!

PR #872, opened 2024-02-26
Grid/Snap: fix triplets/quintuplets, disable broken septuplets

jean-emmanuel:
This PR addresses the regression described in this [ticket](https://tracker.ardour.org/view.php?id=9316).

- triplets are now drawn and snapped to correctly
- quintuplets are now drawn and snapped to correctly
- septuplets are still broken (see comments in last commit) and should be disabled until libtemporal is able to cope with these subdivisions


x42:
Thanks for tracking this down. Amazing!

I took the liberty to rewrite your last commit (disable broken septuplet option) to use C-style comments (see our styleguide) and used `#if 0` to comment out code, which is preferable to changing each line.

Rebased and merged as 8.4-37 83f5d40280

PR #873, opened 2024-02-26
ui: editor: add missing h-separator after last track header

jean-emmanuel:
For consistency with the editor canvas.


x42:
Nice and simple. Thanks

merged as 8.4-34 81bd75e7ae

PR #874, opened 2024-02-26
mixer: use ardour widgets instead of gtk natives for plugin list

jean-emmanuel:
The native dropdown and the "clear" button felt somewhat out of place, this brings some more unity in that regard.

The second commit is an experiment that might be worth considering: having all plugins available in the list with the search option could be a time saver, don't you think ?

### DIFF
--- a/gtk2_ardour/mixer_ui.cc
+++ b/gtk2_ardour/mixer_ui.cc
@@ -112,6 +112,13 @@ using namespace std;
 using PBD::atoi;
 using PBD::Unwinder;
 
+static const gchar *_plugin_list_mode_strings[] = {
+	N_("Favorite Plugins"),
+	N_("Recent Plugins"),
+	N_("Top-10 Plugins"),
+	0
+};
+
 Mixer_UI* Mixer_UI::_instance = 0;
 
 Mixer_UI*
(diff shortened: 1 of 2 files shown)


x42:
I have not forgotten about this.

Yet I'm not sure if "ALL plugins is a good idea". Some users (notably on windows) have a few thousand, and the sidebar also shows presets (some synths have hundreds).  Keep in mind that GTK::Treeview performance degrades significantly with many entries (> 1k).

If you want to find a specific plugin from all available one the dynamically built context-menu or plugin-selector are a lot more ergonomic.

jean-emmanuel:
I understand, I have a few hundred plugins and selecting the option for the first time already takes a short yet noticeable additional time. If the rest is fine do you want me to drop the last commit ?

x42:
I grabbed the first commit 8.6-86 4e44f44e71  but not the "All Plugins".

I tested the "ALL" plugins and it's not just slow (took 4-5 sec), but also somewhat useless to scroll though and find the right one.

It would be nice to have a "search all plugins" options. 
One idea: by default the list is empty, and only when a user typed 3 chars in search, plugins are listed.
Not great, not terrible. Maybe you have a better idea?

jean-emmanuel:
Hey, that's good news thank you ! I like the idea (indeed the ALL option seemed to only make sense with a few chars typed in the search field), I'll try to fiddle with it 
 as soon as I have some time to play with the code again)

x42:
Can we close this an open a new PR with a search-function, once that gets done?

PR #877, opened 2024-03-05
fix compiler flags in ytk and ydk on OSX (define DISABLE_VISIBILITY)

paseri3739:
Added DISABLE_VISIBILITY to compile patched GTK+ on OSX correctly.

### DIFF
--- a/libs/tk/ydk/wscript
+++ b/libs/tk/ydk/wscript
@@ -188,6 +188,7 @@ def build(bld):
         obj.cflags   += ['-xobjective-c']
         obj.uselib   += ' OSX' # -framework Cocoa -framework CoreFoundation -framework ApplicationServices
         obj.includes += ['ydk/quartz', 'ydk/quartz/gdk', 'ydk/gdk/quartz']
+        obj.defines += ['DISABLE_VISIBILITY']
         obj.export_includes += ['ydk/gdk']
         obj.export_includes += ['ydk/quartz']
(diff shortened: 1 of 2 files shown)


x42:
Why is this patch needed? and why would you want to disable visibility? gdk/gtk internal functions should remain hidden, no?

tkysfjk:
@paseri3739 @x42 
Seems that this relates to [error: aliases are not supported on darwin](https://tracker.ardour.org/view.php?id=9673&nbn=8).
I found alternative solution which uses `CFLAGS/CPPFLAGS` with `--keepflags` waf option instead of modifying codes.

ex.
```
$ PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/opt/homebrew/opt/libarchive/lib/pkgconfig" \
    CPPFLAGS="-DDISABLE_VISIBILITY -I/opt/homebrew/include -I/opt/homebrew/Cellar/libarchive/3.7.7/include" \
    CFLAGS="-DDISABLE_VISIBILITY -I/opt/homebrew/include -I/opt/homebrew/Cellar/libarchive/3.7.7/include" \
    LDFLAGS="-L/opt/homebrew/lib -L/opt/homebrew/Cellar/libarchive/3.7.7/lib" \
    ./waf configure --keepflags
```

pauldavisthefirst:
I would still like some explanation on why you feel this change is required.

sdevisch (approved):

PR #878, opened 2024-03-10
update libaaf to v1.0-1-gdef35bf

agfline:
Finally there ;)
This Ardour version fixes some issue while importing audio files and I've started to implement the logging callback function.
I left 2 commented code blocks which I used to route libaaf logs to an external file, which can be very handy for debugging. But if that's really not an option for you, you can remove it.


x42:
Thanks!

I'll split this up in 3 commits.
In general we avoid changing files in libs and gtk2_ardour and tools in the same commit.

PS. It also seems you're not using `tools/clang-format`

agfline:
>  In general we avoid changing files in libs and gtk2_ardour and tools in the same commit.

I take note of it.

> It also seems you're not using tools/clang-format

Sorry, I forgot... I'll pay more attention next time !

x42:
Applied as 8.4-62 d03cdf86e1

Thanks a lot!

PR #882, opened 2024-03-24
Add a hint about quotes in post-export.

automaciej:
When I saw the post-export hint, I thought I needed to quote the arguments. It took me a while to debug what was going wrong: I used quotes in my post-export command, and Ardour was doing a similar thing as well.

Usually, either you provide a string which will be interpolated by the shell, in which case you need to add quotes, or you build an array of strings, and in this case you don't need to add quotes - but you build an array and not write a command template. Ardour's approach is a departure from this mode, so let's save future people time trying to figure this detail out.


x42:
Makes sense. Thanks.

Rebased and merged as 8.6-85 18949a8730

PR #883, opened 2024-03-25
Fix undefined symbol when built with VST3 disabled.

0EVSG:
Commit cd5369c added some cleanup code to `libs/ardour/session.cc`, which drags in the type info symbol of VST3PluginInfo. When built with `--no-vst3`, the resulting binary dies at startup:

```
ld-elf.so.1: /usr/local/lib/ardour8/libardour.so.3: Undefined symbol
"_ZTIN6ARDOUR14VST3PluginInfoE"
```

Make that code compile conditional on VST3_SUPPORT. Occurs on FreeBSD 14.0-RELEASE, clang 16.0.6, amd64.

Making the include statement conditional is not strictly necessary to fix the undefined symbol, but may prevent similar problems in the future.

Original bug report: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=277938

### DIFF
--- a/libs/ardour/session.cc
+++ b/libs/ardour/session.cc
@@ -136,7 +136,10 @@
 #include "ardour/utils.h"
 #include "ardour/vca_manager.h"
 #include "ardour/vca.h"
+
+#ifdef VST3_SUPPORT
 #include "ardour/vst3_plugin.h"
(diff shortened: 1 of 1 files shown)


x42:
Thanks. That was indeed an oversight!

PR #885, opened 2024-04-12
update german translation

edogawa23:


x42:
Rebased and applied as 8.6-1 c9c419213f

Thanks, and sorry for not picking this up before 8.6

PR #888, opened 2024-04-26
Fix for deadlock adding MIDI track when using pipewire jack

hernando:
I know that Ardour does not officially support pipewire and I'm not sure whether this deadlock comes from a non-compliance of pipewire-jack to JACK semantics or pipewire simply uncovered a problematic piece of code in Ardour. I experienced the deadlock with pipewire 0.3. I didn't try a more recent version because it would be complicated to upgrade. In any case, I think it was worth submitting the fix as it may help others.

Without this fix, at the moment a MIDI track is added in an otherwise empty session, the application deadlocks. One thread is waiting in the reader lock for the `_io_lock` mutex in `IO::connected_latency` (io.cc:1275), while another thread acquired the mutex with a writer lock in `IO::ensure_ports` (io.cc:490). The deadlock comes from the function `ensure_ports` being waiting in `_session.engine().register_input_port(...)`, which is calling `jack_port_register` and this one is locked inside `pw_thread_loop_wait`.

On the other hand, there is the thread from `JACKAudioBackend::process_thread` that is calling `jack_cycle_wait` in an infinite loop which in turns calls another pipewire function that polls.

Maybe this is not a deadlock, but a starvation problem because I don't understand why `jack_port_register` doesn't make progress.

The solution consist of moving the writer lock used by `ensure_ports` inside `ensure_ports_locked` and reduce its scope to the minimum. In particular, the mutex is not held while calling `register_input_[in|out]put_port`.

### DIFF
--- a/libs/ardour/io.cc
+++ b/libs/ardour/io.cc
@@ -384,31 +384,34 @@ IO::ensure_ports_locked (ChanCount count, bool clear, bool& changed)
 	assert (!AudioEngine::instance()->process_lock().trylock());
 #endif
 
-	std::shared_ptr<Port> port;
-	vector<std::shared_ptr<Port> > deleted_ports;
-
 	changed    = false;
 
-	for (DataType::iterator t = DataType::begin(); t != DataType::end(); ++t) {
+	{
+		Glib::Threads::RWLock::WriterLock wl (_io_lock);
+		vector<std::shared_ptr<Port> > deleted_ports;
 
-		const size_t n = count.get(*t);
+		for (DataType::iterator t = DataType::begin(); t != DataType::end(); ++t) {
+			const size_t n = count.get(*t);
 
-		/* remove unused ports */
-		for (size_t i = n_ports().get(*t); i > n; --i) {
-			port = _ports.port(*t, i-1);
+			/* remove unused ports */
+			{
+				for (size_t i = n_ports().get(*t); i > n; --i) {
+					auto port = _ports.port(*t, i-1);
 
-			assert(port);
-			_ports.remove(port);
+					assert(port);
+					_ports.remove(port);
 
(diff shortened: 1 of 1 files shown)


pauldavisthefirst:
The problem is that pipewire adopted the INSANE semantics of JACK2 here. 

JACK1 serializes *all* server->client callbacks, because parallel handling of most of them is incredibly hard to do correctly, and why require every client to do its own mutex-based serialization when the server can just take care of it.

JACK2 stupidly IMO (and I feel more and more strongly about this over time) parallelizes server->client callbacks by using multiple threads in the client. Pipewire has inherited/opted for this design, and it is crazy.

To be honest, I'm really tired of continuing to find problems caused by this. The JACK1 design is simple, doesn't force complex design on clients. The downside of course is that JACK2 does what it does so that process() callbacks can continue to be delivered even when other callbacks are happening (for "click-free" connections), which a lot of people like or even regard as important.

Telling clients that they must be able to handle server callbacks at any time in any order and in parallel with process() is just nuts and forces hard-to-prove-correct semantics on the clients.

x42:
possibly related to https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3183

x42:
> I experienced the deadlock with pipewire 0.3.

really update to pipewire 1.0 -- this issue was fixed by pipewire already - which now serializes events.

x42:
It is also rather impressive that you managed to dive deep enough to find and fix (!) threading issues in Ardour's codebase.

hernando:
> It is also rather impressive that you managed to dive deep enough to find and fix (!) threading issues in Ardour's codebase.

Thanks, I have experience on this. For deadlocks, gdb's `thread apply all bt` and reading the code is all you need to find the problem most of the time. Fixing it without breaking something else is another story.

PR #891, opened 2024-05-03
set explicit StartupWMClass for better Linux desktop UX

cbix:
At least on GNOME (tested Wayland + X11) this fixes classifying open Ardour windows and consequently assigning the correct icon etc.


x42:
Rebased and merged as 8.6-83 27154d9769

Thank you!

PR #892, opened 2024-05-04
Fix typos

mattias-ohlsson:


x42:
Applied as 8.6-104 10109c3147

Thank you!

PR #895, opened 2024-05-10
AAF: set session name from AAF filename instead of AAF composition name

agfline:
https://github.com/agfline/LibAAF/issues/5#issuecomment-1952193515


x42:
Rebased and merged as 8.6-139 fc1377ae46
Thanks!

PR #896, opened 2024-06-14
Ignore connection optional ports in LV2 plugins

danielappelt:
This PR allows Ardour to load LV2 plugins that have CV ports. Until now, these plugins were completely ignored. However, there are plugins that work perfectly without using their CV input/output ports, like [Cardinal's main variant](https://github.com/DISTRHO/Cardinal#main).
This PR makes Ardour aware of existing LV2 CV ports without using them.

### DIFF
--- a/libs/ardour/ardour/lv2_plugin.h
+++ b/libs/ardour/ardour/lv2_plugin.h
@@ -256,7 +256,8 @@ class LIBARDOUR_API LV2Plugin : public ARDOUR::Plugin, public ARDOUR::Workee
 		PORT_AUTOCTRL = 1 << 9,  ///< Event port supports auto:AutomationControl
 		PORT_CTRLED   = 1 << 10, ///< Port prop auto:AutomationControlled (can be self controlled)
 		PORT_CTRLER   = 1 << 11, ///< Port prop auto:AutomationController (can be self set)
-		PORT_NOAUTO   = 1 << 12  ///< Port don't allow to automate
+		PORT_NOAUTO   = 1 << 12, ///< Port don't allow to automate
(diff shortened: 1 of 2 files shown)


x42:
Oh bummer. This is one of the issues where discussion before starting to work would have been prudent.

> Ignoring it is not an option. The plugin expects a buffer to be there. Also in the future we will support CV, and then old sessions would break if plugins now no longer have scratch buffers.

from https://discourse.ardour.org/t/mdaepiano-port-0-has-now-known-data-type/110310/7

danielappelt:
Yes. Sorry. I just gave it a shot. The buffer should not be necessary for "connection optional" ports, right? Would this PR be acceptable when adding checks for this?

x42:
Well, the goal is to rather implement support for CV ports (also for VST3).

Also note that ports that may be optionally connected must be set to NULL: https://lv2plug.in/ns/lv2core#connectionOptional when not connected (search for `lilv_instance_connect_port`).  At least you can help yourself for the time being.

Then the question remains.. what happens to sessions that have plugins with now unconnected CV ports once Ardour supports CV?  We'd have to introduce special cases (which not too hard but just more work and maintenance overhead).

danielappelt:
Thanks again for your feedback! For now, I have changed the logic to only check for connection optional ports. I changed the PR to draft as the implementation "To leave a port unconnected, the host MUST explicitly connect the port to NULL." is still missing. I will try to look into this at the end of this week.

x42:
Keep in mind that we should only leave optional CV ports unconnected (not other ports with that property)

danielappelt:
Sorry, it took me some time to get back to this. @x42 can you take a look now?

x42:
Looks good!
Rebased and merged as 8.6-516 af8a7df991. Thanks

PR #897, opened 2024-06-28
osc: Send /select/plugin plugin_id feedback

matthijskooijman:
Whenever the active plugin changes (because the plugin changes or the selected strip changes), the name, some other properties and parameter values for the newly selected plugin is sent as feedback.  However, there was no clear indication of which plugin was selected exactly.

Applications or surfaces that need to know the active plugin could track the currently selected plugin by remembering the most recently selected index, but unexpected plugin changes (e.g. when switching to a strip with fewer plugins) and the asynchronous nature of OSC messages make this more complicated than it should be.

By sending the active plugin id as feedback, clients do not have to guess.

When no plugin is selected (e.g. on startup or when no strip is selected) or the currently selected index is not a plugin (but a some other kind of processor), a value of -1 is sent.

This uses the `/select/plugin` message, which is also used to change the selected plugin. There is a slight discrepancy: the control messages accepts a delta on the selected plugin, while the feedback sends an absolute value.

### DIFF
--- a/libs/surfaces/osc/osc.cc
+++ b/libs/surfaces/osc/osc.cc
@@ -2883,26 +2883,6 @@ OSC::_sel_plugin (int id, lo_address addr)
 			return 1;
 		}
 
-		/* find out how many plugins we have */
-		sur->plugins.clear();
-		for (int nplugs = 0; true; ++nplugs) {
-			std::shared_ptr<Processor> proc = r->nth_plugin (nplugs);
-			if (!proc) {
-				break;
-			}
-			if (!r->nth_plugin(nplugs)->display_to_user()) {
-				continue;
-			}
-#ifdef MIXBUS
-			/* need to check for mixbus channel strips (and exclude them) */
-			std::shared_ptr<PluginInsert> pi = std::dynamic_pointer_cast<PluginInsert>(proc);
-			if (pi && pi->is_channelstrip()) {
-				continue;
-			}
-#endif
-			sur->plugins.push_back (nplugs);
-		}
-
 		// limit plugin_id to actual plugins
 		if (sur->plugins.size() < 1) {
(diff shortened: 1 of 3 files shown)


matthijskooijman:
I'm also working on making the documentation on plugin parameters in the manual complete, I'll add a commit (or maybe separate PR) to document this change as well when it is merged.

matthijskooijman:
I realized that there are actually two ways to identify a plugin:

 1. The user-visible number, usually named "piid". This is a 1-based index into the list of visible, non-channelstrip plugins only.
 2. The plugin id. This is a 1-based index into the list of all plugins as passed to nth_plugin

The `/select/plugin/parameter` message uses the first, and changing the currently selected plugin with `/select/plugin` also applies a delta on the first id. But OSCSelectObserver had only access to the second identifier, so the `/select/plugin` feedback introduced by this PR would return the second identifier.

I've added two additional refactoring commits to this PR to fix this and make sure that the right identifier is returned.

I have not actually been able to test this, since these two identifiers are usually the same, except for:

 1. a plugin that is not `PluginInsert`, but `nth_plugin()` only returns `PluginInsert`s, so this can never happen AFAICS.
 2. a plugin with `display_to_user()` returning false, but I am not sure if this is possible or exists in practice (all cases in the source I can see with `display_to_user()` returning false are specific Processors, not plugins.
 3. a plugin with `is_channelstrip() returning true (mixbus channelstrips according to the comments), but I do not have mixbus set up (and no interest in this).

Any suggestions on a way to trigger situation 2. are welcome.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #898, opened 2024-06-30
Fix phantom retriggering when using Novation Launchpad

jp-bennett:
Running Ardour with "-D trigger" revealed that the Novation Launchpad Mini has a habit of sending multiple pad_press events for a single press. These extra phantom presses were getting stuck in the explicit_queue. Rather than a complicated debounce algorithm, this code just pops the event off the queue when it's triggering the same box multiple times.


jp-bennett:
As per IRC, the better solution is to add an if (!_in_use) at midi_surface.cc line 281. Leaving this PR open for now, as a bookmark for the issue.

pauldavisthefirst:
Should be fixed by e175410f54

PR #900, opened 2024-06-30
Two fixes to OSCSelectObserver: apply feedback config and fix send/plugin paging

matthijskooijman:
These are two small fixes to the OSC selected strip code that I ran into while testing for another (to be submitted) PR. Since these are small and unrelated fixes, it seemed good to submit these separately. See commit messages for details.


x42:
Sorry for the delay, rebased and merged as 8.6-502 a9a5787399

matthijskooijman:
No worries, thanks for merging!

PR #901, opened 2024-06-30
osc: Allow enabling parameter value feedback for all plugins

matthijskooijman:
Currently, feedback is only sent about parameter for a single, selected plugin. By setting this new "all_plugins" bit in the `/set_surface` message (or preferences), behaviour is changed to provide feedback for all plugins in the currently selected strip.

If enabled, this bit changes the format of the `/select/plugin/*` messages by adding an extra (first) argument containing the plugin id (piid).  This matches the format that is accepted by the `/select/plugin/parameter` message already (which allows the piid to be specified or omitted).

The main change is in the last commit, the commits leading up to that are small refactorings to simplify reviewing the big change. The first five commits are the same is #897 and #900, so should probably be ignored here and reviewed in those PRs (I'll rebase once those are merged).

I'll also update the documentation, once https://github.com/Ardour/manual/pull/266 is merged.

### DIFF
--- a/libs/surfaces/osc/osc.cc
+++ b/libs/surfaces/osc/osc.cc
@@ -2883,26 +2883,6 @@ OSC::_sel_plugin (int id, lo_address addr)
 			return 1;
 		}
 
-		/* find out how many plugins we have */
-		sur->plugins.clear();
-		for (int nplugs = 0; true; ++nplugs) {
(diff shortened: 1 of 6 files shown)


sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #902, opened 2024-07-03
update german translation

edogawa23:


x42:
rebased and merged as 8.6-308 0f654ebde4
Thanks!

PR #904, opened 2024-07-17
better translation for 2nd screen of new user wizard

edogawa23:


x42:
Rebased and merged as 8.6-333 787f5b2cc8. Danke!

PR #905, opened 2024-07-23
clang-format source preparations

kiilerix:
These changes makes it possible to build after applying clang-format to most source files. And also a bit of cosmetics.

Those who runs clang-format casually might also need these changes at some point.

These changes allow me to work on top of a local full reformat, making sure my diffs are consistent even if the upstream repo isn't.


libs/ardour/vst_plugin.cc
```diff
@@ -193,6 +193,7 @@ VSTPlugin::set_parameter (uint32_t which, float newval, sampleoffset_t when)
 		} else {
 			cerr << "effSetBypass failed rv=" << rv << endl; // XXX DEBUG
```
kiilerix, line 194: Correction: We also need an empty line here, to avoid that the next line aligns with the `// XXX` above.

x42:
Rebased and merged as 8.6-339 74edf63d01

PR #910, opened 2024-08-19
Remove unused libs/pbd/pbd/stl_functors.h

aledomu:


x42:
8.6-500 867eaa0b13 good riddance

PR #913, opened 2024-08-29
Adds basic transport control for Korg microKONTROL

raphaelbastide:


x42:
Nice, at first I thought those boxes look like spam :) but it does indeed clarify the pad usage.

Do you use the device exclusively as control surfaces? I can see a conflict that if you use it for both, then the notes will also be sent to a synth plugin in addition to controlling Ardour.

We'd need to filter out MIDI messages that are used to control Ardour.

Furthermore are the pads always sending MIDI Channel 1 messages?
Other Pads (elg m-audio oxygen) default to channel 10 (drums), but are configurable.

raphaelbastide:
I used the pads for control surface exclusively, that may explain the very basic mapping I did. I understand my file is far from perfect, you are right it needs to be filtered, but the hardware seam difficult to handle too (I barely understand what I do). Anyway, I have a new version of this file I am going to replace soon in case it is in the interest of the community to have that working.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

luzpaz:
@x42 is this still relevant ?

PR #916, opened 2024-08-31
Turn PinMappings class into a type alias

aledomu:


x42:
At some point in time there was the idea to add Lua Bindings for PinMappings, in which case a class is preferable, so I kept it.

x42:
Then again we can always revert it back to a class later. So far there's no reason to directly expose it.

aledomu:
I thought the original intent was only to extend the original class over its shortcomings at its time. If bindings are planned, then a type alias communicates the wrong intent for the future. Closing.

x42:
On 2nd thought, I was thinking to merge  it anyway, and revert it to a class when needed.

aledomu:
There you have it again to cherry-pick

x42:
Rebased and merged as 8.6-499 e48d97ed69
Thanks!

PR #917, opened 2024-09-01
Remove MSVC quirk handling of StackAllocator

aledomu:
By avoiding the use of `boost::align_storage`, and following the notes, this quirk should not be triggered anymore. But I want @johne53 to confirm.

### DIFF
--- a/libs/pbd/pbd/stack_allocator.h
+++ b/libs/pbd/pbd/stack_allocator.h
@@ -19,7 +19,6 @@
 #ifndef PBD_STACK_ALLOCATOR_H
 #define PBD_STACK_ALLOCATOR_H
 
-#include <boost/type_traits/aligned_storage.hpp>
 #include <limits>
 
 #include "pbd/libpbd_visibility.h"
@@ -61,22 +60,22 @@ public:
 	};
 
 	StackAllocator ()
-		: _ptr ((pointer)&_buf)
+		: _ptr ((pointer)_buf)
 	{ }
 
 	StackAllocator (const StackAllocator&)
-		: _ptr ((pointer)&_buf)
+		: _ptr ((pointer)_buf)
 	{ }
 
 	template <typename U, size_t other_capacity>
 	StackAllocator (const StackAllocator<U, other_capacity>&)
-		: _ptr ((pointer)&_buf)
+		: _ptr ((pointer)_buf)
 	{ }
 
 	/* inspired by http://howardhinnant.github.io/stack_alloc.h */
 	pointer allocate (size_type n, void* hint = 0)
 	{
-		if ((pointer)&_buf + stack_capacity >= _ptr + n) {
+		if ((pointer)_buf + stack_capacity >= _ptr + n) {
 			DEBUG_STACK_ALLOC ("Allocate %ld item(s) of size %zu on the stack\n", n, sizeof (T));
 			pointer rv = _ptr;
 			_ptr += n;
@@ -108,12 +107,12 @@ public:
 
 	bool operator== (StackAllocator const& a) const
 	{
-		return &_buf == &a._buf;
+		return _buf == a._buf;
 	}
 
 	bool operator!= (StackAllocator const& a) const
 	{
-		return &_buf != &a._buf;
+		return _buf != a._buf;
 	}
 
 	template <class U>
@@ -147,12 +146,10 @@ private:
 
 	bool pointer_in_buffer (pointer const p)
 	{
-		return ((pointer const)&_buf <= p && p < (pointer const)&_buf + stack_capacity);
+		return ((pointer const)_buf <= p && p < (pointer const)_buf + stack_capacity);
 	}
 
-	typedef typename boost::aligned_storage<sizeof (T) * stack_capacity, 16>::type align_t;
-
-	align_t _buf;
+	alignas(16) value_type _buf[stack_capacity];
 	pointer _ptr;
 };
 
--- a/libs/pbd/pbd/stack_allocator.h
+++ b/libs/pbd/pbd/stack_allocator.h
@@ -72,6 +72,11 @@ public:
 		: _ptr ((pointer)_buf)
 	{ }
 
+	StackAllocator& operator= (const StackAllocator&)
+	{
+		return *this;
+	}
+
 	/* inspired by http://howardhinnant.github.io/stack_alloc.h */
 	pointer allocate (size_type n, void* hint = 0)
 	{
@@ -142,8 +147,6 @@ public:
 #endif
 
 private:
-	StackAllocator& operator= (const StackAllocator&);
-
 	bool pointer_in_buffer (pointer const p)
 	{
 		return ((pointer const)_buf <= p && p < (pointer const)_buf + stack_capacity);
--- a/libs/ardour/ardour/chan_mapping.h
+++ b/libs/ardour/ardour/chan_mapping.h
@@ -104,22 +104,8 @@ public:
 	 */
 	bool     is_subset (const ChanMapping& superset) const;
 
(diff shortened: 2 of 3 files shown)


johne53:
@aledomu - your change to **libs/pbd/pbd/stack_allocator.h** works fine when building libpbd but I see problems later, when trying to compile some other library which #includes that file. For example here's what I see when trying to build libardour:-

Capture-02

The errors relate to the **copy c'tor for StackAllocator** (lines 66 to 68):-

    StackAllocator (const StackAllocator&)
        : _ptr ((pointer)&_buf)
    { }

x42:
Also just realized that std::at () can throw. So we likely have to revert 0cf73d459b35232aa1f082c2

x42:
The memory pool itself should indeed be noncopyable. 
Copying it would allocate memory (no rt safe), and the point here is to avoid that and instead use the stack.

aledomu:
> Also just realized that std::set::at () can throw. So we likely have to revert [0cf73d4](https://github.com/Ardour/ardour/commit/0cf73d459b35232aa1f082c2476b73f2fb31ae92)

Would operator[] be enough? It does not throw any exception.

> The errors relate to the copy c'tor for StackAllocator (lines 66 to 68):-

Weird, I'm not getting any errors in Linux, but that should trigger too if it's indeed faulty.

johne53:
> > The errors relate to the copy c'tor for StackAllocator (lines 66 to 68):-
> 
> Weird, I'm not getting any errors in Linux, but that should trigger too if it's indeed faulty.

No it's not your code that's faulty - it's the code you've removed from **ardour/chan_mapping.h** and **ardour/plugin_insert.h**. The stack-based allocations flagged up a copy c'tor issue back then with MSVC - though at first I got told that the issue would get fixed from VS2015 onwards, which unfortunately, it wasn't... or maybe MSVC realises somehow that copying isn't valid for objects of that type??

So if you revert your changes from those 2 files it should build again here (in which case I could then just modify my comments in the relevant 2 files).

x42:
@johne53 this pull request has not yet been merged. So far the `_MSC_VER < 1900` check is still in place

x42:
> Would operator[] be enough? It does not throw any exception.

No. operator[] is not const and may add items.

johne53:
> @johne53 this pull request has not yet been merged. So far the `_MSC_VER < 1900` check is still in place

Yes, but it won't be if we end up applying the @aledomu patch fully - in other words, the changes to **ardour/chan_mapping.h** and **ardour/plugin_insert.h** aren't going to help anything

x42:
OK, so we have the option to keep the status-quo and leave the PinMapping special case for MSVC,

(and only update libs/pbd/pbd/stack_allocator.h which is not used by MSVC).

johne53:
Another thing worth considering is that if I apply aledomu's entire patch "as is" but then I comment out the copy c'tor for StackAllocator it doesn't then magically start compiling. What happens is that I then see the compiler errors change to say this:-

Capture-03

So it's looking like there's some code somewhere which needs that copy c'tor ?!?

aledomu:
It's needed by the collection itself, yes.

aledomu:
> OK, so we have the option to keep the status-quo and leave the PinMapping special case for MSVC

I'd even prefer to bracket calls to .at() with a try-catch if it's that needed, just to keep the API idiomatic for any C++ developer. Anyways, that does not affect the issues here.

aledomu:
A semantic API is indeed nicer. Feel free to restore it as long as the name is informative (and leave a comment so no one else tries again to use `at()`).

aledomu:
@johne53 I just made the assign operator public, please try again compiling this PR.

libs/pbd/pbd/stack_allocator.h
```diff
@@ -108,12 +112,12 @@ class /*LIBPBD_API*/ StackAllocator
 
 	bool operator== (StackAllocator const& a) const
 	{
-		return &_buf == &a._buf;
+		return _buf == a._buf;
```
x42, line 115: This does not seem right. Rather than copying data, the target will point to existing data.
Memory corruption will ensure when the copy or original is modified

libs/pbd/pbd/stack_allocator.h
```diff
@@ -61,22 +60,27 @@ class /*LIBPBD_API*/ StackAllocator
...
-		: _ptr ((pointer)&_buf)
+		: _ptr ((pointer)_buf)
 	{ }
 
+	StackAllocator& operator= (const StackAllocator&)
+	{
+		return *this;
+	}
+
 	/* inspired by http://howardhinnant.github.io/stack_alloc.h */
```
x42, line 80: This is incorrect. A copy must not share the underlying memory.
When the copy is modified it will corrupt the memory of the original and vice versa

x42:
I'm sorry but the recent changes go entirely in the wrong direction.  The intention here is to have some POD that is local to each stack  and uses as memory pool for a complex data type (here a std:map).

That memory must not be shared with copies of that map.

x42:
They are just here as pull request on github.

You can either get a diff https://github.com/Ardour/ardour/pull/917.diff
Or use the branch from their github clone: https://github.com/aledomu/ardour/tree/remove-msvc-quirk-fix

johne53:
Thanks Robin but this isn't soething I'm familiar with...

I initially tried a git pull from https://github.com/aledomu/ardour/tree/remove-msvc-quirk-fix but it was denied (saying "repository not found") so I then looked at that page in a browser and it seems to suggest that I needed to create a 'pull request'. So I tried to create one and it then told me "There isn't anyting to compare" :-(

Any idea what I'm doing wrong? A normal git pull should work, surely??

x42:
`remove-msvc-quirk-fix` is a branch in the repo.not the repository itself.

I just created a  branch in the Ardour git repo for you, as follows
```
git checkout -b remove-msvc-quirk-fix master
git pull https://github.com/aledomu/ardour.git remove-msvc-quirk-fix
```

You should now be able to directly use "remove-msvc-quirk-fix" branch from git.ardour.org

(and then (`git push -u origin remove-msvc-quirk-fix`)

libs/pbd/pbd/stack_allocator.h
aledomu: If both `_buf`s are in the same address, they have the same identity, and hence their underlying data will be equal. Notice that `_buf` here is not a `boost::aligned_storage` but a simple array. Without indexing, it is already a pointer.

libs/pbd/pbd/stack_allocator.h
aledomu: I should have written it, but I did not intend for this to be the final code. I just want to check if by changing the interface to meet that of a copyable type it compiles at all, and if it does, then discuss what the behaviour will be. It's obvious this will not do the right thing.

johne53:
Thanks Robin.

@aledomu - this has gone back to giving me the same errors from my original post:-

https://github.com/Ardour/ardour/pull/917#issuecomment-2324360305

libs/pbd/pbd/stack_allocator.h
aledomu: Doing some research, I realised that 'alignas' works not only with C-arrays but with `std::array` as well. I will use that to avoid this kind of confusion for the lack of value semantics.

aledomu:
After quite a few attempts with [godbolt.org](godbolt.org), I can confirm the following:

* This does compile for Visual Studio >= 17.4 (2022). but does not link except for the arm64 platform.
* Prior versions do not compile, and all errors seem to be triggered from MSVC's implementation of `std::map`, although sometimes they might show up as an error in other places, like it did with `boost::aligned_storage`.

Closing this PR. I'll update the notes in another PR.

johne53:
Many thanks for giving this a try @aledomu - even if things didn't work out, it was worth trying!

PR #918, opened 2024-09-02
Use portable std::chrono in PBD::get_microseconds

aledomu:


x42:
We cannot do this. std::chrono uses a syscall and is not rt-safe.

x42:
To elaborate:

On Linux gettimeofday and clock_gettime are just function calls that retrieve a number published by the kernel. The complexity is O(1). rt-safe, wait-free. Tech details are in [vdso(7)](https://man7.org/linux/man-pages/man7/vdso.7.html).

`std::chrono` is implementation dependent, and cannot guarantee that. In a recent real world example (guitarix.vst) this proved to be significant and std::chrono was a cause of audio dropouts.

> < brummer10> x42, I've followed your suggestion for ParallelThread and using clock_gettime()
> < brummer10> after running a couple of debugging sessions, I must admit that clock_gettime() is indeed more suitable. While with std::chrono, nearly every second process callback fails into a timeout (note, one from five allowed) with clock get time this reduced to every 10 - 20. circle.
> < brummer10> in turn this reduced the dsp load b some milli-percent
> < x42> brummer10: nice that you were able to confirm the issue with std::chrono vs clock_gettime()

PR #920, opened 2024-09-06
VST3 non optimal calls #9619

bonsembiante:
Follow issue tracking and context in: https://tracker.ardour.org/view.php?id=9619

### DIFF
--- a/libs/pbd/pbd/controllable.h
+++ b/libs/pbd/pbd/controllable.h
@@ -147,7 +147,7 @@ public:
 	static PBD::Signal1<void, std::weak_ptr<PBD::Controllable> > GUIFocusChanged;
 	static PBD::Signal1<void, std::weak_ptr<PBD::Controllable> > ControlTouched;
 
-	PBD::Signal2<void,bool,PBD::Controllable::GroupControlDisposition> Changed;
+	PBD::Signal3<void,bool,PBD::Controllable::GroupControlDisposition,boost::optional<double>> Changed;
 
 	int set_state (const XMLNode&, int version);
 	virtual XMLNode& get_state () const;
--- a/gtk2_ardour/processor_box.cc
+++ b/gtk2_ardour/processor_box.cc
@@ -1019,7 +1019,7 @@ ProcessorEntry::Control::Control (ProcessorEntry& e,std::shared_ptr<AutomationCo
 
 		_button.signal_clicked.connect (sigc::mem_fun (*this, &Control::button_clicked));
 		_button.signal_led_clicked.connect (sigc::mem_fun (*this, &Control::button_clicked_event));
-		c->Changed.connect (_connections, invalidator (*this), boost::bind (&Control::control_changed, this), gui_context ());
+		c->Changed.connect (_connections, invalidator (*this), boost::bind (&Control::control_changed, this, _3), gui_context ());
 		if (c->alist ()) {
 			c->alist()->automation_state_changed.connect (_connections, invalidator (*this), boost::bind (&Control::control_automation_state_changed, this), gui_context());
 			control_automation_state_changed ();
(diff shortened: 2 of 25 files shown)


libs/pbd/pbd/controllable.h
```diff
@@ -147,7 +147,7 @@ class LIBPBD_API Controllable : public PBD::StatefulDestructible, public std::en
 	static PBD::Signal1<void, std::weak_ptr<PBD::Controllable> > GUIFocusChanged;
 	static PBD::Signal1<void, std::weak_ptr<PBD::Controllable> > ControlTouched;
 
-	PBD::Signal2<void,bool,PBD::Controllable::GroupControlDisposition> Changed;
+	PBD::Signal3<void,bool,PBD::Controllable::GroupControlDisposition,boost::optional<double>> Changed;
```
x42, line 150: I am not really a fan of adding the the value to the Changed signal here. That is a huge commitment, particularly since it is unused except for VST3, and  basically a workaround for yabrdge/VST3 only.

Do you think we could instead cache that value on VST3Plugin  or the implementation VST3PI for a quick lookup instead?

bonsembiante: Hey Robin! Thanks for your answer.

Sure, I could try another approach, I thought that this was the best way because it seems logic to me that the value created with the event goes with the signal.

Regarding caching the value, how does it come to your mind?
One way could be using a sort of stack or queue, but how could it be mantained? Delete the value after one read? What would happend to the other calls to "get_value" caused by the same event? They will lose the chance to get a cached value.

Another way could be caching the value with a timestamp, but I'm thinking that it could be too complex for something like this.

Or caching it during the performEdit until the next beginEdit, but maybe there is a time when we need genuinely fetch the actual value through "get_value" and I could get a cached old one.

What do you think?

Also, which are the cons for passing the value in the signal?

Regards

bonsembiante:
I did some tests and yes, it works, but every performEdit ends triggering 3 calls to "get_value", so caching the value and "uncaching" it after the next read only save us 1 call to normalizedToPlain.

I'm still thinking that sending the value in the signal is a best approach because it seems logic. But I'm also trying to find another way if you think that is not good to send the value in the signal.

Do you know if it will be good to mantain the value cached during all the edit? I mean, save it in the cache when the "beginEdit" call comes, and delete it at the "endEdit" call. I think that way could cover the rest of the cases but I don't know how could that impact in the plugin side. Also, in my analysis of the issue I detected calls to normalizedToPlain after the "endEdit" call, and in these cases the cache option wouldn't be right, because the "endEdit" arrives, we remove the value from the cache queue, but some Changed signal handler is still being executed and it won't reach a cached value, still creating unnecessary calls to the plugin.

libs/pbd/pbd/controllable.h
x42: > Also, which are the cons for passing the value in the signal?

The purpose of the signal is to just inform others that something has changed. That is semantically different from passing a ValueChange.  Also, when an optional value is passed, all API users (notably control surfaces) will need to check if its boost::none, and provide a fall back.  Then the value may change more rapidly than a given UI can update.

> Regarding caching the value, how does it come to your mind?

The basic idea I had is to cache the normalized value at the point when the `Changed()` signal is emitted (here `VST3PI::ParamValueChanged`).   Then return the last cache value in `VST3PI::get_parameter`.

As for the the implementation my first thought that for every call 
   `_shadow_data[p] = value;`
we can add 
 ` _shadow_data_plain[p] = normalizedParamToPlain (id, _shadow_data[p]);`

--

Then again for other plugins (some have 1000+ controls) this approach may add a bit of load if a given value is never used.

x42: Sorry for not replying sooner. Somehow this dropped below my radar.

So just change the VST3 `OnParameterChange` signal to pass along the value (and cache it)?
That sounds reasonable.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #921, opened 2024-09-07
update german translation

edogawa23:


x42:
Rebased and merged as 8.6-514 4b31a1078a. Thanks!

PR #923, opened 2024-09-15
Fix for Boost 1.8.5

fpesari:
The existing code does not work anymore on Boost 1.8.5 (tested on the latest openSUSE Tumbleweed). Replaced it with what should hopefully be equivalent code.


x42:
Thanks. 
I've slightly adjusted the commit message to explain the actual issue. Merged as 8.7-9 f94bde59d7

PR #924, opened 2024-09-15
Define _GNU_SOURCE in hidapi build

fpesari:
I tried compiling Ardour 8.7 on openSUSE Tumbleweed but the build failed due to problems with resolution of functions like `wcsdup`, `strtok_r`, etc. It seems that even if the correct headers are included, those functions won't be included unless `_GNU_SOURCE` is defined (the explanation is in `wchar.h`). 

I added the definition of `_GNU_SOURCE` to the `wscript` in the case of Linux builds so that the code will be compiled correctly.

### DIFF
--- a/libs/hidapi/wscript
+++ b/libs/hidapi/wscript
@@ -21,6 +21,7 @@ def configure(conf):
             autowaf.check_pkg(conf, 'libudev', uselib_store='UDEV', mandatory=False)
             if conf.is_defined('HAVE_UDEV'):
                 conf.define ('HAVE_HIDAPI', 1)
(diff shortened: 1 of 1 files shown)


x42:
Prefer _POSIX_C_SOURCE (see da520cecf3e 4b8b5acfc4) for compatibility with non-gnu compilers.

x42:
From wcsdup(3)
```
       wcsdup():
           Since glibc 2.10:
               _POSIX_C_SOURCE >= 200809L
           Before glibc 2.10:
               _GNU_SOURCE
```

I've added that define in 8.7-10 9c64e7d141

PR #925, opened 2024-10-07
Minor improvements

kiilerix:


pauldavisthefirst:
[fd1dd69] why put the string in a const char* rather than just directly in the #define ?

kiilerix:
That might be premature optimization. But also, it is an attempt at not telling it to do something stupid. I try to avoid giving the linker the extra work of de-duplicating all the different instances of the same string.

x42:
Rebased and merged as  8.9-23 b7249c8fe7

PR #927, opened 2024-10-11
Update adwaita_dark-ardour.colors

cooltehno:
Hello Paul, Robin! May be these PR(s) are not correct...
 I've found some new item "processor automation line" and it makes pale gray color of the automation line in all custom themes. So I've made the "processor automation line"-item the same like an old "automation line"-item (for all custom themes).


x42:
Rebased and merged as 8.9-7 8e4398bc37
Thanks!

PR #930, opened 2024-10-11
Update captain_light-ardour.colors

cooltehno:
Also here I've changed an old "automation line"-item (get rid of alias with midi meter color for common item).


x42:
Rebased and merged as 8.9-12 c64abef1d0
Thanks you!

PR #937, opened 2024-10-12
Add script that duplicates regions on the grid

danielappelt:
Ardour's regular duplicate function makes a copy of each selected region and positions it immediately after the original. This script duplicates a (single) region and places the copy according to the selected grid (for example a quarter note after the start of the original region). This is useful for single drum hit samples and other shorter audio material.


x42:
Nice. Rebased and merged as 8.9-20 6a2d3fcbec
Thanks!

PR #938, opened 2024-10-14
Minor FP8 fix and debug messages

kiilerix:


x42:
Rebased and merged as 8.9-27 0d4bce8663
Cheers!

PR #945, opened 2024-10-28
Remove Sequence::overlaps and Sequence::overlaps_unlocked

ngeiswei:
While attempting to create a PR of an old fix I did a while ago (probably years ago) in `Sequence::overlaps_unlocked`, I discovered that neither `Sequence::overlaps_unlocked` nor `Sequence::overlaps` were used anywhere.  In other words they are dead code.

The question is: are they dead for good, or are they dead to be resuscitated in the foreseeable future?  I'll let you be the judge of that.  If they are to be resuscitated, then I'll cancel that PR and create a new one with my fix (which covers rare but existing corner cases).

### DIFF
--- a/libs/evoral/Sequence.cc
+++ b/libs/evoral/Sequence.cc
@@ -1170,45 +1170,6 @@ Sequence<Time>::contains_unlocked (const NotePtr& note) const
 }
 
 template<typename Time>
-bool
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
I've done this one manually in a branch (midilen) that I'm working, so it will appear in master when I merge (soon). Thanks for the reminder.

PR #947, opened 2024-11-01
Add text control type for LuaDialog. 

zabooma:
Type "text" is based on scrollable Gtk::TextView. 

Example use:

`            local dialog_options = {
                {type = "text", key = "progression", title = "Progression", default = "Am|Dm|E7"}
            }
`

Update track_organizer lua script to use new control type.


x42:
Thanks you!

I've rebased and merged it as 1155da84dd85 and 9.0-pre0-369 a6e2f856a2

PR #949, opened 2024-11-19
Replace for-iterator loops with for-each loops and algorithm functions under libs/

aledomu:


pauldavisthefirst (commented):
This can be very dangerous to do automatically. Perhaps the sed replacements handle it, but there are container-specific cases where erase invalidates the iterator and the code structure doesn't allow for i = container.erase()

Also, I don't see the point in using a type in the for loop if these changes are made. Why not just for (auto & i : container) { ... } ?

x42:
or better yet  with explicit const `for (auto const& i : container) { ... }`

x42:
for loops are one of the only good uses for `auto`. IMHO

libs/ardour/location.cc
```diff
@@ -664,10 +664,8 @@ Location::get_state () const
 {
 	XMLNode *node = new XMLNode ("Location");
 
-	typedef map<string, string>::const_iterator CI;
-
-	for(CI m = cd_info.begin(); m != cd_info.end(); ++m){
-		node->add_child_nocopy(cd_info_node(m->first, m->second));
+	for (const std::pair<const string, string>& m : cd_info) {
+		node->add_child_nocopy(cd_info_node(m.first, m.second));
```
keryell, line 668: The usual modern way is
```suggestion
	for (auto&& [some_stuff, another_thing] : cd_info ) {
		node->add_child_nocopy(cd_info_node(some_stuff, another_thing));
```
picking the right names obviously.

PR #951, opened 2024-11-26
AAF: fix NULL pointer dereference

agfline:
access to audioEssenceFile when audioEssenceFile is NULL


x42:
Rebased and merged as 9.0-pre0-506 a70a2a2bbb

Thank you!

PR #952, opened 2024-11-26
AAF: improve log messages

agfline:


x42:
Rebased and merged as 9.0-pre0-507 3e4b3fcdd3

Thanks!

PR #953, opened 2024-12-22
fix some typos in the german translation

jkbd:
7 typos less...


x42:
Rebased and merged as 9.0-pre0-605 0054078289

Vielen Dank!

PR #955, opened 2024-12-23
AAF: Update Ardour with libaaf v1.0-22-gfab4651

agfline:
* Fix INT_MAX and INT_MIN macros not declared error
* Fix essence file locating when inside 2 sub directories
* Fix clip length beyond end of file


x42:
Rebased and merged as 9.0-pre0-607 03b036ff06

Thanks for your continued effort!

PR #956, opened 2024-12-26
Make Stripable::Sorter::operator() const

ngeiswei:
This is required by C++17.  I also made, by the way, `Stripable::Sorter::_mixer_order` const, just cause it makes sense.


x42:
Rebased and merged as 9.0-pre0-604 34fe8c46e9
Thanks.

PR #957, opened 2024-12-31
one more typo fix in german translation

edogawa23:
before I forget...


x42:
rebased and merged as 9.0-pre0-623 731d6e4a88
Daanke :)

PR #962, opened 2025-01-14
Search all plugins sidebar

jean-emmanuel:
Following up PR #874 !

### DIFF
--- a/gtk2_ardour/mixer_ui.cc
+++ b/gtk2_ardour/mixer_ui.cc
@@ -125,6 +125,7 @@ static const gchar *_plugin_list_mode_strings[] = {
 	N_("Favorite Plugins"),
 	N_("Recent Plugins"),
 	N_("Top-10 Plugins"),
+	N_("Search All Plugins"),
 	0
 };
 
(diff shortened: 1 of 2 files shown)


x42:
I can already see user searching for "wav" and Ardour grinds to a halt listing 1000+ waves plugins :)
so be it  rebased and merged as 9.0-pre0-728 a8f6fcc2df
Thanks you!

jean-emmanuel:
Hahaha maybe the results should only be displayed when when a certain ceiling is reached with a hint otherwise such as "Too many matches, type more characters" ?

jean-emmanuel:
It occurred to me we might as well show a limited subset of results (20, 50?) regardless of how many chars are typed. I did a quick test here and it felt a little smoother to use, however I'm not sure about how/if to hint for the fact that some results are not shown: adding an ellipsis kind-of row looked fine but I didn't manage to make it undragable, but maybe I'm just overthinking this, what do you think ?

x42:
We already have the same problem in the plugin selector's search.. so maybe just leave it as-is.

PR #946, opened 2024-11-01
Create Behringer CMD LC-1 MIDI map

tonilink:
A midi map for the Behringer CMD LC-1 controller

PR #964, opened 2025-01-31
Fix typos

luzpaz:
Found via codespell


luzpaz:
Noticed typos in #946

x42:
We try to avoid touching files in `gtk2_ardour/*` and `libs/*` in the same commit.
So ideally make this two separate commits in the future

prokoudine:
I will do the splitting manually

prokoudine:
Pushed, you can close now

PR #966, opened 2025-03-10
9897 - fix ticker.cc loop range end wrap

jsundqvist:
https://tracker.ardour.org/view.php?id=9897


x42:
Nice find!

now makes me wonder if `end_sample < start_sample` is actually an issue further up and should not happen in the first place.

jsundqvist:
@x42 here's another one https://github.com/Ardour/ardour/pull/967/files

x42:
So it this pull-request ready to go?

I lent my only hardware device that can be sync'ed to MIDI Clock, so I cannot test myself.

jsundqvist:
No, it's accumulating a tiny bit of lag in the loop and getting out of sync, I'll convert to draft.

Other related issues:
* Notes are cut short at the start of the loop (maybe NoteOffs from the previous lap, unrelated to MIDI clock)
* Playback Start/Continue delayed by one beat (probably `_mclk_out_latency` + `midi_clock_beat_at_or_after`, unrelated to loops) 
* Position not reset when Stopping and Continuing.

Let's see if I can make separate changes or if they need to be here. I'm running a Roland TD-27 drum module where MIDI clock is driving the modules built-in Click track, so pretty handy testbed.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #968, opened 2025-03-26
no resolve for MidiPort::realtime_locate at loop end

jsundqvist:
fixes midi notes being cut short at loop start due to `resolve_notes` call after `realtime_relocate` from loop end.

tested and verified working with hardware on 8.12.

### DIFF
--- a/libs/ardour/midi_port.cc
+++ b/libs/ardour/midi_port.cc
@@ -375,9 +375,9 @@ MidiPort::transport_stopped ()
 }
 
 void
-MidiPort::realtime_locate (bool)
+MidiPort::realtime_locate (bool for_loop_end)
 {
-	_resolve_required = true;
+	_resolve_required = !for_loop_end;
 }
 
 void


jsundqvist:
@x42 @pauldavisthefirst

x42:
or in other words: Why should loop-end ***not*** resolve note?

jsundqvist:
It turns out that note resolve at loop end is already handled (correctly) by `MidiNoteTracker::resolve_notes` using `MIDI_CMD_NOTE_OFF` (`0x80`) as opposed to the `MIDI_CTL_ALL_NOTES_OFF` (`0x7B`) sent by `MidiPort::resolve_notes`.

This is what's causing the choking of my TD-27 drum module on the first bar of the loop, and is an incorrect way to resolve notes _unless_ we are also stopping playback, which is probably why the parameter `for_loop_end` exists in the first place.

I verified using Simple Synth that a note extending outside of the loop is resolved at loop end just as before my change.

pauldavisthefirst:
Do you have some citation regarding the use of ALL_NOTES_OFF being "incorrect" ?

pauldavisthefirst:
While some of the points made are fair, I do not accept citations from LLMs as evidence. 

We could potentially add an option to never send All Notes Off (and also All Sound Off).

jsundqvist:
I can't believe I have to wrestle you to get things fixed. Anyway, here's some more LLM jibberish for you:

---

The notion of "unnecessary transmission" for MIDI CC #&NoBreak;123 (All Notes Off) stems from the original intent and limitations of the MIDI specification, as well as practical issues observed by musicians and developers over time. Here's a breakdown of the sources and reasoning behind this bullet point:

MIDI Specification and Intended Use:

* The MIDI 1.0 specification, while defining CC #&NoBreak;123, implies it's a "mode message" meant for channel-wide operational changes rather than continuous control.

* Many MIDI sources (like midi.teragonaudio.com/tech/midispec/ntnoff.htm) explicitly state that "All Notes Off" is for notes turned on by received Note On messages and that it's not supposed to turn off notes played on the local keyboard if a device can't distinguish. This highlights its role as a "panic" or reset function for MIDI input rather than a general-purpose note termination.

* The MIDI specification itself, in its "Channel Mode Messages" section, groups CC #&NoBreak;123 with other reset/mode-altering messages like "All Sound Off" (CC #&NoBreak;120) and "Reset All Controllers" (CC #&NoBreak;121). This categorization suggests they are intended for exceptional situations or system resets, not continuous musical performance.

Historical Implementation and "Stuck Notes":

* In the early days of MIDI, "stuck notes" were a common problem due to dropped Note Off messages, MIDI buffer overflows, or unstable hardware. CC #&NoBreak;123 was a crucial "panic button" to resolve this.

* However, some older keyboards (like the Roland D-50, as discussed in various forum threads on Cantabile Community and Reddit) were designed to send "All Notes Off" automatically when the last key was released. This was a manufacturer-specific behavior, not a general MIDI requirement, and it caused the very side effects mentioned (cutting off releases, interfering with DAWs). This automatic, constant sending is what's considered "unnecessary transmission" in a modern context.

Modern DAW/Sequencer Behavior and Best Practices:

* Modern DAWs and sequencers typically manage Note On/Off messages individually for precise control. They generally don't send CC #&NoBreak;123 constantly unless explicitly instructed to do so or as a "stop" or "panic" function.

* As noted in discussions on forums like Ableton and Squarp, DAWs often send CC #&NoBreak;120 (All Sound Off) or CC #&NoBreak;123 (All Notes Off) when transport is stopped. This is considered acceptable as a reset, but continuous or frequent sending during playback is generally avoided to prevent unintended musical artifacts.

* The Ableton Forum thread "Live 12 - Midi CC 123 not usable" and the Squarp Forum thread "Sending CC higher than 119?" highlight that many DAWs and MIDI controllers treat CCs 120-127 as "reserved for Channel Mode Messages" and often prevent users from mapping them to arbitrary controls. This implies they are not meant for general, continuous data streaming. The squarp forum specifically mentions that "it [CC123] is not really designed to be automated".

* A MuseScore forum discussion ("All notes off overkill?") directly criticizes the practice of sending "All Notes Off" (and even all 128 Note Off messages) on every channel at the start and end of playback, calling it "overkill" and suggesting it can cause "feedback loop circuit breaker" issues in MIDI drivers due to the sudden flood of messages. This reinforces the idea that excessive transmission, even for a "reset" message, can be problematic.

In essence, "unnecessary transmission" refers to situations where CC #&NoBreak;123 is sent more frequently or in contexts beyond its original "panic" or "reset" intent, often by older hardware designs or by _software implementations that don't fully adhere to best practices for modern MIDI sequencing._ **The ideal is to _only_ send it when genuinely needed to clear stuck notes, rather than as a routine part of a MIDI stream.**

jsundqvist:
Or, let's put it like this: I have a hardware device from a well-known (albeit not perfect) manufacturer that is behaving in an unwanted manner because of Ardour.

What "evidence" do you have that the current implementation is correct, and is not causing unwanted side-effects?

x42:
> I favor making the All Notes Off command a preference item, though a complete implementation would do this on a per-track or even per-port basis, depending on what is connected to the port and/or what plugin(s) are in the track.

I was going to suggest that (global pref, which can be overridden at track level), or initially just a per-track option, available from the track header (we already have related *MIDI Patch Restore* option there, to help with some synths). 

What I don't understand is how this change is made on a per-port level, and not in the signal flow itself.
The proposed change would only work around the issue with a specific hardware synth connected a MIDI track's MIDI output.

Surely it would be better handled at the Source, here `MidiRegion::_read_at` when looping (loop range is passed to that function).

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #969, opened 2025-04-03
add fade_in_length() and fade_out_length() methods to audioregion (replace #592)

Houston4444:
Hi.
better late than never, 4 years later I update #592, and because the Ardour code changed a lot, it is simpler to me to post a new PR than update the existing one.

The idea is still exactly the same.

> this PR adds fade_in_length() and fade_out_length() methods to audioregion.
>
> The goal is to give access to theses values from a Lua script.
>
> I also give a script that makes a new playlist in selected tracks with only audible regions. That means it cuts or removes regions if they are below other ones (except during their fade of course). The new playlist is named track_name.MAIN because I attempt to do another script to copy selected regions to the track_name.MAIN playlist.
The idea came making music, with many takes on a guitar track playlist, and always have to move many regions to get a silence at a moment. This script allows to have a very clean playlist easier to edit with exactly the same audible contents.
>
>The script has no undo method because I didn't find any way to remove a playlist.


Houston4444:
> I think this should return a timepos_t (not samples).

It is possible of course.

> Also isn't this information already available by accessing the envelope of the audioregion?!

mmmh, envelope is the gain line of a region, isn't it ? I think the fades are not included in this part.

x42:
> mmmh, envelope is the gain line of a region, isn't it ? I think the fades are not included in this part.

An indeed, we would need to expose `AudioRegion::fade_in()`  and `::fade_out()` for that;
but allowing a user to directly modify them could break some internal assumptions made about those AutomationLists...

so you're right and
 `timepos_t AudioRegion::fade_in_length ()'`  is indeed preferable..

Houston4444:
> `timepos_t AudioRegion::fade_in_length ()` is indeed preferable..

Ok, done.

x42:
Thanks, rebased, squashed and merged as 9.0-pre0-1057 5550654513
It's awesome that you've added an example script too!

Houston4444:
Thanks !

PR #970, opened 2025-04-06
Fix typos in gtk2_ardour

luzpaz:
Not sure but this may include a breaking change, ITC then this PR is to bring awareness to it. Note, this PR hasn't been tested.


luzpaz:
Hi @x42, would you mind reviewing at your convenience ?

luzpaz:
soft bump

prokoudine:
@x42 Unsure about `them[e]ing` — do we refer to this in code somewhere? the rest lgtm.

x42:
These are search terms that a user might enter to find prefs.  So the more the merrier (and a user may also make that typo).

PR #971, opened 2025-04-11
Fix short_version characters's priority

Houston4444:
`short_version` is used in narrow strips to limit the number of letters of strip names or plugin names. Reading its code, something jumped out at me and explain why many times there is nearly no link between the original name and the short name.

This function removes characters until the desired limit is reached.
Originally, It removes characters in this order
* white spaces and punctuation
* lower-case vowels
* upper-case vowels
* lower-case consonants
* upper-case consonants

In this PR, it removes characters in this order:
* white spaces and punctuation
* lower-case vowels
* lower-case consonants
* upper-case vowels
* upper-case consonants

Upper case has priority over lower case ! It explain why
`ACE Reasonable Synth` gave in 5 characters `CRsnS` (no link with original, instead of  `ACRsS`, not perfect but better). It allows to use CamelCase and have at least the first letter of each word.


Houston4444:
Sorry, I made some git mistakes, I start again

x42:
I think we should remove that ancient custom `short_version` function and instead rely on `Pango::EllipsizeMode`
which also works for other languages.

PR #972, opened 2025-04-11
fix short_version (upper/lower) case priority

Houston4444:
Replaces #971.

`short_version` is used in narrow strips to limit the number of letters of strip names or plugin names. Reading its code, something jumped out at me and explain why many times there is nearly no link between the original name and the short name.

This function removes characters until the desired limit is reached.
Originally, It removes characters in this order :
* white spaces and punctuation
* lower-case vowels
* upper-case vowels
* lower-case consonants
* upper-case consonants

In this PR, it removes characters in this order:
* white spaces and punctuation
* lower-case vowels
* lower-case consonants
* upper-case vowels
* upper-case consonants

Upper case has priority over lower case ! It explain why
`ACE Reasonable Synth` gave in 5 characters `CRsnS` (no link with original, instead of  `ACRsS`, not perfect but better). It allows to use CamelCase and have at least the first letter of each word.

Note that "Y" is considered as a consonant, "S" is never removed, I did not change that.

### DIFF
--- a/libs/pbd/convert.cc
+++ b/libs/pbd/convert.cc
@@ -119,19 +119,19 @@ short_version (string orig, string::size_type target_length)
 		orig.replace (pos, 1, "");
 	}
 
-	/* remove upper-case vowels, starting at end */
+	/* remove lower-case consonants, starting at end */
 
 	while (orig.length() > target_length) {
-		if ((pos = orig.find_last_of (_("AEIOU"))) == string::npos) {
+		if ((pos = orig.find_last_of (_("bcdfghjklmnpqrtvwxyz"))) == string::npos) {
 			break;
 		}
 		orig.replace (pos, 1, "");
 	}
 
-	/* remove lower-case consonants, starting at end */
+	/* remove upper-case vowels, starting at end */
 
 	while (orig.length() > target_length) {
-		if ((pos = orig.find_last_of (_("bcdfghjklmnpqrtvwxyz"))) == string::npos) {
+		if ((pos = orig.find_last_of (_("AEIOU"))) == string::npos) {
 			break;
 		}
 		orig.replace (pos, 1, "");


x42:
I think we should remove that ancient custom short_version function and instead rely on Pango::EllipsizeMode
which also works for other languages.

pauldavisthefirst:
My understanding is that Pango ellipsizing is a completely different concept. This one was called "disemvoweling" and comes up with entirely different results. On long text, disemvoweling still makes most languages legible, which ellipszing does not. However, it is possible that on short text (e.g. plugin names), the results are not the same.

Houston4444:
@x42, If we talk about this enum https://docs.gtk.org/Pango/enum.EllipsizeMode.html, I think we are not talking about the same thing. EllipsizeMode only removes the start, the middle or the end of the string. The `short_version` mechanism is pretty good to recognize an item with fewer characters.
@pauldavisthefirst, Note that, for a good  internationalization, it would be nicer to include vowels with accent but there are many (at least 16 in french).

For me, the biggest problem remain the fact narrow strips limits the strip name to 5 characters  (https://tracker.ardour.org/view.php?id=6610). I try since few days a build with this limit set to 10, and I have never seen any truncated name. It probably depends on the font though, but I think this limit could be set to 8 without any risk.

pauldavisthefirst:
The vowel and consonant lists are translatable and can be any size.

pauldavisthefirst:
The idea would be, e.g. that the French translation includes tfrnaslations for the vowel list, and the Hungarian translation also, with its own list. There's no need or reason for a unviersal list.

Houston4444:
something like this ?
```
string lang = std::getenv("LANG");
string sht_lang = lang.substr(0, 2);
string locale_vowels = "";

if (sht_lang == "fr"){
	locale_vowels = "àéèïôùy";
} 
/* others "else if" for other languages */

/* later */
if (locale_vowels.size()){
	while (orig.length() > target_length) {
		if ((pos = orig.find_last_of (_(locale_vowels.c_str()))) == string::npos) {
			break;
		}
		orig.replace (pos, 1, "");
	}
}

pauldavisthefirst:
we localize strings in ardour with gettext, which means the _() macro

hence _("aeiou") is localizable. no other verbiage necessary.

Houston4444:
Ok good. However, I am not able to add libs/pbd to the translations, I am pretty sure I am not so far. Here is the branch where I am trying to make it: https://github.com/Houston4444/ardour/commit/eb17b5c5b1f8ac3b2ed6ae716cf4b6fd3d2eb53c

Translation file is built and installed in `/usr/share/ardour9/locale/fr/LC_MESSAGES/libpbd4.mo`, but it seems to not affect the program.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #974, opened 2025-04-24
Fix potential vulnerable cloned functions: Save stack space while handling errors

npt-1707:
**Description**
This PR fixes a potential vulnerability in luaG_runerror() that was cloned from lua but did not receive the security patch. The original issue was reported and fixed under https://github.com/lua/lua/commit/42d40581dd919fb134c07027ca1ce0844c670daf.
This PR applies the same patch to eliminate the vulnerability.

References
https://nvd.nist.gov/vuln/detail/cve-2022-33099
https://github.com/lua/lua/commit/42d40581dd919fb134c07027ca1ce0844c670daf

### DIFF
--- a/libs/lua/lua-5.3.5/ldebug.c
+++ b/libs/lua/lua-5.3.5/ldebug.c
@@ -657,8 +657,11 @@ l_noret luaG_runerror (lua_State *L, const char *fmt, ...) {
   va_start(argp, fmt);
   msg = luaO_pushvfstring(L, fmt, argp);  /* format message */
   va_end(argp);
-  if (isLua(ci))  /* if Lua function, add source:line information */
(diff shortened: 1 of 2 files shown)


x42:
Thank you!

npt-1707:
Thanks for merging my PR, @x42! 

Just wanted to let you know that I plan to report this as a CVE. Please let me know if you have any concern. Thanks!

pauldavisthefirst:
We would ignore most CVEs filed against Ardour, as would most DAW authors. Why? Because a central part of our function is loading 3rd party code that executes with the full priviledges of the user running Ardour. You don't need a stack overflow or any other sort of security issue to execute arbitrary code in the context of any DAW (except Bitwig, I think). Just create an audio/MIDI plugin and convince a user it is worth using.

PR #976, opened 2025-04-28
Fix typo in ardour.keys.in causing spacebar to not work

edwar4rd:
### Description

This PR addresses a typo introduced in 6f45a9f9ac6028d82828d9c38a03961206e6bef8 causing spacebar to not work at all in dev builds.


pauldavisthefirst:
Thanks for noticing this!

PR #977, opened 2025-05-01
Don’t install test programs

nphilipp:
Installing from a snapshot tarball (`waf dist`) dropped `run-tests` into `/usr/bin`. This should fix it.


x42:
Rebased and merged as 9.0-pre0-1180 3c63a81c1c
Thank you!

PR #982, opened 2025-05-16
Add Nektar Impact GXP MIDI Map

jondbennett:
Just adding the file


jondbennett:
Pretty sure. When I use the MIDI monitor to look at messages from the GXP61, and it is configured to send the transport messages to the DAW MIDI port there does not seem to be a way to change the channel number. It appears to always be 16 no matter the global midi channel setting for the non-DAW port.

prokoudine:
Thanks! Merged, you can close this now

jondbennett:
Thank you so much. A midnam file is to follow.

PR #985, opened 2025-06-26
AAF: fix when AAF file contains no clip

agfline:
If a file contains no clip, sample rate is 0, causing issue with marker positioning in test files.
This patch adds a 48kHz fallback.


x42:
Rebased and merged as 9.0-pre0-1431 d234b2aec4

Thank you for your continued effort!

PR #986, opened 2025-07-01
"MIDI PC/CC" plugin script

BrentBaccala:
Initial version of a Lua script that handles MIDI Program Change (PC) and Continuous Controller (CC) messages.

Put this plugin on a track that can hear the PC and CC messages.  It will activate and deactivate tracks when PC messages are seen, and run arbitrary Lua code when CC messages are seen.

Configuration is done using statements in each track's comment block.

See the plugin description for more documentation.

### DIFF
--- a/gtk2_ardour/lua_proc_handlers.cc
+++ b/gtk2_ardour/lua_proc_handlers.cc
@@ -36,13 +36,23 @@
 
 using namespace ARDOUR;
 
-/* The Lua "prelude" loaded into every handler interpreter.  It supplies the
- * script-facing registry API (register_handler / unregister_handler /
- * clear_handlers) and the single C++-callable entry point __dispatch(id,val).
- * Keeping the registry in Lua keeps the C++ surface minimal (the design
- * report's explicit preference) and gives handlers a plain Lua table to
- * reason about.  pcall isolates a throwing handler so one bad callback can't
- * take down the GUI thread. */
+/* The Lua "prelude" loaded, on the GUI thread, into every handler
+ * interpreter.  It supplies the script-facing registry API (register_handler
+ * / unregister_handler / clear_handlers) and the single C++-callable entry
+ * point __dispatch(id,val).
+ *
+ * The handler interpreter does not exist until the DSP (RT) thread sends its
+ * first message.  That first self:dispatch() emits AccessLuaScript, which is
+ * marshalled to the GUI thread, where dispatch() -> get_or_create() builds the
+ * interpreter, loads this prelude, and runs the script's gui_init() (which
+ * should call register_handler to populate __rt_handlers).  That first
+ * message -- and every message after it -- is then delivered by calling
+ * __dispatch(id, val) in the interpreter, which looks up the handler
+ * registered for id and invokes it.
+ *
+ * Keeping the registry in Lua keeps the C++ surface minimal and gives handlers
+ * a plain Lua table to reason about.  pcall isolates a throwing handler so one
+ * bad callback can't take down the GUI thread. */
 static const char* handler_prelude =
 	"__rt_handlers = {}\n"
 	"function register_handler (id, fn) __rt_handlers[id] = fn end\n"
@@ -122,8 +132,8 @@ LuaProcHandlers::get_or_create (LuaProc* proc)
 		/* registry API + __dispatch entry point */
 		vm->lua->do_command (handler_prelude);
 
-		/* one-time GUI initialisation; this is where the script calls
-		 * register_handler(...) to populate __rt_handlers */
+		/* one-time GUI initialisation; this is where the script is
+		 * expected to call register_handler(...) to populate __rt_handlers */
 		luabridge::LuaRef gi = luabridge::getGlobal (L, "gui_init");
 		if (gi.isFunction ()) {
 			gi ();
--- a/gtk2_ardour/lua_proc_handlers.cc
+++ b/gtk2_ardour/lua_proc_handlers.cc
@@ -1,0 +1,176 @@
+/*
+ * Copyright (C) 2026 Brent Baccala <cosine@freesoft.org>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <iostream>
+
+#include "pbd/error.h"
+
+#include "ardour/luaproc.h"
+
+#include "lua/luastate.h"
+#include "LuaBridge/LuaBridge.h"
+
+#include "gtkmm2ext/gui_thread.h"
+
+#include "public_editor.h"
+#include "luainstance.h"
+#include "lua_proc_handlers.h"
+#include "ui_config.h"
+
+#include "pbd/i18n.h"
+
+using namespace ARDOUR;
+
+/* The Lua "prelude" loaded into every handler interpreter.  It supplies the
+ * script-facing registry API (register_handler / unregister_handler /
+ * clear_handlers) and the single C++-callable entry point __dispatch(id,val).
+ * Keeping the registry in Lua keeps the C++ surface minimal (the design
+ * report's explicit preference) and gives handlers a plain Lua table to
+ * reason about.  pcall isolates a throwing handler so one bad callback can't
+ * take down the GUI thread. */
+static const char* handler_prelude =
+	"__rt_handlers = {}\n"
+	"function register_handler (id, fn) __rt_handlers[id] = fn end\n"
+	"function unregister_handler (id) __rt_handlers[id] = nil end\n"
+	"function clear_handlers () __rt_handlers = {} end\n"
+	"function __dispatch (id, val)\n"
+	"  local fn = __rt_handlers[id]\n"
+	"  if fn then\n"
+	"    local ok, err = pcall (fn, val)\n"
+	"    if not ok then print ('RT-dispatch handler ' .. tostring(id) .. ' error: ' .. tostring(err)) end\n"
+	"  end\n"
+	"end\n";
+
+LuaProcHandlers* LuaProcHandlers::_instance = 0;
+
+LuaProcHandlers*
+LuaProcHandlers::instance ()
+{
+	if (!_instance) {
+		_instance = new LuaProcHandlers ();
+	}
+	return _instance;
+}
+
+LuaProcHandlers::LuaProcHandlers ()
+{
+	/* One process-wide connection.  gui_context() marshals every RT-side
+	 * emit onto the GUI thread via PBD::AbstractUI::call_slot -- exactly
+	 * the path BasicUI::AccessAction uses (editor.cc). */
+	LuaProc::AccessLuaScript.connect (
+			*this, MISSING_INVALIDATOR,
+			std::bind (&LuaProcHandlers::dispatch, this,
+			           std::placeholders::_1,
+			           std::placeholders::_2,
+			           std::placeholders::_3),
+			gui_context ());
+}
+
+LuaProcHandlers::~LuaProcHandlers ()
+{
+	_vms.clear ();
+}
+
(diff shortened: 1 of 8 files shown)


BrentBaccala:
So, this plugin isn't ready to be committed yet, but I'd like to hear any discussion about it in its present form.

The idea is different from automation, which records onto a track.  The idea here is to handle messages coming in over MIDI during live performance.

The simplest way to use it is to have Program Change messages activate and deactivate tracks.  Just put a statement like "MIDI Program 0" in a track's comment block.

Some open issues that I know about:

- it needs to handle Bank Change messages, too
- people might want tracks muted, not deactivated
- it all runs on a realtime thread, so we get lots of messages like "programming error: Session RT event queued from thread without a UI - cleanup in RT thread"
- it can't trigger action events

We need some way to trigger a Lua function from a realtime thread, and queue its execution on a normal thread.

We need some way to trigger action events (buttons) from a Lua script.

Or maybe there's just a better way of doing it, which is why I'm soliciting feedback now, even though it's not done.

x42:
> So, this plugin isn't ready to be committed yet, but I'd like to hear any discussion about it in its present form.

Perhaps mark it as "Draft" then until you it's ready to be merged?

Note that `load_preset` may not be realtime-safe. It's expected to be called from the UI thread (plugin GUI or Ardour GUI).
It *should* be fine for simple plugins where presets just change port values.

Remote control like this should usually happen asynchronously in a dedicated event-loop (see lib/surfaces/) and not from realtime context. I am not conformable to merge this as-is, since others may think it's acceptable to perform this from realtime thread.

BrentBaccala:
@x42, you just taught me about github's "draft" feature.

Any suggestions for how to defer processing off the DSP thread?  Maybe somehow trigger a Lua function to be called from a normal thread, and avoid all of the "programming error: Session RT event queued from thread without a UI - cleanup in RT thread" messages?

BrentBaccala:
I guess the problem here is that we don't have a function to do what we want, which is to defer some action off the realtime thread.

How about adding "Session::dcall(func)" (deferred call), which puts the Lua function "func" onto a queue to be run in a standard thread?

Is there a better namespace to put it in?

@x42, can you suggest where the queue might get emptied and run?

with some guidance like that, i could take a crack at writing it

pauldavisthefirst:
We have extensively used mechanisms for invoking code in another thread. But this all exists at the C++ level, which has a global memory model. Crossing between threads in Lua is massively more complex, because the interpreters in different threads are not the same.

BrentBaccala:
@pauldavisthefirst, is there any way we can create two threads with the same Lua interpreter, one real time and one not?

I know you said they're all in different threads, but is there some way to create two threads with that special property?

pauldavisthefirst:
I don't there's any model for starting threads from Lua in ardour, and in general, you likely need the non-RT thread to be "the right one", rather than just some thread you create.

x42:
> Any suggestions for how to defer processing off the DSP thread?

What you really want to do is not possible with a Lua script and requires C++ code.

As opposed to closed source DAWs we do not need to rely on Lua for control-surfaces, and async code with realtime constraints is really nicer and easier in C++, particularly since Lua interpreter is single threaded. 

That being said.. for a quick/dirty hack what you're doing is OKish (just don't rely on it live on stage :)

BrentBaccala:
The consensus seems to be that a plugin like this needs to be written in C++.  OK, we can do that.

I would still like to provide Lua scripting ability, for flexibility.

We're going to get MIDI messages on a realtime thread.  How (in C++) can a realtime thread trigger Lua code in a standard thread?  Can you guys provide suggestions, or a pointer to some existing code?

pauldavisthefirst:
@x42 is on vacation this week, and he's by far the best person to answer this.

BrentBaccala:
> As opposed to closed source DAWs we do not need to rely on Lua for control-surfaces...

So, how do people control it, especially on stage?  Stream Decks?  Am I being old school for expecting it to respond to Program Change messages?

Paul says you're on vacation this week.  That's good to know; I won't expect an immediate response.

BrentBaccala:
> That being said.. for a quick/dirty hack what you're doing is OKish (just don't rely on it live on stage :)

What I saw yesterday was a track was active (and I could hear it), but it was grayed out in the GUI.  I guess that's the kind of flaky behavior I can expect.

I'm thinking I want the RT thread to trigger something on the GUI's main event loop, just like a mouse click on a script button.

BrentBaccala:
The [documentation](https://docs.gtk.org/glib/func.timeout_add.html) for g_timeout_add() says "It is safe to call this function from any thread".

I'm thinking that the RT thread should call g_timeout_add() with an interval of zero, and the function passed in should always return FALSE, so it will only be called once.

That will move the processing from the RT thread to the GUI thread, and then it should be safe to invoke Lua, right?

pauldavisthefirst:
We do not/have not used the g_timeout_add() mechanism for x-thread invocation (as mentioned, we have our own mechanism that is used all over the place, but not exposed as a Lua API. It might work, it might work but hide issues, it might not work ... unsure.

BrentBaccala:
@pauldavisthefirst, what is the mechanism we use in lieu of g_timeout_add()?

pauldavisthefirst:
We rarely do it directly, but any thread can connect a handler to a PBD::Signal; then when the signal is emitted, the thread in which that happens has a lock-free FIFO queue that connects it to the thread in which the handler is supposed to run, and queues the handler there. That is tied into the glib event loop via an additional file descriptor that wakes up the GUI thread (in the case we're talking about), and then the handler is invoked in that thread.

See libs/pbd/crossthread*.cc for details, which implement the CrossThreadChannel object.

pauldavisthefirst:
Note that this mechanism requires a thread that wants code to be called in its own context must register with the thread that initiates the event that would cause the code to be run. This allows us to do this is a lock-free manner, ignoring kernel side locking on the underlying communication mechanism. 

This is fairly important when its an RT thread that needs to start the ball rolling - it cannot take locks as part of that process.

x42:
For the case at hand, I'd probably use [Mixer Scenes](https://manual.ardour.org/mixing/mixer-scenes/).
and then recall a given scene using custom [MIDI bindings](https://manual.ardour.org/using-control-surfaces/midi-binding-maps/) for (Preferences > Control Surfaces)  ***Generic MIDI*** surface.

e.g.  `~/.config/ardour8/midi_maps/pgm_scene.map`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<ArdourMIDIBindings version="1.0.0" name="MIDI Patch to Mixer Scene">
  <Binding channel="1" pgm="1" action="Mixer/recall-mixer-scene-1"/>
  <Binding channel="1" pgm="2" action="Mixer/recall-mixer-scene-2"/>
  <Binding channel="1" pgm="3" action="Mixer/recall-mixer-scene-3"/>
</ArdourMIDIBindings>
 ```

**Comment by BrentBaccala** [comment]
@x42, the MIDI Bindings seem close to what I want (I didn't know about that feature).  A Program Change or CC message can invoke an arbitrary GUI action, which could be a shortcut to a Lua action script, but there's no way to pass in the program number or CC value.  We can also trigger functions.  Perhaps allow user-defined Lua functions here that receive the program number or CC value?  If the existing functions can do things like "roll_transport", they're probably on a thread that can do things like adjust the active state of tracks.

What do you think?

**Comment by x42** [comment]
> but there's no way to pass in the program number or CC value

Correct. Those are **_Actions_** which operate on the selection, and hence cannot take any arguments.

**Comment by BrentBaccala** [comment]
I've been working to resolve these issues by creating a new type of XML MIDI binding that runs Lua code.

My current implementation uses a new PBD::Signal with a string argument that is Lua code.  It's called BasicUI::AccessLuaScript and it gets connected and fired like the existing BasicUI::AccessAction:

`BasicUI::AccessLuaScript.connect (*this, invalidator (*this), std::bind (&Editor::access_lua_script, this, _1, _2), gui_context());`

Editor::access_lua_script() starts like this:

```
Editor::access_lua_script (const std::string& script, int midi_value)
{
        if (!_session) return;

        ENSURE_GUI_THREAD (*this, &Editor::access_lua_script, script, midi_value)

        LuaInstance* lua_instance = LuaInstance::instance();
        if (!lua_instance) {
                error << "LuaInstance not available" << endmsg;
                return;
        }
```

Does this seem reasonable?

**Comment by pauldavisthefirst** [comment]
Personally I would pass a const char const * and a size_t so that you can pass in multiple bytes of MIDI data (i.e. the whole message if appropriate)

**Comment by BrentBaccala** [comment]
@pauldavisthefirst I haven't decided exactly what the API should look like, which is why I haven't pushed an actual commit.  Passing the whole MIDI message makes sense.  How then should it be presented to Lua?  I'm thinking about using a global variable `midiin`, mimicking the DSP code, but only passing a single message and not the entire buffer.  Judging by the comment on that code, Robin wasn't crazy about putting `midiin` in Lua's global namespace, but after ten years I guess we're somewhat stuck with it, so probably do the MIDI binding code the same way.

**Comment by BrentBaccala** [comment]
I'm slowly working towards an RT thread safe version of this script.

I'm now thinking that a plugin really is the way to go, instead of an XML MIDI binding.  The nice thing about a plugin is that it will respond to recorded MIDI messages as well as messages from Jack inputs.

So, for example, my latest push, which enhances the roll_transport() function to take a string to identify a marker, can be used to skip a region on playback, something [previously requested](https://discourse.ardour.org/t/skip-range-in-playback/103353).

You create a clip with a single CC message (for some reason, it also has to have a single note as well), drop it on a track, and configure the MIDI PC/CC plugin like this:

```
MIDI CC 100: roll_transport(value, "skip to")`
```

Then when we hit that point during playback, the playhead jumps to the "skip to" marker.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

BrentBaccala:
**Reworked to address the realtime-safety concern (force-pushed)**

@x42 — finally returning to this with the redesign you asked for back in July. I've **force-pushed** the branch, so its history is replaced: the PR is now a small amount of C++ infrastructure plus a rewritten script, rather than the original script-only commits. (Heads-up that earlier inline review comments will now point at outdated lines.)

**The problem you raised.** The original plugin did everything from `dsp_run()` on the realtime thread — `load()`/`pcall()` of user code, `route:set_active()`, `Session:set_control()`, iterating the route list — none of which is RT-safe.

**The approach.** Rather than patch the one script, this adds a small, reusable **RT→GUI dispatch primitive** for DSP Lua plugins:
- `LuaProc::dispatch(id, value)` — Lua-callable as `self:dispatch()`. It does nothing but emit a new static signal `LuaProc::AccessLuaScript (LuaProc*, int, int)`. This is modelled on `BasicUI::AccessAction`, but kept in libardour (control_protocol depends on libardour, not the reverse). The emit rides the same cross-thread `call_slot` path Ardour already treats as RT-callable for `Session::request_*`, and the `(LuaProc*,int,int)` payload fits the `std::function` small-object buffer, so there's no allocation on the RT side.
- `LuaProcHandlers` (gtk2_ardour) connects that signal once with `gui_context()`, and lazily builds a per-LuaProc "handler" Lua interpreter with the full action-script binding set (`register_classes`/`register_hooks`). The script's `gui_init()` registers handlers via `register_handler(id, fn)`; incoming dispatches run the matching handler **on the GUI thread**, where the unsafe operations above are legal. The interpreter is torn down on `DropReferences`. This new interpreter mirrors the existing inline-display interpreter pattern (`ProcessorEntry::LuaPluginDisplay`).

The design rationale is written up in the header comment of `gtk2_ardour/lua_proc_handlers.h`, and `share/scripts/_example_rt_dispatch.lua` is a ~50-line minimal reference for the pattern.

**The script (`midi_pc_and_cc.lua`).** `dsp_run()` now only classifies incoming MIDI and dispatches a `(handler_id, value)` pair; all comment-block parsing, route activation and user CC code moved verbatim to GUI-thread handlers. PC/CC are consumed and everything else passes through untouched (selective passthrough). As you noted earlier, there's essentially no expectation that a program change be processed in hard real time.

**Testing.** Rebased onto current master and builds in a standard `--debug` configuration. I've run it live in an Ardour 9 session, driving real tracks from a MIDI controller — program-change and CC handling work as expected.

---
*This comment and the accompanying code were written by an AI assistant (Claude) on behalf of Brent Baccala (cosine@freesoft.org), based on analysis of the Ardour source and a live test session.*

BrentBaccala:
This PR, as it stands, is in violation of the Ardour generative AI policy, so I don't expect it to be accepted as-is.  It's reverted to draft status.

I was planning to attach the AI chat logs that generated this plugin in an attempt to satisfy @pauldavisthefirst that we've got enough evidence to claim copyright ownership of this code.

Unfortunately, due to a bug in my backup script, some of the most important of those chat logs have been lost.

I'll redo the entire PR using completely logged AI sessions (but not today).

Once we have a PR backed by a complete set of chat logs, I'll advance the theory that we have adequate basis to claim copyright.

PR #988, opened 2025-07-16
Enable SSL-EQ Q Control

dehnhardt:
I added a config option to allow the 'old' behaviour under mixbus which is for comapatibility reasons enabled by default and the the two q-potis alwyas set the send send 11 and 12 values.
If this option is unchecked, with the harrison eq the behaviour is the same, when the SSL eq is selected, the q-potis change the q-value of the mid bands.

The general handling of the EQ-Mode seems to be a bit strange.
The "rt->eq_mode_control()->Changed.connect(..." does not seem to have any effect, instead changing the eq mode in the gui seems to trigger a reload of the track...
Nervertheless it seems to work that way.


libs/surfaces/console1/console1.cc
```diff
@@ -697,7 +725,19 @@ Console1::set_current_stripable (std::shared_ptr<Stripable> r)
 			_current_stripable->mapped_control (EQ_Enable)->Changed.connect (
 			  stripable_connections, MISSING_INVALIDATOR, std::bind (&Console1::map_eq, this), this);
 		}
-
+#ifdef MUXBUS
```
x42: MUXBUS ? :)

libs/surfaces/console1/console1.cc
```diff
@@ -709,6 +749,18 @@ Console1::set_current_stripable (std::shared_ptr<Stripable> r)
 			}
 		}
 
+#ifdef mixbus
```
x42: Another `#ifdef`  should be `#ifdef MIXBUS`

x42 (commented):
Nice, I wondered why you wanted the Mixbus source :)

Are those `#ifdefs` required? It looks like there are already NULL checks for 
`rt->mapped_control (EQ_BandQ, band_num));` etc.

dehnhardt:
> Nice, I wondered why you wanted the Mixbus source :)
> 
> Are those `#ifdefs` required? It looks like there are already NULL checks for `rt->mapped_control (EQ_BandQ, band_num));` etc.

When I access the "eq_mode_control" directly, it is required as this does not exist in ardour. 
I removed the two unnecessary #ifdefs.

libs/surfaces/console1/console1.cc
dehnhardt: Wonder why my last test in MIXBUS worked....
Obvoiusly I did not test enough.

libs/surfaces/console1/console1.cc
dehnhardt: s.o.

dehnhardt:
I had to refine the mapping (send 11/12 or q factor) as it worked when the eq mode was switched but not when the channel was changed.

x42:
Thanks, it looks fine now. I'll let it mature for a few more days..
I am in the process of consolidating Mixbus and Ardour source-trees this week, and I'll pull this in ASAP mid next week.

dehnhardt:
> Thanks, it looks fine now. I'll let it mature for a few more days.. I am in the process of consolidating Mixbus and Ardour source-trees this week, and I'll pull this in ASAP mid next week.

Happy to hear this.
Thank you! Holger

x42:
Rebased and merged as 9.0-pre0-1427 280e30ade9

I removed the YDK special-case (that option was removed recently), fixed some `Wmisleading-indentation` warnings and updated the code to be more in line with https://ardour.org/styleguide.html 

I do get a `-Wsometimes-uninitialized` warning in `libs/surfaces/console1/c1_operations.cc:1212`
```
variable 'gain' is used uninitialized whenever 'if (!control)' condition is true
```

 What is the correct solution here? set gain =1 or return early?

PR #989, opened 2025-07-22
Suppress the mapping of unititialized controls in setup dialog

dehnhardt:
Nathan found a bug when switching the new setting in the Colsole1 setup dialog. The mapping of values to controls does not work in that state because the controls are not initialized already.
Fixed by suppressing the mapping.


x42:
Thanks for the quick fix!

I'll ram it though .clang-format before merging (we use tabs for indent see 27 at https://ardour.org/styleguide.html)

x42:
Rebased and merged as 9.0-pre0-1428 361d0374dc.
Thanks

dehnhardt:
Ok, sorry. I just did it the other way around. Do you maybe have a .clang-format file somewhere?

x42:
tools/clang-format in the source-tree gets *most* things right (except for i18n `_()`)

PR #991, opened 2025-07-28
Session dialog improvements

kiilerix:
I have been working on addressing some usability issues, mainly introduced by 524073d8. This works much better for my use case, and I think it should work fine for other use cases too.

Please look at the individual commits. The last ones might not meet all use case requirements (yet), but please cherry-pick the first more trivial ones.

### DIFF
--- a/gtk2_ardour/session_dialog.cc
+++ b/gtk2_ardour/session_dialog.cc
@@ -173,7 +173,8 @@ SessionDialog::SessionDialog (DialogTab initial_tab, const std::string& session_
 	open_button = add_button (Stock::OPEN, RESPONSE_ACCEPT);
 	open_button->signal_button_press_event().connect (sigc::mem_fun (*this, &SessionDialog::open_button_pressed), false);
 
-	_disable_plugins.set_label (_("Safe Mode: Disable all Plugins"));
+	_disable_plugins.set_label (_("Safe Mode: _Disable all Plugins"));
+	_disable_plugins.set_use_underline ();
 	_disable_plugins.set_relief (Gtk::RELIEF_NORMAL);
 	_disable_plugins.set_mode (true);
 	_disable_plugins.set_active (ARDOUR::Session::get_disable_all_loaded_plugins());
@@ -707,7 +708,9 @@ SessionDialog::setup_new_session_page ()
 	session_new_vbox.set_spacing (8);
 
 	Label* name_label = manage (new Label);
-	name_label->set_text (_("Session name:"));
+	name_label->set_text (_("_Session name:"));
+	name_label->set_use_underline ();
+	name_label->set_mnemonic_widget (new_name_entry);
 
 	HBox* name_hbox = manage (new HBox);
 	name_hbox->set_spacing (8);
@@ -720,7 +723,9 @@ SessionDialog::setup_new_session_page ()
 
 	//Folder location for the new session
 	Label* new_folder_label = manage (new Label);
-	new_folder_label->set_text (_("Create session folder in:"));
+	new_folder_label->set_text (_("_Create session folder in:"));
+	new_folder_label->set_use_underline ();
+	new_folder_label->set_mnemonic_widget (new_folder_chooser);
 	HBox* folder_box = manage (new HBox);
 	folder_box->set_spacing (8);
 	folder_box->pack_start (*new_folder_label, false, false);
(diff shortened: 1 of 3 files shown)


x42 (changes requested):
This is used in downstream projects

x42 (commented):
I have no issues using keyboard shortcuts with the current layout. it works just fine here without a mouse (I do that many times a day, new sessions, recent sessions..) One can Tab to different widgets just fine.

What issue does this address?

x42:
Since three downstream projects will need to have the current dialog (with different logo), we cannot merge this as-is.
It may require a a couple if `#ifdef`s. :(

x42:
Just for comparison:

New (kiilerix):
image

current Ardour/git:
image

Mixbus:
image

LiveTrax2:
image

VBM:
image

kiilerix:
> I have no issues using keyboard shortcuts with the current layout. it works just fine here without a mouse (I do that many times a day, new sessions, recent sessions..) One can Tab to different widgets just fine.
> 
> What issue does this address?

It is very interesting information if keyboard navigation works for you in this dialog. It doesn't work for me at all. No relevant mnemonics, no tabbing to the buttons for changing page, and no working tab navigation in the file chooser widget. I assumed everybody saw the same, and I worked towards making it work for me instead of complaining.

I'm using Gnome/Wayland/Fedora. What Linux desktop environment are you using? Perhaps I can find a workaround that achieves enough of the same.

Meanwhile, perhaps some of the changes adding new mnemonics can be used in all cases.

x42:
Ah you're right I cannot use the keyboard to navigate to "New" on the side.  That'll need some `set_can_focus` calls and then also subscribe to a signal when the button is activated.

On our ongoing quest to modernize the look/feel we've removed most gtk notebook tabs (also editor sidebar) and the excess borders those added. For interaction ArdourButtons are preferable since they implement multi-touch support.

I've ping'ed you on IRC, it would [have been] nice to discuss the changes before moving forward.

kiilerix:
My first attempt was to try to find a way to add keyboard navigation (mnemonics and tabbing) to the ArdourButtons, but could not find a way to make it work. Is it possible without significant changes to ArdourButton? ArdourButton doesn't seem to be much like the "native" gtk widgets.

FWIW, I do not think these big yelling buttons look like a modernization and I see no UI hints that they actually change pane/page. I also don't understand the use case for multi-touch when changing page in a "notebook". I thus don't see any improvements, only regressions.

x42:
It would be nice to find a working compromise here.
Perhaps we can patch clearlooks to style the tabs and get rid of the borders.

kiilerix:
Agreed, any solution would be great, no matter if would be a compromise or a perfect solution. Also great if my contribution here could be used.

First, let's clarify:

Do we agree on the main problem I'm trying to fix: The lack of tab navigation and mnemonics in the session dialog?

Do you consider it feasible to make tab navigation & mnemonics work with ArdourButton as it is used now? Or should ArdourButton mainly be used on the canvas with the fine tuned UI, while plain gtk widgets might be preferable in plain dialogs? Do you kind of agree on the direction of these changes: Change these ArdourButton to plain "set_show_tabs"?

Then comes the challenge of Ardour upstream not only carrying disabled but working downstream code, but apparently also carrying dead code that is used by hidden-source downstreams, and constraints of fixed UI design where we can't pragmatically analyze the actual use cases. That's pretty much a show-stopper for contributing.

I learned that you have been hurt by GTK and decided to fork it. So now you are stuck with the old buggy GTK, not only for canvas (where I agree on using custom widgets), but also for plain dialogs (where I'm less convinced custom widgets are the best solution). But here we are, and our task now is to try to work around old gtk2 bugs.

Is the main problem with my "gtk notebook tabs" approach that they seem to add another layer of embossed borders? Do I understand correctly that the changes would be more acceptable if it somehow didn't do that?

pauldavisthefirst:
We were not hurt by GTK.

GTK2 has been EOL'ed by numerous Linux distributions, and will surely be EOL'ed by them all in the not too distant future. 

So we vendored it into our own source tree to make it reasonably easy to build Ardour moving forward. We have already added multitouch to it, improved macOS rendering behavior, and there will likely be more to come. 

There was never any chance of us moving to newer versions of GTK, so this has not changed the future at all.

pauldavisthefirst:
Re: mnemonics ... I cannot think of any other dialog where we use mnemonics on buttons. There are almost no keyboard bindings available, and certainly almost no "obvious" ones, so I am not clear what your thinking is around this.

Re: navigation ... tab navigation is strongly desirable.

x42:
Yes, ideally we'd allow tab navigation, while retaining the borderless plain layout with themable ArdourButtons.

`set_can_focus();` would be a first step.

x42:
Could you check if recent git address your issues?

You can tab to sidebar buttons, and activate them with space and/or enter..

sdevisch (approved):

sdevisch (approved):

sdevisch (approved):

sdevisch (approved):
Approved.

PR #992, opened 2025-08-07
unlimited scrolling on pianoroll caused fpe

ferprojo:
Found that when I accidentally scrolled too much on the piano roll window, it triggered a floating point exception. I looked at how the Editor implemented set_samples_per_pixel, and did the same, just returning when spp < 1. I'm new to ardour development so I don't know if this is correct, but it fixed it for me.


x42:
I'd assume the better solution would be   `if (spp < 1) { spp = 1; }`
Then again ideally that case should not happen in the first place, and we just add an  `assert (spp > 0)`;

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #993, opened 2025-08-12
Add missing include

nphilipp:


x42:
Fixed by removing the call to `PBD::stacktrace`.

PR #994, opened 2025-08-17
update german (gtk2_ardour only for now) translation

edogawa23:


x42:
Danke vielmals. 

Rebased and merged as 9.0-pre0-1603 66599dac36

PR #996, opened 2025-08-21
Add pitch option to dialog

chmaha:
Allow for selecting pitch at the same time as temperament.

### DIFF
--- a/share/scripts/scl_to_mts.lua
+++ b/share/scripts/scl_to_mts.lua
@@ -2,8 +2,8 @@ ardour {
 	["type"]    = "EditorAction",
 	name        = "Scala to MIDI Tuning",
 	license     = "MIT",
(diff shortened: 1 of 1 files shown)


x42:
I'm 50/50 on this. Ideally a user could also specify a .kbm file that specifies the MIDI note that the given frequency apples to and a root note for the scale...

chmaha:
Well, I've personally never used a .kbm file. I use scala tuning files exclusively and I bet a fair few others do too. It's what sfizz and sforzando use for their own temperament tuning.

x42:
Thanks you!

Rebased and merged as 9.0-pre0-1767 afc721e7ed
As we discussed offline, I've added a ref-note entry to also make this useful for scales other than 12-TET

PR #997, opened 2025-08-22
Add support to create midi regions from Lua

danielappelt:
This extension to the Lua API allows to create empty MIDI regions from a script. This may be useful to easily create layers of MIDI regions.

### DIFF
--- a/gtk2_ardour/luainstance.cc
+++ b/gtk2_ardour/luainstance.cc
@@ -54,6 +54,7 @@
 #include "region_view.h"
 #include "processor_box.h"
 #include "time_axis_view.h"
+#include "midi_time_axis.h"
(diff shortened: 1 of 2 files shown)


chousemp3:
NICE, I WONDER HOW MIDI 2.0 DEVELOPMENT IS GOING
ANY PROGRESS?

pauldavisthefirst:
There is no work on MIDI 2.0 occuring and none is planned. There are vanishingly few MIDI 2.0 devices and plugins in the world, and for the most part my take on MIDI 2.0 is that it serves an easily documented but actually very rare need.

x42:
I've split this into two commits and and also added the CLASSINFO (for windows).
Rebased and merged as 9.0-pre0-1704 d09a972b73

PR #998, opened 2025-08-23
Add patch file for Roland INTEGRA-7

bweir:
This merge request adds a patch file for the Roland INTEGRA-7 sound module.

The patch file is adapted from the Cakewalk Sonar Instrument Definition files found on the Roland website. I wrote the [ins2midnam.py script](https://codeberg.org/brettweir/ins2midnam.py) to convert from the .INS format Cakewalk uses to the .MIDNAM format adopted by Ardour.

In the course of writing this script, I found the DTD for the MIDNAM format was no longer available on the MIDI Association website. I was unable to find it anywhere, so I had to reverse engineer the MIDNAM format from other patch files.

I'm happy to address any comments or feedback on this pull request. Having this MIDNAM file has already made using Ardour and the INTEGRA together a lot more enjoyable for me, and I hope that others will find this work useful.

### DIFF
--- a/share/patchfiles/Roland_INTEGRA-7.midnam
+++ b/share/patchfiles/Roland_INTEGRA-7.midnam
@@ -1,0 +1,8030 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE MIDINameDocument PUBLIC "-//MIDI Manufacturers Association//DTD MIDINameDocument 1.0//EN" "http://www.midi.org/dtds/MIDINameDocument10.dtd">
+<MIDINameDocument>
+  <Author>Brett Weir (adapted from the Sonar Instrument Definition files provided by Roland)</Author>
+  <MasterDeviceNames>
+    <Manufacturer>Roland</Manufacturer>
+    <Model>INTEGRA-7</Model>
+    <CustomDeviceMode Name="Default">
+      <ChannelNameSetAssignments>
+        <ChannelNameSetAssign Channel="1" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="2" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="3" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="4" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="5" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="6" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="7" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="8" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="9" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="10" NameSet="Presets" />
+        <ChannelNameSetAssign Channel="11" NameSet="Presets" />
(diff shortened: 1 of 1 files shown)


prokoudine:
Do you think Roland's copyrights apply here?

bweir:
That's a fair question, and I'm not sure.

My immediate thoughts:

- The contents of this file are substantially similar to other files already in the repo (with the possible exception of SuperNATURAL patches), so if it is infringing, it isn't more infringing than the existing repository.

- My understanding is that configuration files/APIs qualify for fair use, but I think this is actively being litigated (<https://en.wikipedia.org/wiki/Google_LLC_v._Oracle_America%2C_Inc.>).

- The INTEGRA is not useful at all without instrument definitions and they are distributed to all users as part of the manual for the purpose of programming the INTEGRA. It's hard to imagine Roland wanting to limit access to instrument definitions, but companies have done stranger things.

All of this being said, I am not a lawyer, and I don't know the circumstances by which the existing patch files were added to the Ardour project.

All Roland downloads on the website require you to agree to a EULA that seems to be boilerplate. Here is the instrument definition download:

<https://www.roland.com/global/support/by_product/integra-7/updates_drivers/7164a999-7992-4010-b841-51872a7565e1/>

It could be possible to sidestep the issue by rebuilding the configs using configs already in the repository, but it would be very time-consuming to verify that the mapping is accurate for the INTEGRA hardware.

chousemp3:
> The patch file is adapted from the Cakewalk Sonar Instrument Definition files found on the Roland website. I wrote the [ins2midnam.py script](https://codeberg.org/brettweir/ins2midnam.py) to convert from the .INS format Cakewalk uses to the .MIDNAM format adopted by Ardour.

THATS SMART, I WONDER WHAT KIND OF GOODIES THESE DIGITAL AUDIO SOFTWARE PROGRAMS ARE JAMMED PACKED WITH

x42:
The EULA states

> "Unauthorized copying, uploading to another network, distributing, modifying and/or attaching this software to other media or any other sales item is expressly forbidden."

In your case you modify (convert to MIDNAM) and distribute it.

> It could be possible to sidestep the issue by rebuilding the configs using configs already in the repository

indeed, but if you agreed to the EULA, someone else needs to do that.
https://en.wikipedia.org/wiki/Clean-room_design  - which is how Ardour's VST2 support came to be.

bweir:
It sounds like there isn't a path forward for the merge request in its current form. I will post instructions somewhere for reproducing the INTEGRA-7 patch files from Roland's downloads for those who want to perform the conversion themselves. Then I will ponder the path forward.

Thanks for your review and feedback!

gfontanilla:
> > It could be possible to sidestep the issue by rebuilding the configs using configs already in the repository
> 
> indeed, but if you agreed to the EULA, someone else needs to do that. https://en.wikipedia.org/wiki/Clean-room_design - which is how Ardour's VST2 support came to be.

Would copying relevant parts of this [sound list file](https://static.roland.com/assets/media/pdf/INTEGRA-7_Sound_List_je01_W.pdf) be infringing? The file is searchable directly from Google; there was no EULA to agree to before downloading it. The following is the copyright notice in the document itself:

```
Copyright © 2012 ROLAND CORPORATION
All rights reserved No part of this publication may be reproduced in any form without the written permission of ROLAND CORPORATION
```

@x42 @bweir

PR #999, opened 2025-08-23
Add support to set color transparency from Lua

danielappelt:
This extension to the Lua API allows to set the color transparency (like in Preferences -> Appearance -> Colors -> Transparency) from a script. I use it, for example, to quickly adapt region/automation lane transparency when working with layers of MIDI regions.


x42:
Nice. Thank you!

Rebased and merged as 9.0-pre0-1706 4b5a55546a

PR #1000, opened 2025-08-29
Add MIDI note strumming feature

chousemp3:
- Implement strum_notes functionality to apply progressive timing offsets to selected MIDI notes
- Add strum_notes_forward() and strum_notes_backward() methods to MidiView class
- Register strum-forward and strum-backward actions in EditingContext
- Add 'Transform' submenu to note context menu with Strum options
- Provides guitar-style strumming for creating realistic note sequences

The feature applies 1/32 beat timing offsets progressively to selected notes, creating a strummed effect where notes play in sequence rather than simultaneously.

### DIFF
--- a/gtk2_ardour/editing_context.cc
+++ b/gtk2_ardour/editing_context.cc
@@ -1939,17 +1939,9 @@ EditingContext::popup_note_context_menu (ArdourCanvas::Item* item, GdkEvent* eve
 	if (sel_size < 2) {
 		items.back().set_sensitive (false);
 	}
-
-	Menu* transform_menu = manage (new Menu);
-	MenuList& transform_items = transform_menu->items();
-	transform_menu->set_name ("ArdourContextMenu");
-
-	transform_items.push_back(MenuElem(_("Velocity, etc..."), sigc::bind(sigc::mem_fun(*this, &EditingContext::transform_regions), mvs)));
-	transform_items.push_back (SeparatorElem());
-	transform_items.push_back(MenuElem(_("Strum forward"), sigc::mem_fun(mrv, &MidiView::strum_notes_forward)));
-	transform_items.push_back(MenuElem(_("Strum backward"), sigc::mem_fun(mrv, &MidiView::strum_notes_backward)));
-
-	items.push_back(MenuElem(_("Transform"), *transform_menu));
+	items.push_back(MenuElem(_("Mathematical Operations..."), sigc::bind(sigc::mem_fun(*this, &EditingContext::transform_regions), mvs)));
+	items.push_back(MenuElem(_("Strum Forward"), sigc::mem_fun(mrv, &MidiView::strum_notes_forward)));
+	items.push_back(MenuElem(_("Strum Backward"), sigc::mem_fun(mrv, &MidiView::strum_notes_backward)));
 
 	_note_context_menu.popup (event->button.button, event->button.time);
 }
--- a/libs/ardour/strum.cc
+++ b/libs/ardour/strum.cc
@@ -40,9 +40,11 @@ Strum::operator()(std::shared_ptr<ARDOUR::MidiModel> model,
 	}
 
 	// Get all notes from all sequences and sort them by start time
-	Notes all_notes;
+	std::vector<NotePtr> all_notes;
 	for (std::vector<Notes>::iterator s = seqs.begin(); s != seqs.end(); ++s) {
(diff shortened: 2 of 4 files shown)


x42:
Please strip all the irrelevant parts (.md, .py) from the pull request. those will not be merged. relevant here are only the .cc .and .h files in gtk2_ardour.

Next have a look at `MidiOperator` in libardour. The *view* is really just that a view and controller.
The strum operation should not be in the GUI, but implemented as Operator in libardour.  Compare to "quantize" operation. or Legatize, Transpose, etc.

x42:
Perhaps also check out https://github.com/x42/midifilter.lv2/blob/master/filters/midistrum.c

x42:
Nice. That looks better. 

Quick question: what happen if notes end up overlapping due to strum?

chousemp3:
I'm thinking it'll just do some kind of glitched polyphonic notes
but I will test that out and let you know

chousemp3:
OKAY so I've tested it out, right now the strum operator only handles the StartTime and not the note length, so overlaps are handled by MidiModel overlaps. By default Ardour uses the InsertMergeRelax policy from the session_configuration_vars, so it allows overlaps to exist. All that happens is both notes play or more play at the same time so my theory is right, it could create some weird glitched polyphonic notes or dissonance if you play multiple notes at the same time with midi strum. So my theory was right after testing :+1:

libs/ardour/strum.cc
```diff
@@ -0,0 +1,86 @@
...
+	Command* cmd = new Command(model, name());
+
+	if (seqs.empty()) {
+		return cmd;
+	}
+
+	// Get all notes from all sequences and sort them by start time
+	Notes all_notes;
+	for (std::vector<Notes>::iterator s = seqs.begin(); s != seqs.end(); ++s) {
+		all_notes.insert(all_notes.end(), (*s).begin(), (*s).end());
```
x42: Fails to compile.
```
no matching member function for call to 'insert'
```

 Note that ardour is C++17;  see http://en.cppreference.com/w/cpp/container/multiset/insert.html

gtk2_ardour/editing_context.cc
```diff
@@ -1939,7 +1942,17 @@ EditingContext::popup_note_context_menu (ArdourCanvas::Item* item, GdkEvent* eve
 	if (sel_size < 2) {
 		items.back().set_sensitive (false);
 	}
-	items.push_back(MenuElem(_("Transform..."), sigc::bind(sigc::mem_fun(*this, &EditingContext::transform_regions), mvs)));
+
+	Menu* transform_menu = manage (new Menu);
+	MenuList& transform_items = transform_menu->items();
+	transform_menu->set_name ("ArdourContextMenu");
+
+	transform_items.push_back(MenuElem(_("Velocity, etc..."), sigc::bind(sigc::mem_fun(*this, &EditingContext::transform_regions), mvs)));
```
x42: All other items here "Quantize" "Legatize" etc are also transformation, so they would need to go in that menu as well. 
Hence, I think it is preferable to leave it a flat menu. 

"Velocity, etc..."  does vastly under-represent  https://manual.ardour.org/working-with-midi/transformation-midi/

Perhaps "Mathematical Operations..." ?

chousemp3: Thanks! Yeah I think this is ready from my side. The compilation error should be fixed now - I changed the Notes container to std::vector<NotePtr> with push_back() to be C++17 compliant. And I flattened the menu structure like you suggested, using "Mathematical Operations..." as the label.

Feel free to take it from here and do the rebase/commit organization! Really appreciate all the guidance on the MidiOperator architecture and getting this to follow Ardour's patterns properly 👍

x42:
Thank you!

I've just rebased it and split it into 4 commits 9.0-pre0-1792 9dafed2d76

prokoudine:
@chousemp3 Do you have plans to enhance this feature to make it more on par with @x42's plugin?

x42:
@prokoudine refers to https://x42-plugins.com/x42/x42-midifilter which has options for duration, strum direction, acceleration, etc.

image

chousemp3:
> @chousemp3 Do you have plans to enhance this feature to make it more on par with @x42's plugin?

maybe , i think ardour can and will become better than ableton if I and others decide to team up and get to work, but until then. think of this as just a hobby / community fun stuff, there is no incentive for me to make it better as of now.
Now that can change in the future like i said. its not just up to me. 
that's what i think.

PR #1001, opened 2025-09-08
Add Program Change dropdown to Virtual Keyboard

BrentBaccala:
This was coded by Claude Sonnet 4, running in Windsurf, using the following prompt:

> Ardour has a virtual keyboard with an upper row of controls that starts "channel", then "pitchbend", then "modulation", then 4 continous controllers, then "octave", then a few more controls.  I want you to add a new control called "Program" just to the left of "Octave".  "Program" should be a dropdown (like "Octave"), with 128 choices from 0 to 127.  When one of these is selected, the corresponding MIDI Program Change message should be sent on the current channel.

I tested and reviewed the AI's code.  It looks fine; I accepted it without any changes other than adding copyright messages.

### DIFF
--- a/gtk2_ardour/virtual_keyboard_window.cc
+++ b/gtk2_ardour/virtual_keyboard_window.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2019 Robin Gareus <robin@gareus.org>
+ * Copyright (C) 2025 Brent Baccala <cosine@freesoft.org>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


x42:
I intentionally did not add a patch-change selector to the MIDI keyboard.

Ardour has dedicated bank/patch selector, which not only lists the program by name, but also include bank select.
The bank/patch selector is also updated to reflect the current program.

BrentBaccala:
@x42, I see.  Didn't know about that feature.

I'm using the virtual keyboard to test Ardour's handling of program change messages coming in from a controller.  I don't see any way to route patch selects on a track to Ardour's control surface bindings, which is my new way of doing this, since it doesn't look like a plugin is going to cut it (as we've discussed on another PR).

I'm not sure if sending Program Changes from the virtual keyboard would be useful for anybody but me.

x42:
> I'm using the virtual keyboard to test Ardour's handling of program change messages coming in from a controller.

I see, and yes, short of an actual keyboard that is indeed a good use-case.

sdevisch (approved):

sdevisch (approved):

sdevisch (dismissed):

sdevisch (approved):
Approved.

PR #1003, opened 2025-09-18
update german translation to current

edogawa23:


x42:
Danke vielmals. Rebased and merged as 9.0-pre0-1802 72eda5e98f

PR #1004, opened 2025-09-22
Fix tooltip typo in rta_windows.cc

luzpaz:
Fixes minor user-facing typo


luzpaz:
cc @x42 if you don't mind reviewing when you're available :pray:

x42:
Thanks for cleaning up after me.
Rebased and merged as 9.0-pre0-1823 5c4f95a0f8

PR #1024, opened 2025-12-13
Update Polish translation

songo:
Vastly updated Polish translation. I consider it practically complete.


prokoudine:
Cherry-picked and pushed to master. Thanks for the update! You can close it now, as I don't have the privileges :)

PR #1059, opened 2026-02-22
Generic MIDI Surfaces: fix for erroneous feedback

geraldog:
last_value is updated in MIDIControllable::midi_sense_note without regard to the fact that last_value must be different from feedback value in MIDIControllable::write_feedback.

Update to only touch last_value in MIDIControllable::midi_sense_note if Surface Feedback is not in use.

Tested on AKAI MIDIMix with Normal mapping. Without the change there is erroneous feedback eventually. With the fix the lit up buttons correspond to the actual state of the Stripables.

### DIFF
--- a/libs/surfaces/generic_midi/midicontrollable.cc
+++ b/libs/surfaces/generic_midi/midicontrollable.cc
@@ -321,7 +321,9 @@ MIDIControllable::midi_sense_note (Parser &, EventTwoBytes *msg, bool /*is_on*/)
 		}
 	}
 
-	last_value = (MIDI::byte) (_controllable->get_value() * 127.0); // to prevent feedback fights
+	if (!_surface->get_feedback ()) {
+		last_value = (MIDI::byte) (_controllable->get_value() * 127.0); // to prevent feedback fights
+	}
 }
 
 void


x42:
I am not sure that this is correct.

`feedback` controls the other direction: It sets whether data is to be sent back to the device; it should be irrelevant for any data received from the controller.

`last_value` is relevant for devices that are do do not have *motorized* faders.


* enable-feedback: A message is sent to the device when a mapped value changes
* motorized: A physical knob’s position is assumed to always matche the control value

https://discourse.ardour.org/t/encoder-initial-value-when-a-session-is-loaded/100319/11?u=x42

geraldog:
Hi Robin,

I would have opened an Issue at first but your project does not accept those here on GitHub (no blame).

The Issue at hand is seen with my AKAI MIDIMix (which certainly does not have motorized faders): the state of the lit-up feedback buttons eventually degrades and no longer corresponds to the state of the Ardour Stripables, specially on Bank Change.

From the looks of it, I would risk state there's a race condition between MIDIControllable::midi_sense_note and MIDIControllable::write_feedback regarding the state of last_value.

I don't claim my patch is the ultimate solution, but it does solve the real issue, so it's certainly a start.

PR #1085, opened 2026-03-17
pianoroll: update loop range when switching regions in loop mode

pyrotek45:
this makes it so, when you have the bottom panel open, and you click on a midi region with loop mode on, it sets up the range to that region so you don't have to click the loop mode button over and over to reset it. 

[Screencast_20260316_235546.webm](https://github.com/user-attachments/assets/0631dec4-a073-452e-bf8c-a7da05e0064d)

### DIFF
--- a/gtk2_ardour/pianoroll.cc
+++ b/gtk2_ardour/pianoroll.cc
@@ -54,6 +54,7 @@
 #include "pianoroll_midi_view.h"
 #include "note_base.h"
 #include "prh.h"
+#include "public_editor.h"
 #include "timers.h"
 #include "ui_config.h"
 #include "verbose_cursor.h"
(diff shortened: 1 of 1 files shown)


pyrotek45:
Why re open these? Robin said what i was doing was wasting you guys time. I prefer not to waste your time and apologize for any time I did waste.

pauldavisthefirst:
I'm currently watching over my mother-in-law with the rest of her family as she dies. I want to be able to read over these more carefully and get the maximum value that we can out of what you've provided here. But I can't do that until I'm back home - I don't currently know when that will be.

It's not a waste of time, but it's also not straightforward either.

x42:
If at all, this should be a preference or dedicated mode:  "loop follows selection"

and certainly not be unconditional.

x42:
PS. internally Ardour doesn't expect discontinuous changes of the loop range while looping. The proposed change may cause audible artifacts

PR #1087, opened 2026-03-24
Extend MIDI parser and model to keep the meta-event in the tracks

keryell:
## Better MIDI meta-event support, a headless import API, and related fixes

This started as "preserve MIDI meta-events when importing SMF files" and grew into a broader set of MIDI-import improvements. It addresses:

- https://discourse.ardour.org/t/importing-midi-files-with-most-metadata/113061
- and, as a side effect, helps with lyrics https://discourse.ardour.org/t/lyrics-integrated-into-ardour/113321

following the meta-event definitions in:

- *Standard MIDI Files 1.0*, in *The Complete MIDI 1.0 Detailed Specification, Document Version 96.1*, MIDI Manufacturers Association.
- *Recommended Practice (RP-019): SMF Device Name and Program Name Meta Events*, 1999, MIDI Manufacturers Association.

### 1. Preserve SMF meta-events through import / model / export

Text meta-events (title, copyright, track/instrument name, lyrics, markers, program/device name — types `0x01`–`0x09`) now survive a round-trip instead of being discarded on import:

- The Evoral sequence/model keeps meta-events in the tracks rather than dropping them.
- `SMF::is_meta` recognises the RP-019 `0x08`/`0x09` (Program Name / Device Name) types.
- Meta-events are written back out to the MIDI files under `interchange/.../midifiles/`.
- Type-0 (single-track) files are handled as well as type-1.
- `SMF::read_event` hardened along the way.

Meta-events are imported when **"Import MIDI markers"** is enabled in the *Add Existing Media* dialog.

### 2. Lua API for MIDI meta-events

- The MIDI model is exposed to Lua, with helpers to list meta-events and read their raw buffers.
- An example script dumps the non-note content (meta-events) of MIDI regions/files, for inspection and for lyric workflows.

### 3. Headless MIDI import: `ARDOUR.LuaAPI.import_midi`

A new libardour-level binding that imports an SMF into a session entirely without the GUI — the GTK-free distillation of `Editor::do_import` (SMF tempo-map import + `Session::import_files` + track/region creation). It works both headless (`ardour-lua`) and from in-GUI scripts, with options for tempo map, markers, and per-channel splitting. Covered by a CppUnit test.

### 4. SMF section markers → timeline markers

- **Bug fix:** `SMF::load_markers` only scanned one track, so type-1 files lost their markers (only the last track's name survived). It now scans every track. `MarkerAt` records the meta-event type so consumers can tell markers from lyrics/text.
- `import_midi` adds the `0x06` Marker events ("Part A", "Part B", …) as session timeline Location markers, so song sections appear on the ruler.

### 5. Shutdown crash fix

`pthread_cancel_all()` at process exit could abort with *"FATAL: exception not rethrown"* when it cancelled the background MIDNAM loader mid-parse; the loader's `catch (...)` swallowed the forced-unwind. Now rethrown. (Also submitted standalone as #1125.)

### 6. Demo script

`import_and_play.sh` — a reusable showcase that imports + channel-splits a MIDI file via `import_midi`, drops a GM synth (a-fluidsynth + soundfont) on each track, launches the GUI under PipeWire/JACK, and connects the master to the speakers.

### Testing

- CppUnit suite for the import binding (success + rejection cases).
- Unit test for meta-event preservation.
- Verified end-to-end on real abc2midi arrangements (multi-track, lyrics, section markers).

ardour-lyrics-meta-events


pauldavisthefirst:
Was this written using an LLM ? 

Parts of it appear useful and reasonable. It isn't clear to me that meta-events belong in our Sequence object, but I'll give it some thought.

libs/evoral/SMF.cc
pauldavisthefirst, line 1: It seems incorrect to me that the return 0 statements here are deleted. This value is used by the caller to indicate the specific condition of an Ardour-specific sequencer-specific event.

keryell:
> Was this written using an LLM ?

Yes, most of it.

> Parts of it appear useful and reasonable. It isn't clear to me that meta-events belong in our Sequence object, but I'll give it some thought.

The goal of this PR is to get some thoughts.
I am not sure about what you mean, since I do not have the big picture.
If I want do some processing on the MIDI track according to the value of some meta-events of the track, I have the feeling these should be in the track. It is possible to dive into the `interchange/.../midifiles` but that seems inefficient and requires to have the session saved first.

libs/evoral/SMF.cc
keryell: Isn't it the idea, if we want the meta-event to be returned so it is in the track, by copying it on line 377?
```c++
memcpy(*buf, event->midi_buffer, size_t(event_size));
```
There is still later on line 386:
```c++
		if (is_meta) {
			return 0;
		}
```
left for the caller interpretation.
Actually, looking at the change you made in https://github.com/Ardour/ardour/commit/1ee76f41789862e45acc28a4f46440b0198d7cb8 shows that it is actually the `size == 0` which is used to discard the meta-event, not the the return value.

Your comment made be look at Type 0 track and there the meta-event was not kept. I have updated it a new commit.

keryell:
I am still working on this but I have been distracted by improving the tool used to produce the right MIDI file at the first place https://github.com/sshlien/abcmidi

keryell:
This allows a fully scripted solution I use for my bands to go from an ABC source file to a running Ardour playing/recording session along abc2midi https://github.com/sshlien/abcmidi/
The demo script https://github.com/keryell/ardour/blob/f7ddb5c843102416c29fc40edaf59a133ddaa21f/import_and_play.sh is hard-coded to use https://github.com/keryell/musique/tree/master/rond_saint-vincent/anneaux_d_or as an example.

PR #1091, opened 2026-04-10
improve jack port diagnostics and ignore other ports to improve responsiveness

udoprog:
I've had an issue with jack/pipewire where meters and audio stutters as you alt+tab or hover over panel icons in kwin.

The cause seems to be churn induced by the rapid addition and removal of video nodes in pipewire that are used by kwin to stream the window preview. The result is that for low latency configurations on my system (64 samples or less) hovering over a panel icon or alt+tabbing causes significant stuttering in the UI and xruns, presumably because ports are being reconfigured.

In this patch I've added code to ignore jack ports with the type `"other"` early. This encompasses the short-lived video nodes in kwin and avoids most of this churn on my system. I've also improved diagnostics for when ports are added / removed.

This is a video showcasing how meters are affected on my system without this patch (sorry, couldn't do audio right now, but it is also stuttering):
[Screencast_20260410_221530.webm](https://github.com/user-attachments/assets/5b4351ef-8edf-4490-9078-7a825e824e7c)

With the patch the vast majority of stuttering is avoided.

At the moment I'm just looking for a review. I'm not sure whether `"other"` ports really aren't used. It would be nice to include some option I could use to avoid this stuttering because it's the biggest issue I'm having at the moment. I want to be able to alt+tab while working without stuttering.

### DIFF
--- a/libs/backends/jack/jack_audiobackend.h
+++ b/libs/backends/jack/jack_audiobackend.h
@@ -301,7 +301,7 @@ class JACKAudioBackend : public AudioBackend {
 	/* used to manage _jack_ports, specifically for ports belonging to the
 	   JACK backend or other clients.
 	*/
-	void jack_registration_callback (jack_port_id_t, int);
+	bool jack_registration_callback (jack_port_id_t, int);
 
 	typedef std::map<std::string,std::shared_ptr<JackPort> > JackPorts;
 	mutable SerializedRCUManager<JackPorts> _jack_ports; /* can be modified in ::get_port_by_name () */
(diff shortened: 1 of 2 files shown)


pauldavisthefirst:
This is presumably with pipewire? JACK1 or JACK2 don't support video ports?

udoprog:
That is correct, this is with pipewire.

pauldavisthefirst:
My main review comment would be to change the debug output to use DEBUG_TRACE (DEBUG::Ports,...)

Otherwise, seems like a sensible response to a (relatively) new situation ...

pauldavisthefirst:
I'll mention this to wtay however, since it seems like a bug for JACK to report on port types that it doesn't support ... unless of course the PW implementation does in fact support a video port type (which it might).

pauldavisthefirst:

PR #1092, opened 2026-04-12
Perform horizontal scrolling when mouse is on the midi region and shift + scroll

aivelon:
When mouse is on the midi region on the track and whether a midi note is selected or not, it should scroll horizontally when shift + scroll is performed. That is how it used to work in 8.x version.
Right now, when shift+scroll is performed, then it changes velocity when midi note is selected, or performs normal vertical scroll when not selected. It basically ignores vertical scroll.
The regression happened from this commit: c509de6814

Resolves: https://tracker.ardour.org/view.php?id=10108


pauldavisthefirst:
manually rebased, and merged. thanks!

PR #1094, opened 2026-04-13
Lua script that creates MIDI tracks with drum names

tonilink:
Based on the original script by PSmith.


tonilink:
I'll close this pull request because I made a new Icon for it. Something that looks like more a Midi drum sequencer.

PR #1095, opened 2026-04-13
Lua script, new MIDI drums tracks

tonilink:
Lua script that creates new MIDI tracks for drums
(New version with a different icon that resembles more a midi sequencer)


luzpaz:
Can you show  a screencast of this feature in action ?

tonilink:
Here: https://youtu.be/4xkyq4XfBzQ

PR #1096, opened 2026-04-14
start rejecting invalid session state (fixes crash in #10274)

udoprog:
In this patch I've fixed loading of sessions that contain "invalid" float values which causes various crashes I've found in relation to https://tracker.ardour.org/view.php?id=10274 when I tried to load a corrupted session file.

#### Background

While it's not expected that Ardour will save infinites or nans to a session file, it is a format that is specifically supported by the string_to_* family of functions (either explicitly or implicitly through the glib double-to-string implementation), so hardening session loading against memory issues when these are encountered is important. Session files are just regular files after all, and might be shared across systems. It is in my view a useful goal that Ardour tries to eliminate potential vectors for logic bugs that might lead to more serious issues by validating untrusted inputs.

#### Implementation

This is just the first step. I removed support for loading floats from the existing `get_property` function and added another called `get_property_or_reject`. This was just done to ensure that all cases where floats are involved are caught as compilation errors. Additionally the last parameter acts as a filter, that can be used to perform more input validation. I've only included one such example right now that I think is correct, but more could be added.

I've also added a non-fallible (as in it cannot fail) variant which is simply called `get_property_or_default`, that if a property is rejected will simply use a default value instead.

I think the behavior of `get_property_or_reject` could and probably should more-or-less be merged into `get_property`. Right now I'm keeping it separate because 1) it's easier for my editor to search for it and 2) I personally don't like overloads that differ too much. Once filters have been added everywhere appropriate the remaining calls can be folded back into `get_property` with a default permissible filter.

Since how to approach this is matter of opinion I'm open to feedback. Consider this patch a suggestion.

### DIFF
--- a/libs/pbd/pbd/string_convert.h
+++ b/libs/pbd/pbd/string_convert.h
@@ -20,6 +20,7 @@
 #define PBD_STRING_CONVERT_H
 
 #include <string>
+#include <functional>
 #include <stdint.h>
 
 #include "pbd/libpbd_visibility.h"
@@ -64,8 +65,10 @@ LIBPBD_API bool string_to_int64 (const std::string& str, int64_t& val);
 
 LIBPBD_API bool string_to_uint64 (const std::string& str, uint64_t& val);
 
+LIBPBD_API bool string_to_float (const std::string& str, float& val, std::function<bool(const float&)> filter);
 LIBPBD_API bool string_to_float (const std::string& str, float& val);
 
+LIBPBD_API bool string_to_double (const std::string& str, double& val, std::function<bool(const double&)> filter);
 LIBPBD_API bool string_to_double (const std::string& str, double& val);
 
 template <class T>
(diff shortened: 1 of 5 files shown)


x42:
While I'm a fan of "be liberal what you accept", I'd not bother. Here it is better to crash than to load partial state or incorrect state by falling back to default values.

(There will also be a lot of changes required for 3+ derivative projects)

PS. In 99% of all cases where a session does not load, using a backup session to continue is preferable.

udoprog:
> While I'm a fan of "be liberal what you accept", I'd not bother. Here it is better to crash than to load partial state or incorrect state by falling back to default values.

We can make the methods crash intentionally and with a message. My preference would be to error and present a dialog to the user ("Could not load corrupt session").

> (There will also be a lot of changes required for 3+ derivative projects)

The old methods can be kept, so no changes would be needed unless the downstream projects want the changes.

> PS. In 99% of all cases where a session does not load, using a backup session to continue is preferable.

👍 This is understandable.

What say you? I don't really mind working on it, but I'd like to hear your disposition first.

x42:
> My preference would be to error and present a dialog to the user ("Could not load corrupt session").

Yeah, that's the ideal solution, and what already happens in various other circumstances.

It might be as simple as throwing a C++ exception in `node_get_property`. So a *one-line change* (maybe not literally).

udoprog:
I'm getting the impression the current direction when facing "invalid state" leans towards skipping over corrupt state (e.g. see bool which has a fixed set of legal representations) or read straight into a member field ignoring the `bool` outcome (fall back to existing / default on errors). Are you thinking they could also opt to throw an exception or just non-finite floats?

On throwing exceptions, I am concerned about exception safety due to the wide uses of set_state-like methods I'm seeing. What's your vibe on the current state of ardour regarding that?

udoprog:
Since I don't know what to do I've simply reduced the scope of the change to:

1. Reject infs and nans as legal float values when loading state.
2. Add `get_property_if` which takes a filter function and use it in a few instances.

I've kept 2. and used it in one instance where if we accepted `0`, it would immediately divide by zero.

PR #1098, opened 2026-04-14
a-fluidsynth: add configurable velocity spread

cyrusadkisson:
## Summary
- add configurable velocity spread to ACE Fluid Synth note-on handling
- add spread tightness options for wider or tighter humanization
- sort generic plugin enumeration menus by numeric value so the tightness menu appears in the intended order

## Screenshot
ACE Fluid Synth velocity spread UI

## Testing
- ./waf -j16
- verified the ACE Fluid Synth UI updates render and the plugin rebuilds cleanly


x42:
This will not only break existing sessions, but it is also the wrong approach. We prefer DRY design.
This function does not belong in the plugin, and you an already achieve his by using a MIDI filter before the synth,
or preferably using deterministic randomness so it sounds the same on every export using Ardour's MIDI transform.

x42:
> sort generic plugin enumeration menus by numeric value so the tightness menu appears in the intended order

This is also not what is supposed to happen.  Unless the plugin specifies a https://lv2plug.in/ns/ext/port-props#displayPriority
there is heuristic grouping by name.

luzpaz:
@cyrusadkisson are you perhaps planning to change-up the approach ?

luzpaz:
@cyrusadkisson ?

cyrusadkisson:
I'm using the velocity randomizer here instead. https://x42-plugins.com/x42/x42-midifilter

PR #1101, opened 2026-04-28
Multiple fixes and improvements on Windows/MSYS2 builds

xris1658:

PR #1104, opened 2026-05-05
Fix DPF MIDI CC regex

xris1658:
Match dot with `\.` instead of `.`


pauldavisthefirst:
manually rebased and merged. Thanks for noticing! Did this actually cause any problems?

xris1658:
Not really :) Just noticed this while working on PR #1101.

PR #1105, opened 2026-05-06
Fix seg-fault when LV2 plugins have “non-URI” features

smondet:
(disclaimer: not an LV2 expert, at all!)

I was trying the plugins at <https://github.com/rerdavies/ToobAmp> and they make Ardour seg-fault.

The issue is that calling `lilv_node_as_uri` on a non-URI “feature” returns `NULL` which then fails at the first `strcmp`.

This PR just guards the call with `lilv_node_is_uri(..)`.


x42:
Non URI features are not allowed by the LV2 standard. Please ask the plugin author to fix and validate [1] the plugin.

Another invalid plugin crashes plugin host. News at 11. :)

[1] https://lv2plug.in/pages/validating-lv2-data.html

x42:
A more correct version of the fix would check for `!lilv_node_is_uri ()`  instead of testing if it's a string.
and if so also error out, instead of ignoring it,  and write log to `PluginScanLogEntry` instead of Log Window.

smondet:
> A more correct version of the fix would check for !lilv_node_is_uri () instead of testing if it's a string.

It does, the `is_string` is to display the warning then there is a `is_uri`

x42:
Rebased, cleaned up (indenting, styleguide) and merged as 9.3-6 16576fea16
Thank you!

PR #1106, opened 2026-05-07
update german translation

edogawa23:
I left out all the MIDI CC names for now, it's entirely common to refer to these by their english names, in german.
I tried my best to find reasonable wording in Chord editor, especially for the Drop operations I'm not sure how these should be called, as I'm not aware of these being so common, and couldn't find anything during a brief research.

### DIFF
--- a/gtk2_ardour/ardour_ui.cc
+++ b/gtk2_ardour/ardour_ui.cc
@@ -3019,12 +3019,11 @@ ARDOUR_UI::pending_state_dialog ()
 	ArdourDialog dialog (_("Crash Recovery"), true);
 	Label  message (string_compose (_("\
 This session appears to have been modified\n\
-without save, or in middle of recording when\n\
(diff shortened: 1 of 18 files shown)


edogawa23:
What's happening? I see this got committed but the PR's changes tab still says it's not... ?

jean-emmanuel:
The PR was manually merged, it can be closed :).

prokoudine:
That said, @x42 pushed a small translation-breaking change, so you should feel safe sending another commit to this PR and then closing it once I merge manually

x42:
> That said, @x42 pushed a small translation-breaking change

I assume you refer to  https://forum.harrisonconsoles.com/thread-13066-post-74352.html#pid74352

prokoudine:
@x42 I really meant to say "string freeze breaking", although we didn't announce an official strings freeze this time  :)

PR #1108, opened 2026-05-17
mcp_http: split multi-form tools into explicit variants

zabooma:
Implements an MCP tool-schema compatibility follow-up for clients that reject top-level `oneOf` / multi-form input
  schemas.

  ### Summary

  - Split tools with multiple location/input forms into explicit variants, for example:
    - `transport_locate_sample`
    - `transport_locate_bbt`
    - `transport_loop_location_samples`
    - `transport_loop_location_bbt`
  - Added dispatch aliases so the new public tool names reuse the existing implementation paths.
  - Updated tool metadata to avoid schemas that Azure/OpenAI-style function-call validators reject.
  - Keeps behavior equivalent to the existing tools, but makes the MCP surface easier for stricter clients and models
  to call correctly.

  This is intended as a compatibility improvement, not a functional change to Ardour session behavior.

### DIFF
--- a/libs/surfaces/mcp_http/mcp_http_server.cc
+++ b/libs/surfaces/mcp_http/mcp_http_server.cc
@@ -151,13 +151,76 @@ json_escape (const std::string& s)
 static std::string
 canonical_tool_name (std::string tool_name)
 {
+	static const struct {
+		const char* alias;
+		const char* canonical;
+	} aliases[] = {
+		{"transport/locate_sample", "transport/locate"},
+		{"transport/locate_bbt", "transport/locate"},
+		{"transport/loop_location_samples", "transport/loop_location"},
+		{"transport/loop_location_bbt", "transport/loop_location"},
+		{"markers/add_range_samples", "markers/add_range"},
+		{"markers/add_range_bbt", "markers/add_range"},
+		{"markers/set_auto_loop_samples", "markers/set_auto_loop"},
+		{"markers/set_auto_loop_bbt", "markers/set_auto_loop"},
+		{"markers/set_auto_punch_samples", "markers/set_auto_punch"},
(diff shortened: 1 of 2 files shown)


zabooma:
@x42 Here is one followup PR for the MCP. I noticed that some models had issues with the MCP metadata when going through Azure gateway. this PR simplifies the schema only, no changes in the functionality.

x42:
Rebased and merged as  9.4-32 402e5747d0
Thanks.

--

PS. Since you are testing different models.

Toying around with it, I used gemini to write a 12 bar blues piano, drums, bass. This kinda worked, but all MIDI notes were exactly 1 tick long: Note start Start == End.  so it sounded weird.  Is that an oddity of gemini or related to the issue with BBT here?

zabooma:
> Rebased and merged as 9.4-32 [402e574](https://github.com/Ardour/ardour/commit/402e5747d0a5b26cd805e3c32fd5b29c052641c1) Thanks.
> 
> --
> 
> PS. Since you are testing different models.
> 
> Toying around with it, I used gemini to write a 12 bar blues piano, drums, bass. This kinda worked, but all MIDI notes were exactly 1 tick long: Note start Start == End. so it sounded weird. Is that an oddity of gemini or related to the issue with BBT here?

It seems that it used "drum mode" for all tracks. let me try this here with all agents I can run.

Thanks for merging!

x42:
> It seems that it used "drum mode" for all tracks. let me try this here with all agents I can run.

I didn't expect "drum mode" to be thing for LLM MIDI. Especially since it did not switch the MIDI track(s) to percussive view. The prompt I used was very basic:

"Write a 12 bar blues using piano, bass and drum MIDI tracks in the current Ardour session"

Anyway not a biggie.

zabooma:
#1110 will address this issue. the json bulk import schema was overly complicated and models were often confused with how to use it.

PR #1110, opened 2026-05-18
mcp_http: simplify MIDI JSON import format

zabooma:
## Summary

  This updates the Ardour MCP MIDI JSON import format to make bulk note generation simpler and more reliable for MCP clients and LLM-based assistants.

  The previous format exposed low-level `note_on` / `note_off` events, optional comment events, and multiple event shapes. Some clients handled this correctly, but others were confused by the schema and generated invalid or unnecessarily verbose payloads.

  This change makes `midi_events` a list of note objects instead.

  ## What Changed

  - Replaced `note_on` / `note_off` import pairing with one note object per note.
  - Added `l` as the note length field, measured in Ardour beats.
  - Kept `repeat` support for copying notes from another relative bar.
  - Removed comment event support from the import schema.
  - Changed `repeat` to require an explicit source bar number, with no special `repeat: 0` behavior.
  - Updated `midi_note_get_json` so exported JSON uses the same import-compatible note format.
  - Updated MCP tool descriptions and README documentation.

  ## New MIDI Event Shape

  Example melodic note:

  ```json
  {
    "bar": 1,
    "b": 1,
    "n": 60,
    "v": 80,
    "l": 0.5
  }
```
  Meaning:

  - bar: region-relative bar number, starting at 1.
  - b: beat within the bar, starting at 1. Decimal values are allowed.
  - n: MIDI note number.
  - v: velocity. Optional, defaults to 64.
  - l: note length in Ardour beats. Optional.

  Defaults:

  - Melodic mode: omitted l defaults to 1.0.
  - Drum mode: omitted l defaults to 0.0.

  ## Repeat Example

  ```json
  {
    "bar": 2,
    "repeat": 1
  }
```

  This copies all notes from relative bar 1 into relative bar 2.

  ## Rationale

  The simplified schema is easier for MCP clients and LLMs to use correctly. It avoids requiring clients to pair note-on and note-off events, while still supporting compact bulk MIDI generation and repeated bar patterns.

  ## Testing

  - Verified tools_json.inc parses as valid JSON.
  - Ran git diff --check.
  - Built successfully with ./build-macos.sh.
  - Tested with Gemini CLI; it generated a correct bulk JSON import payload in one attempt.

   [Sample prompts and system prompt suggestion](https://manual.ardour.org/using-control-surfaces/mcp-http/)

### DIFF
--- a/libs/surfaces/mcp_http/README.md
+++ b/libs/surfaces/mcp_http/README.md
@@ -135,11 +135,20 @@ Use the JSON tools for fast note generation and round-trip editing:
 - Beat and tick fields:
   - `b` is beat number in the bar and may be fractional. Example: `b: 1.5` is an off-beat 8th note in 4/4.
   - `t` is additional tick offset inside the beat, using `ticks_per_quarter`.
+- Note length:
+  - `l` is note length in Ardour beats and may be fractional. Example: `l: 0.5` is an 8th note in 4/4.
+  - For melodic MIDI, omit `l` to use a default length of 1 beat.
+- Repeat bars:
+  - Use `{ "bar": 2, "repeat": 1 }` to copy all notes from relative bar 1 into relative bar 2.
+  - A repeat item should only include `bar` and `repeat`; do not include note fields on the same item.
 - Drum mode behavior:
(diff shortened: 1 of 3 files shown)


x42:
Thanks for update on short notice! Rebased and merged as 9.4-47 4724b12071

PR #1111, opened 2026-05-19
feat: allow bypassing plugins using midi generic control surface

hatch01:
Answer: https://discourse.ardour.org/t/use-midi-to-control-events-in-ardour/81124
> We do not have binding to plugin bypass buttons, but it would be relatively easy to do. You should add it as a feature request on the bug tracker.

PR is still in development, not fully tested.

### DIFF
--- a/gtk2_ardour/plugin_ui.cc
+++ b/gtk2_ardour/plugin_ui.cc
@@ -606,6 +606,7 @@ PlugUIBase::PlugUIBase (std::shared_ptr<PlugInsertBase> pib)
 	if (_pi) {
 		_pi->ActiveChanged.connect (active_connection, invalidator (*this), std::bind (&PlugUIBase::processor_active_changed, this, std::weak_ptr<Processor> (_pi)), gui_context ());
 		_bypass_button.set_active (!_pi->enabled ());
+		_bypass_button.set_controllable (_pi->bypass_control ());
 	} else {
 		_bypass_button.set_sensitive (false);
 	}
--- a/gtk2_ardour/processor_box.cc
+++ b/gtk2_ardour/processor_box.cc
@@ -242,6 +242,10 @@ ProcessorEntry::ProcessorEntry (ProcessorBox* parent, std::shared_ptr<Processor>
 		_processor->PropertyChanged.connect (name_connection, invalidator (*this), std::bind (&ProcessorEntry::processor_property_changed, this, _1), gui_context());
 		_processor->ConfigurationChanged.connect (config_connection, invalidator (*this), std::bind (&ProcessorEntry::processor_configuration_changed, this, _1, _2), gui_context());
 
+		if (pi) {
+			_button.set_controllable (pi->bypass_control ());
+		}
+
 		const uint32_t limit_inline_controls = UIConfiguration::instance().get_max_inline_controls ();
 
 		set<Evoral::Parameter> p = _processor->what_can_be_automated ();
--- a/gtk2_ardour/route_properties_box.cc
+++ b/gtk2_ardour/route_properties_box.cc
@@ -98,6 +98,12 @@ ProcessorUIFrame::ProcessorUIFrame (std::shared_ptr<Route> r, std::shared_ptr<Pr
 	set_tooltip (&_enable_btn, _("Bypass"));
 
 	_enable_btn.set_active (_proc->enabled ());
+	{
+		std::shared_ptr<PluginInsert> pi = std::dynamic_pointer_cast<PluginInsert> (p);
+		if (pi) {
+			_enable_btn.set_controllable (pi->bypass_control ());
+		}
+	}
 	_collapse_btn.signal_clicked.connect ([&] () {
 			bool a = !_collapse_btn.get_active ();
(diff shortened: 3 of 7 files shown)


x42:
I am somewhat torn if this is a good idea. Plugins that have an automatable control can already use MIDI bindings using the respective bypass/enable control in the plugin UI, or editor automation header slider or  Mixer strip inline control..

besides, it should be as simple as
```c++
  _bypass_button->set_controllable (_pi->bypass_control ());
}
```

and the default ctrl+middle click MIDI learn works.

x42:
try: 
```diff
diff --git a/gtk2_ardour/plugin_ui.cc b/gtk2_ardour/plugin_ui.cc
index ba11a781b0..3c30a93db8 100644
--- a/gtk2_ardour/plugin_ui.cc
+++ b/gtk2_ardour/plugin_ui.cc
@@ -606,6 +606,7 @@ PlugUIBase::PlugUIBase (std::shared_ptr<PlugInsertBase> pib)
        if (_pi) {
                _pi->ActiveChanged.connect (active_connection, invalidator (*this), std::bind (&PlugUIBase::processor_active_changed, this, std::weak_ptr<Processor> (_pi)), gui_context ());
                _bypass_button.set_active (!_pi->enabled ());
+               _bypass_button.set_controllable (_pi->bypass_control ());
        } else {
                _bypass_button.set_sensitive (false);
        }
diff --git a/libs/ardour/ardour/plugin_insert.h b/libs/ardour/ardour/plugin_insert.h
index b58a686526..0c8552fab9 100644
--- a/libs/ardour/ardour/plugin_insert.h
+++ b/libs/ardour/ardour/plugin_insert.h
@@ -243,6 +243,8 @@ public:
 
        std::shared_ptr<ReadOnlyControl> control_output (uint32_t) const;
 
+       std::shared_ptr<PBD::Controllable> bypass_control () const;
+
        std::string describe_parameter (Evoral::Parameter param);
 
        samplecnt_t signal_latency () const;
diff --git a/libs/ardour/plugin_insert.cc b/libs/ardour/plugin_insert.cc
index eb2a77d684..89aebc3e58 100644
--- a/libs/ardour/plugin_insert.cc
+++ b/libs/ardour/plugin_insert.cc
@@ -2975,6 +2975,15 @@ PluginInsert::control_output (uint32_t num) const
        }
 }
 
+std::shared_ptr<PBD::Controllable>
+PluginInsert::bypass_control () const
+{
+       if (_bypass_port == UINT32_MAX) {
+               return std::shared_ptr<PBD::Controllable> ();
+       }
+       return std::const_pointer_cast<AutomationControl> (automation_control (Evoral::Parameter (PluginAutomation, 0, _bypass_port)));
+}
+
 string
 PluginInsert::describe_parameter (Evoral::Parameter param)
 {
```

hatch01:
> I am somewhat torn if this is a good idea. Plugins that have an automatable control can already use MIDI bindings using the respective bypass/enable control in the plugin UI, or editor automation header slider or Mixer strip inline control..
> 
> besides, it should be as simple as
> 
> ```c++
>   _bypass_button->set_controllable (_pi->bypass_control ());
> }
> ```
> 
> and the default ctrl+middle click MIDI learn works.

Yes, in a perfect world, this would be enough to use integrated bypass in plugins.
However, some don't have bypass (for example, a lot of Guitarix plugins or Dragonfly, I think)

hatch01:
Thank you for your help !
This is a good start; it seems to work well in the plugin fullscreen window but not yet in the bottom bar and also in the processor box.
I'll keep working on it later

x42:
> Yes, in a perfect world, this would be enough to use integrated bypass in plugins.
> However, some don't have bypass (for example, a lot of Guitarix plugins or Dragonfly, I think)

In which case the plugin should add it, instead of the host working around it.  Hard bypassing a reverb sounds or guitar FX like distortion sounds like a bad idea. There will be audible artifacts when doing so.

However In this case `PluginInsert::bypass_control ()` returns a null-pointer. So you cannot bind it to begin with, so it's fine.

Except it can he hard to explain to users why only some plugin's bypass can be used to MIDI learn.

>This is a good start; it seems to work well in the plugin fullscreen window but not yet in the bottom bar and also in the processor box.

Indeed. Different a buttons is used for the bottom attachment in collapsed view  (same mechanism should appluy there, and I'd leave the processor-box LED alone).

hatch01:
>  Hard bypassing a reverb sounds or guitar FX like distortion sounds like a bad idea. There will be audible artifacts when doing so.

Of course artifacts could happen, but this is already possible in Ardour to do so. This PR just adds the bindings to do it using MIDI control.
We could probably do it today with MIDI -> OSC -> ardour.

For example, in a workflow for live concerts, when switching guitar effects between songs, it would most likely be ok and easier than asking our tech guy to switch and not forget any guitars.


In a more technical way, my last push is working well functionally.

hatch01:
Hmm, after more analysis, I understood that plugin bypass is not saved in the sessions (if it's not an internal bypass)
So my solution now needs to record the MIDI bindings at every Ardour restart.

hatch01:
Need #1122 to be merged first.

PR #1112, opened 2026-05-21
Display relevant MIDI channel on show patch selector

gfontanilla:
**Context:**

Rather than virtual instruments, I rely more on external synthesizers for sound generation. To maximize the use of a mutli-timbral synth, the most logical setup for me is to have one Ardour MIDI track for each available channel on the synth via the channel selector "Inbound" and "Playback" settings. The MIDNAM selector and patch selector come in handy when deciding on which sounds to use, and the restore patch feature is convenient as it eliminates the need to save the settings into a scene/performance on the synth itself. 

**Issue:**

After selecting a patch on a particular track, closing and and then reopening the patch selector, it always defaults to displaying channel 1 even if the previously selected patch is on a different channel. From what I understand, there can be a program/patch change set per channel (1-16). However, given the setup described above, I'm only interested in one particular channel per MIDI track. As such, each time I reopen the patch selector, I have to recall which channel the MIDI track is "assigned" to, and then manually select that in the dropdown.

**Proposal:**

This code change makes it so that the patch selector displays the relevant MIDI channel, reusing part of the logic found in `MidiTimeAxisView::get_preferred_midi_channel`. I would have liked to call the method itself, but it doesn't seem possible to do it from `PatchChangeWidget` due to the inheritance setup.

Ideally, the patch selector view "remembers" the state it was in prior to closing, but at the same time I don't think it's efficient to do that for each MIDI track and I have yet to find a similar pattern in other parts of the code (I'm still pretty new, so I'm happy to take suggestions/improvements).

display_relevant_channel

### DIFF
--- a/gtk2_ardour/midi_util.h
+++ b/gtk2_ardour/midi_util.h
@@ -42,6 +42,7 @@ inline static void clamp_to_0_127(uint8_t &val)
 	}
 }
 
+uint8_t get_preferred_midi_channel_from_chn_mask (uint16_t chn_mask);
 
 void
 build_controller_menu (Gtk::Menu& menu, ARDOUR::InstrumentInfo const & instrument_info, uint16_t channel_mask,
--- a/gtk2_ardour/midi_time_axis.cc
+++ b/gtk2_ardour/midi_time_axis.cc
@@ -1762,25 +1762,7 @@ MidiTimeAxisView::get_preferred_midi_channel () const
 	}
 
 	uint16_t const chn_mask = midi_track()->get_playback_channel_mask();
-	int chn_cnt = 0;
-	uint8_t channel = 0;
-
-	/* pick the highest selected channel, unless all channels are selected,
-	 * which is interpreted to mean channel 1 (zero)
-	 */
-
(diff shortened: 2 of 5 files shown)


gtk2_ardour/midi_util.h
```diff
@@ -42,6 +42,28 @@ inline static void clamp_to_0_127(uint8_t &val)
 	}
 }
 
+inline static uint8_t get_preferred_midi_channel_from_chn_mask (uint16_t chn_mask)
```
x42: Why did you chose to use `inline static` here instead of implementing this in  gtk2_ardour/midi_util.cc?

gfontanilla: To be honest, `inline static` is new to me (though I did look it up), but my intention here was to follow the pattern set by `clamp_to_0_127`, the function just above this one. I guess I should have just implemented it in `gtk2_ardour/midi_util.cc`. I'll try that and push a commit after testing.

Thanks for the feedback!

gfontanilla:
Moved function implementation to `gtk2_ardour/midi_util.cc`, squashed and rebased. For review :pray: 

@x42

PR #1113, opened 2026-05-21
Fix suspicious condition

gfontanilla:
While looking through the MMC implementation, I came upon this suspicious looking condition. Oddly enough, it's been around for a while https://github.com/Ardour/ardour/commit/41b2de41d69c2ecc381867e502e4760267275425 . This just fixes it to what it likely should have been.


pauldavisthefirst:
manually rebased and merged ,thanks for noticing this.

PR #1118, opened 2026-05-30
Use natural sort for plugin manager

gfontanilla:
Fix for issue #0010372

Related PR: https://github.com/Ardour/ardour/pull/1116

Screenshot at 2026-05-30 12-53-06

PR #1119, opened 2026-05-31
Use natural sort for clips sidebar

gfontanilla:
Fix for issue #0010374

Related PR: https://github.com/Ardour/ardour/pull/1118

Screenshot at 2026-05-31 01-11-12

**Note**: In this PR, the function `naturally_less` is overloaded to accept `std::string` params. When passing as a callback to `std::sort`, the function has to be wrapped in a lambda since the compiler cannot differentiate between overloaded functions when used in a callback in this case. An alternative is to not overload/use a different function name to avoid the lambda wrapping.


pauldavisthefirst:

PR #1122, opened 2026-06-02
feat: promote host-level plugin bypass to a standard persistable control

hatch01:
For plugins that lack a native bypass port, Ardour previously relied on a simple boolean _active state that bypassed the standard Evoral::ControlSet. 
Because it was not a standard AutomationControl, it was not saved with state serialization, Mixer Scenes, and MIDI binding persistence, which is very confusing for end-users.

I know this PR might be considered controversial, as, of course, disabling a plugin using automation while music is playing can create audio artifacts. Generally, using real-time automation on this host-level bypass should be discouraged for users.

However, this change is very useful for:
- Not losing enable/disable states across session reloads.
- Switching between Mixer Scenes (e.g., in pause mode, or just to compare two mix versions without worrying about transition artifacts).
- Linking the enable button to a MIDI control surface to control the state using custom hardware.

### DIFF
--- a/libs/ardour/ardour/plug_insert_base.h
+++ b/libs/ardour/ardour/plug_insert_base.h
@@ -109,6 +109,20 @@ public:
 		Variant         _value;
 	};
 
+	/** A host-level bypass control for plugins without native bypass port. */
+	class HostBypassControl : public AutomationControl {
+	public:
+		HostBypassControl (Session& s,
+		                   const Evoral::Parameter& param,
+		                   const ParameterDescriptor& desc,
+		                   std::shared_ptr<AutomationList> list = std::shared_ptr<AutomationList> ());
+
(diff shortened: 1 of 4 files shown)


hatch01:
I've been thinking about it some more, and Carla, for her part, doesn't have a bypass button but does have a dry/wet knob.
This would require more extensive modifications, but perhaps to avoid artifacts, a quick fade from 100% wet to 100% dry could be useful.
This could even be taken further with a user-visible dry/wet control on all plugins to allow for easy parallel processing (I don’t know if this has already been discussed elsewhere).
If we take this concept to its logical conclusion, the knob would simply adjust the dry/wet ratio; consequently, when it comes to automation, we could automate the dry/wet ratio but not the knob itself, which is merely a shortcut for that setting.

PR #1127, opened 2026-06-15
Update Polish translation

songo:
It also contains a fix for the MIDI Tools being too wide in the Polish version.


prokoudine:
Thanks a lot! Cherry-picked and pushed to master, you can close this now

PR #1128, opened 2026-06-17
Fix Bad Optional Access

dehnhardt:
Small fix or bad optional access.


pauldavisthefirst:

PR #1129, opened 2026-06-17
Fix path to ytk .mo files

afh:
Use correct path to `libs/tk/ytk` directory when searching for `.mo` files.


pauldavisthefirst:

afh:
Thanks for the quick response and action, @pauldavisthefirst, much appreciated :100:

PR #1130, opened 2026-06-19
patchfiles: make manufacturer consistent across all E-mu midnams

songo:
In a couple of E-mu midnams, the `<Manufacturer>` field is set to `E-mu`, while in the majority it is `E-mu Systems`. As an E-mu gear user, I find this inconvenient. I think it would be better to standardize them all to `E-mu Systems`.


pauldavisthefirst:
manually rebased and merged, thank you

PR #1132, opened 2026-06-23
Mingw patches

EZ4Stephen:


gtk2_ardour/wscript
```diff
@@ -717,7 +717,7 @@ def build(bld):
         obj.source += [ 'lv2_plugin_ui.cc' ]
         obj.use += [ 'SUIL' ]
 
-    if bld.is_defined('HAVE_FFTW35F') and bld.env['build_target'] != 'mingw' and bld.env['build_target'] != 'msvc':
```
x42, line 720: This will likely break our cross compile setup with mingw.

x42:
we use https://gist.github.com/x42/469f412b51294d5123be82f34906ba3f with mingw to fix the thread-local issue when cross-compiling.  see also https://github.com/msys2/MINGW-packages/issues/2519

x42:
Using a shared_ptr instead of thread-local significantly changes behavior and is also not an viable option when using this with realtime threads.

pauldavisthefirst:
@x42 i think you misunderstood the thread local issue

The changes simply switch to using an accessor function to get the thread local value, rather than using it directly. This is caused by MSVC not allowing cross-DLL access to thread-local values (directly). 

There's no change in whether or not  a shared ptr is used, or whether the var is thread local.

pauldavisthefirst:
Re: the wscript change,  I suspect that the intended result is "on native builds we can always use the fftw3f_threads name for the library, but not on mingw cross-compiles".

so the line is still required but should no longer include a test for MSVC.

gtk2_ardour/wscript
EZ4Stephen: But other platforms that pass having HAVE_FFTW35F (that aren't MSVC) get the `bld.env['LIB_FFTW3F'] += ['fftw3f_threads']`... .

Also, while this may appear only MSYS2-oriented for its pacman repo, I view it moreso as a Windows MINGW edit, as there doesn't appear to be any other MinGW implementation for windows with a package manager. The likelihood of someone on windows trying to get the MinGW packages from vcpkg are very low in my opinion, as the effort needed for doing so would be massive.

I got one idea though. Would have to test locally, but surely combining the mingw filter with the platform not being windows would let your nightly builds remain unchanged while allowing Windows msys2 users to make use of the fftw3f_threads from the MSYS2 pacman repo?

TLDR: Could I share a different edit such that the cross compilation isn't affected, and Windows users (Almost definitely using MSYS2) gain the convenience of installing fftw using pacman instead of building from source?

PR #1133, opened 2026-06-23
Fix USB surface compile when libusb-dev is not installed

bodowi:
libusb is not a required dependency for Ardour, and this change allows Ardour to compile successfully even when libusb-dev is not installed on the system.

The usage of HAVE_USB was not consistent across configure() and build().

I have tested the change in an Ubuntu 26 environment (clean build) and the build was successful.


pauldavisthefirst:

PR #1136, opened 2026-07-23
Zoom to 2 times the length of region/source in bottom pianoroll

aivelon:
Currently, the maximum zoom out level is set to the length of the midi region or the source clip. I want to be able to zoom out further so that its easier to view the whole region with boundaries and to also be able to easily resize it.

**Before**: 

https://github.com/user-attachments/assets/2d080ccd-81fd-41c5-82b2-cc13263ce3c0

**After**:

https://github.com/user-attachments/assets/965621e4-96ab-4c64-98d2-f64fe3a35dcc


pauldavisthefirst:
manually rebased and merged, Thanks, good change.

PR #1137, opened 2026-07-26
Piano roll window set size correctly to allow user resize

aivelon:
Currently, the last size of the piano roll window (when it was open) gets set to the minimum size of the window when its opened again. This change just resets the size correctly from the saved size such that the user can still reduce its size.

**Before:**

https://github.com/user-attachments/assets/947bbe03-b616-475e-9bce-e1c1271d3eac

**After:**

https://github.com/user-attachments/assets/6ff6d2f2-fd86-4b26-9124-e70aeed9bfc3


pauldavisthefirst:
manually rebased and merged. excellent catch, i almost certainly use set_size_request() too often and in the wrong places.

PR #1138, opened 2026-07-26
Use pixel based logic to determine ends lengths for resizing midi notes

aivelon:
I observed that the midi notes have very large length of the region in the midi note where the resize boundary begins. This leads to the note not even being resizable if the note is long and zoomed in enough.
Implemented a pixel length based logic where only a 8px length will be considered as the max length of the region for resizing in the note. This will be applicable even if the note is long and zoomed in. 
Also made the mouse cursor icon limits and the actual resize limits consistent.

**Before:**

https://github.com/user-attachments/assets/43c7ee35-787d-48ab-bd14-b128be770a72

**After:**

https://github.com/user-attachments/assets/362eb220-7e53-41c6-95fd-3869f6f57650

### DIFF
--- a/gtk2_ardour/midi_view.cc
+++ b/gtk2_ardour/midi_view.cc
@@ -4503,14 +4503,14 @@ MidiView::sysex_left (SysEx *)
 }
 
 void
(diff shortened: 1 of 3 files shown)


pauldavisthefirst:
manually rebased and merged. good idea!

PR #1139, opened 2026-07-28
Fix issue with point trim on region/clip end

aivelon:
When a click action is performed (not drag) at the end of a midi region or clip, it trims the clip from the start to a nearby snap point when snapping is on.
The issue is that start/end trimming for point trim was being decided by the mouse button being clicked which is incorrect. It should be decided by the operation being performed (StartTrim/EndTrim).

**Before:**

https://github.com/user-attachments/assets/7800e05d-3162-4fc7-9caa-04c8c42b367c

**After:**

https://github.com/user-attachments/assets/78ff139d-b1fb-4005-9699-51c1e7895c90

### DIFF
--- a/gtk2_ardour/editor.h
+++ b/gtk2_ardour/editor.h
@@ -1878,9 +1878,6 @@ private:
 	void start_canvas_autoscroll (bool allow_horiz, bool allow_vert, const ArdourCanvas::Rect& boundary);
 	void stop_canvas_autoscroll ();
 
-	/* trimming */
-	void point_trim (GdkEvent*, Temporal::timepos_t const &);
-
 	void trim_region_front();
 	void trim_region_back();
 	void trim_region (bool front);
--- a/gtk2_ardour/editor.h
+++ b/gtk2_ardour/editor.h
@@ -1879,7 +1879,7 @@ private:
 	void stop_canvas_autoscroll ();
 
 	/* trimming */
-	void point_trim (GdkEvent*, Temporal::timepos_t const &);
+	void point_trim (GdkEvent*, Temporal::timepos_t const &, bool is_start);
 
 	void trim_region_front();
 	void trim_region_back();
--- a/gtk2_ardour/editor_drag.cc
+++ b/gtk2_ardour/editor_drag.cc
@@ -3014,10 +3014,7 @@ TrimDrag::finished (GdkEvent* event, bool movement_occurred)
 		editing_context.commit_reversible_command ();
 
 	} else {
-		/* no mouse movement */
-		if (adjusted_current_time (event) != adjusted_time (_drags->current_pointer_time (), event, false)) {
-			_editor.point_trim (event, adjusted_current_time (event));
-		}
+		/* no mouse movement. do nothing*/
 	}
 
 	for (list<DraggingView>::const_iterator i = _views.begin (); i != _views.end (); ++i) {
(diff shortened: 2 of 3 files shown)


pauldavisthefirst:
Hmm, I believe that the idea here was be able to start/end trim anywhere in the region by using button 1 or button 2. So that even if you're nominally close to the start, you can end trim ....

aivelon:
I may be missing something here, but this point trim is only called on a start or end drag with no mouse movement. So, when the mouse is at one end of the clip and a click is performed with a nearby snapping point (making this condition true:  https://github.com/Ardour/ardour/pull/1139/changes#diff-35e5d54f5699924b5ea51c7133a3b24ca2334034fd2214877955c22060e62932R3018) then point trim takes place. In this scenario, I think its clear that user wants to trim from left or right based on the mouse position (instead of left or middle click).
Is there some other use case/scenario for point trim which I am missing?

pauldavisthefirst:
the idea was you could click *anywhere* in the region and trim either the start or the end to that point, depending on the button that was used. 

OTOH, I'm not convined that this is necessarily that useful. 

Still, with this change, you could no longer truncate a long region by using a button-2 click near the start ...

aivelon:
Point trimming with left click seems like a bad choice to me since it can cause a lot of accidental trims. But middle click to trim from end seems practical for setting the end of the clip without going to the end. The only reason that was not working right now was due to the condition I mentioned above. Let me just do the middle click trim from end.

Also, should we implement this with a keyboard key + mouse click combination instead of simple left/middle click to make it more intentional?

pauldavisthefirst:
a modifier is ok but there may be none left available.

aivelon:
I looked into modifiers. It looks like alt + click on a region is free to use. But Alt is usually used for turning off snapping, so that means we have to trim without snapping always (even if snapping is on), which does not seem right.
Other modifiers for single left click are all taken. For middle click, simple middle click with no modifiers is free, and also Ctrl + middle click is free. Shall we use these two for start and end trims.
Basically, 
middle click = Trim from start
Ctrl + middle click = Trim from end

Let me know what you think.

jean-emmanuel:
Note that middle click drag is used for horizontally-locked region dragging. I'm not sure to see the point in implementing it with a key modifier since we can already trim to the edit point using a single keyboard shortcut.

jean-emmanuel:
Also @pauldavisthefirst this really just look like a bug, it only occurs when there's something to snap to nearby and it does exactly the opposite of what one could expect (if anything ought to happen at all). ie trimming the start when clicking on the end handle and vice versa. 
 
> Still, with this change, you could no longer truncate a long region by using a button-2 click near the start ...

It doesn't seem to work this way though, button-2 click on a region does nothing (unless exactly on the the trim handles and near a grid line and with snap enabled). 

(edit: wrong @ mention removed)

aivelon:
> Note that middle click drag is used for horizontally-locked region dragging. I'm not sure to see the point in implementing it with a key modifier since we can already trim to the edit point using a single keyboard shortcut.

I have missed the keyboard shortcut to trim to the edit point. If that exists, then it looks like there is no need for this. I believe we can just remove this point trimming on mouse click all together. Right now, it just does not work, except for the scenario mentioned here due to snapping. 
Let me know if that makes sense.

aivelon:
Removed point trimming entirely when click at region end without drag. The behaviour is same as what is captured in this PR description.

aivelon:
@pauldavisthefirst  please review and let me know if there are any concerns.

PR #1141, opened 2026-08-10
Grab focus on new group dialog

aivelon:
The new track group creation dialog does not grab focus when opened. This means the user has to click on the dialog first and then type in the name.

Fix: added present call on show for the dialog.

https://github.com/user-attachments/assets/118b6762-a287-4f84-a1ea-4a0bbd75881f


pauldavisthefirst:
manually rebased and merged, thanks.

PR #1142, opened 2026-08-11
Komplete Kontrol A-Series control surface, plus a hidapi O_CLOEXEC fix

nuketownada:
**Protocol research, captures and probe tooling:** <https://github.com/nuketownada/ardourkomplete> — `docs/a61-hid-map.md` is the map every constant here derives from, and `tools/` holds the Python probes that measured it. `kka_protocol.h` points there too, so it stays findable from the tree rather than only from this description.

Draft, for feedback rather than merge. The surface works on real hardware but is deliberately incomplete, and I would rather learn now that the shape is wrong than after binding the rest of it.

## The hidapi commit is independent and I am happy to split it out

`hidapi: use O_CLOEXEC, not FD_CLOEXEC, in the Linux backend` has nothing to do with this surface and fixes something that affects **maschine2** equally. `libs/hidapi/linux/hid.c` passes `FD_CLOEXEC` to `open()` in three places. It is an `fcntl()` descriptor flag, not an `open()` flag, and its value is 1:

| line | written | actually means |
|---|---|---|
| 419 | `O_RDONLY \| FD_CLOEXEC` | `O_WRONLY` |
| 508 | `O_RDONLY \| FD_CLOEXEC` | `O_WRONLY` |
| 1066 | `O_RDWR \| FD_CLOEXEC` | access mode `3` |

Line 508 is how `hid_enumerate()` reads a candidate's ids in order to filter on them, so it fails for any non-root process and every candidate is skipped — `hid_open(vid, pid, ...)` cannot match anything. Line 1066 is worse because it looks like it works: access mode 3 yields a descriptor with neither `FMODE_READ` nor `FMODE_WRITE`, so `open()` succeeds and `ioctl()` still works, while every `read()`/`write()` returns `EBADF`.

Measured here on Linux 6.18: `hid_enumerate(0, 0)` returned 10 devices while `hid_enumerate(vid, 0)` returned **0 for all seven vendor ids present**; with the fix, 1–6 each.

This restores the file to the upstream snapshot it came from. `8fea1ea42e` vendored libusb/hidapi at `4578ea24`, which uses `O_CLOEXEC` at all three sites and is otherwise byte-identical to the copy here — these three lines are the entire difference. hidapi has used `O_CLOEXEC` there since 2022.

Say the word and it becomes its own PR.

## The surface

One module for the **A25, A49 and A61**. They are one device with three keybeds, and the keybed is not the surface's concern — keys leave over the device's own MIDI port, while knobs, buttons, the 4-D encoder and the display arrive over HID. The entire per-model difference is a name and a key count, so it is a table rather than a class hierarchy. Only the A61 is tested; the other two are probed and checked against the expected HID report-descriptor length, warning loudly on mismatch so the first A25 user files a bug rather than silently misreading controls.

Working: interactive mode (`a0 03 04`, which takes the knobs off the MIDI port and hands over the display), full input decode, transport bound with LED feedback, and hotplug with mode re-assertion.

## Points I would particularly like an opinion on

1. **`available()` reports only whether hidapi initialises** and deliberately does not check for a present device, since returning false unloads the module permanently and a user who plugs in later would never get it back. Correct reading of that contract?
2. **`enumerate()` returns three device names under one manufacturer**, so the GUI shows three rows and the chosen name lands in `cpi->config`, which the factory ignores. `contourdesign` does exactly the same with its three ShuttlePRO variants, so I followed it — but is that the intended use?
3. **Threading.** Session signals are connected with `this` as the trailing `PBD::EventLoop`, so they marshal onto the surface's own thread alongside the read poll; the only lock covers `_handle` and LED state, because `stop()` arrives on the GUI thread. maschine2 has the same GUI-thread teardown race without the lock, so I may have misread the intended pattern.
4. **`hid_open()` by ids is avoided** in favour of enumerating unfiltered and opening by path, even with the hidapi fix in place, so the module still works against an unpatched system hidapi under `USE_EXTERNAL_LIBS`.
5. **Builds by default under `HAVE_HIDAPI`** rather than behind an opt-in flag like `--maschine`, matching the other USB surfaces.

## Not done

Knobs are decoded but unbound, pending a decision on whether they should follow the selected strip or act as a bank across tracks. Banking, mute/solo and 4-D jog are likewise unbound. The display protocol is solved and the surface blanks the panel, but nothing renders glyphs yet.

## Testing

An A61, firmware build 2022-01-05, on Linux 6.18. Every control was exercised individually against the protocol map and checked report by report; transport was verified in both directions against the GUI; unplug/replug was verified to re-assert interactive mode and restore lamp state. Teardown restores MIDI mode and blanks the panel, which matters because once interactive mode is entered the firmware stops drawing the panel and never resumes.

No code or data was taken from `hugovangalen/komplementary-kontrol` (GPL-3); where its published findings agree they are cited as corroboration only, and in two places the map above documents where it is wrong.

### DIFF
--- a/libs/surfaces/komplete_kontrol_a/interface.cc
+++ b/libs/surfaces/komplete_kontrol_a/interface.cc
@@ -56,18 +56,12 @@ new_komplete_kontrol_a (Session* s, std::string const& config)
 		return 0;
 	}
 
-	/* Honour the result.  Ardour's activate() calls set_active() again, and
-	 * our set_active() re-runs start() whenever the protocol is not already
-	 * active -- so handing back an object that failed to start means the
-	 * device is opened twice and every failure is logged twice.  Reporting
-	 * the failure here instead gives one error and lets activate() fail
-	 * cleanly, which is also what makes the GUI checkbox behave.
+	/* Deliberately inactive.  ControlProtocolManager::activate() runs
+	 * set_state() between here and set_active(), so anything that configures
+	 * the surface has to land before the hardware is touched; starting the
+	 * device here would open it against state that has not been restored yet.
+	 * Failure to start is reported by activate(), which tears us down.
 	 */
-	if (kka->set_active (true)) {
-		delete kka;
-		return 0;
-	}
-
 	return kka;
 }
 
--- a/libs/surfaces/komplete_kontrol_a/interface.cc
+++ b/libs/surfaces/komplete_kontrol_a/interface.cc
@@ -1,5 +1,4 @@
 /*
- * Copyright (C) 2026 Ada <ada@6bit.com>
  * Copyright (C) 2026 Joshua Perry <josh@6bit.com>
  *
  * This program is free software; you can redistribute it and/or modify
--- a/libs/surfaces/komplete_kontrol_a/interface.cc
+++ b/libs/surfaces/komplete_kontrol_a/interface.cc
@@ -1,0 +1,76 @@
+/*
+ * Copyright (C) 2026 Ada <ada@6bit.com>
+ * Copyright (C) 2026 Joshua Perry <josh@6bit.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <stdexcept>
+
+#include "pbd/error.h"
+
+#include "ardour/rc_configuration.h"
+#include "control_protocol/control_protocol.h"
+
+#include "komplete_kontrol_a.h"
+
+using namespace ARDOUR;
+using namespace PBD;
+using namespace ArdourSurface;
+
+static ControlProtocol*
+new_komplete_kontrol_a (Session* s, std::string const& /* config */)
+{
+	KompleteKontrolA* kka = 0;
+
+	try {
+		kka = new KompleteKontrolA (*s);
+	} catch (std::exception& e) {
+		PBD::error << "Failed to instantiate Komplete Kontrol A-Series: " << e.what () << endmsg;
+		delete kka;
+		return 0;
+	}
+
+	kka->set_active (true);
+	return kka;
+}
+
+static void
+delete_komplete_kontrol_a (ControlProtocol* cp)
+{
+	delete cp;
+}
+
+static std::map<std::string, std::vector<std::string> >
+enumerate_komplete_kontrol_a ()
+{
+	return { { "Native Instruments", { "Komplete Kontrol A25",
+	                                   "Komplete Kontrol A49",
+	                                   "Komplete Kontrol A61" } } };
+}
+
+static ControlProtocolDescriptor komplete_kontrol_a_descriptor = {
+	/* name       */ "NI Komplete Kontrol A-Series",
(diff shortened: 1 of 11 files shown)


joshperry:
Had my bot write the PR, let me know if that's not ok and I can do it myself. Operating transport keys demo: https://imgur.com/a/ni-komplete-kontrol-ardour-UBIdrj9

I'm pretty sure feedback point 2 is correct, this is what it looks like in the UI:

image

x42:
So Ada is Claude's sister? :)

joshperry:
Yeah, she taught him everything he knows. She's been coding since before computers even existed!

libs/surfaces/komplete_kontrol_a/interface.cc
```diff
@@ -0,0 +1,87 @@
...
+using namespace PBD;
+using namespace ArdourSurface;
+
+static ControlProtocol*
+new_komplete_kontrol_a (Session* s, std::string const& /* config */)
+{
+	KompleteKontrolA* kka = 0;
+
+	try {
+		kka = new KompleteKontrolA (*s);
```
limitedAtonement: Ignoring `config` means that if the user selects "A25", if an A61 is available, it will be used, right?

libs/surfaces/komplete_kontrol_a/interface.cc
```diff
@@ -0,0 +1,87 @@
...
+		delete kka;
+		return 0;
+	}
+
+	/* Honour the result.  Ardour's activate() calls set_active() again, and
+	 * our set_active() re-runs start() whenever the protocol is not already
+	 * active -- so handing back an object that failed to start means the
+	 * device is opened twice and every failure is logged twice.  Reporting
+	 * the failure here instead gives one error and lets activate() fail
+	 * cleanly, which is also what makes the GUI checkbox behave.
```
limitedAtonement: initialize should return an inactive object and let `activate` start it once. See `libs/surfaces/mackie/interface.cc:44` for an example.

libs/surfaces/komplete_kontrol_a/komplete_kontrol_a.cc
```diff
@@ -0,0 +1,875 @@
...
+		error << _("Komplete Kontrol A-Series: failed to enter interactive mode") << endmsg;
+		/* Nothing was taken over, so there is nothing to hand back; a
+		 * graceful close here would only add three more failing writes.
+		 */
+		close_device (false);
+		return -1;
+	}
+
+	connect_session_signals ();
+	refresh_transport_leds ();
```
limitedAtonement: `take_over_device` just did this refresh call.

libs/surfaces/komplete_kontrol_a/komplete_kontrol_a.cc
```diff
@@ -0,0 +1,875 @@
...
+bool
+KompleteKontrolA::blink ()
+{
+	_blink_on = !_blink_on;
+	refresh_transport_leds ();
+	return true;
+}
+
+int
+KompleteKontrolA::stop ()
```
limitedAtonement, line 416: This is called from the GUI thread (called from `set_active` and destructor), but `sigc::connection::disconnect` is not according to https://gnome.pages.gitlab.gnome.org/gtkmm-documentation/chapter-multi-threaded-programs.html. It looks like you deal with `_handle` carefully using the `_device_mutex`; I think these signals need similar treatment. Probably get `stop` to trigger the disconnects in the surface context.

x42:
Please keep in mind that the ardour project does not accept LLM generated code at this point in time: https://ardour.org/development#llm

joshperry:
> Please keep in mind that the ardour project does not accept LLM generated code at this point in time: https://ardour.org/development#llm

That's fair, and if you don't want LLM generated code I respect that.

However, my read of the caselaw differs; the D.C. Circuit expressly declined to reach whether someone who builds and directs the system is the author, and the Copyright Office's Part 2 report (Jan 2025) treats that as an open question. So "no LLM-generated code may be copyrighted" seems broader than written.

So I'd only ask for you to perhaps reconsider your policy in the future. The [linux kernel policy](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-assistants.rst) is interesting as it sidesteps the ownership question by requiring that only humans sign off on the DCO, with an Assisted-by tag for disclosure. This may be similar to only a human claiming copyright in the file headers like I've fixed here.

joshperry:
@limitedAtonement I really appreciate your feedback! Really makes the code much stronger.

PR #1143, opened 2026-08-11
update german translation

edogawa23:


pauldavisthefirst:
manuall rebased and merged, thanks!

PR #1144, opened 2026-08-16
Replace source for audio regions

aivelon:
New dialog created to select audio clip from. 
Regions on the playlist are replaced by new regions with same properties but replaced source. 
Only enabled when audio region(s) are selected. Region length is reduced if the region is longer than the source clip it is being replaced by else length remains same.
Few changes done in the clip picker to show it correctly in the dialog and get the path of the clip selected. Also, it should not run audition as soon as its is instantiated.
Progress bar added for large number of regions.
This action is totally undoable/redoable.

https://github.com/user-attachments/assets/720e9106-654e-49b9-836b-20bdea5d6aea

For large number of clips:

https://github.com/user-attachments/assets/db1c49f2-732c-4d5c-abbc-22a604e3cc50

### DIFF
--- a/gtk2_ardour/ardour.menus.in
+++ b/gtk2_ardour/ardour.menus.in
@@ -935,6 +935,7 @@
       <menuitem action='split-region'/>
       <menuitem action='split-region-at-transients'/>
       <menuitem action='split-multichannel-region'/>
+      <menuitem action='replace-region-source'/>
       <menuitem action='close-region-gaps'/>
       <menuitem action='place-transient' />
       <menuitem action='show-rhythm-ferret'/>
--- a/gtk2_ardour/editor.h
+++ b/gtk2_ardour/editor.h
@@ -1171,6 +1171,7 @@ private:
 	void region_fill_track ();
 	void audition_playlist_region_standalone (std::shared_ptr<ARDOUR::Region>);
 	void split_multichannel_region();
+	void replace_source_for_selected_regions ();
 	void reverse_region ();
 	void strip_region_silence ();
 	void normalize_region ();
--- a/gtk2_ardour/editor_actions.cc
+++ b/gtk2_ardour/editor_actions.cc
@@ -1372,6 +1372,9 @@ Editor::register_region_actions ()
 	/* Open the pitch shift dialogue for any selected audio regions */
 	register_region_action (_region_actions, RegionActionTarget (SelectedRegions|EnteredRegions), "pitch-shift-region", _("Pitch Shift..."), sigc::mem_fun (*this, &Editor::pitch_shift_region));
 
+	/* Replace source clip for selected audio regions */
+	register_region_action (_region_actions, RegionActionTarget (SelectedRegions|EnteredRegions), "replace-region-source", _("Replace Source Clip"), sigc::mem_fun (*this, &Editor::replace_source_for_selected_regions));
+
 	/* Open the transpose dialogue for any selected MIDI regions */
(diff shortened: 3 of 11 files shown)


pauldavisthefirst:
We suspect that this code was written using an LLM. Please read http://ardour.org/development and confirm that it was not.

aivelon:
> We suspect that this code was written using an LLM. Please read http://ardour.org/development and confirm that it was not.

Sure, let me explain how I have developed this feature and where LLM was used (TLDR available below)
**Motivation**: I am an ardour user and I often find myself wanting to replace samples, specially for drum loops, which currently involves adding regions again with the new clip and arranging them again. This is quite tedious and after searching online, I saw that this is missing in ardour.

**Planning**: I first started planning this feature for which I used LLM. I first thought of dragging a clip from the right side clip panel and drop it on to the regions, but that already creates a new region, so I thought of doing it similar to how "Pitch shift" works on multiple regions with a dialog. The LLM suggested to use the already available "TriggerClipPicker" and show it in the dialog. That seemed like the best option to me. For the actual replace source operation, at first I looked at the source and master_source properties of a audio region and I was looking to replace those. The LLM suggested that that is not how its done in other places, like in 'pitch shift' feature, and replacing the region(s) with a new region is better. I honestly did not question this further (maybe I should have) and went with it. I was able to find the copy constructor for creating new region with different source, but the LLM pointed out that there is a RegionFactory::create() method available to copy region properties with new source. I did not plan or discuss anything about undo/redo at this stage.

**Development**: I started by adding the context menu entry which I easily found how to do by searching for other context menu entries. Then I set sensitive property to enable/disable it, similar to other menu entries. Then I created the editor method, where I created the dialog to hold the TriggerClipPicker. I did have to make some changes to the dialog for the issues I found during testing. The dialog would automatically start auditioning the first clip in the tree when it opens. The LLM helped diagnose this issue as it suggested to use signal_button_press_event to set a flag and stop the autoplay in the beginning which triggers without any mouse events, which I then implemented. I also made a small change on my own to remember which directory was selected in the clip picker and use that instead of always showing the first entry in the dropdown. Rest of the things in the clip picker I was able to add quite easily with some web searches, like to get the path of the selected clip. After adding a few validations on the selected audio clip and for the number of channels, it was already working as expected.
Then I tested and found that the undo/redo was not working. I dug deeper into the undo/redo in ardour on my own and found that only the MementoCommand was getting added for this action. I suspected that is because I am not changing a region's properties but replacing it. The LLM helped here to understand how to create a custom command using the PBD::Command base class and also pointed out that its needed in both gtk_ardour and libardour for UI undo/redo and for loading undo/redo on session start. I debated with LLM to create a single class definition but then ended up having different definitions for now (let me know if that needs to be changed). The implementation of the custom command is pretty straight forward since its just loading/saving to xml and implementing the operator and undo methods. For large number of clips, ardour was getting stuck without any progress indication. So, I added a simple dialog with progress bar. I took the code from 'strip_silence_dialog' to implement this.

**TLDR**: I am aware of the ardour policy of not using LLM to write the code and I have made a concious effort to not use LLM generated code here. I have used LLM extensively to get suggestions, find code and even get specific code suggestions/snippets, but at the end, I have written the actual code for implementation which is here in the PR.

aivelon:
I have had another look at my LLM usage and found that it has has done some modifications on the undo/history save parts. I  apologise for the oversight. Let me close this PR and I will create another one soon without any AI generated code.

PR #1146, opened 2026-08-23
Replace source for audio regions

aivelon:
New dialog created to select audio clip from.
Regions on the playlist are replaced by new regions with same properties but replaced source.
Only enabled when audio region(s) are selected. Region length is reduced if the region is longer than the source clip it is being replaced by else length remains same.
Few changes done in the clip picker to show it correctly in the dialog and get the path of the clip selected. Also, it should not run audition as soon as its is instantiated.
Progress bar added for large number of regions.
This action is totally undoable/redoable.
Made sure no LLM generated code is included.

https://github.com/user-attachments/assets/720e9106-654e-49b9-836b-20bdea5d6aea

For large number of clips:

https://github.com/user-attachments/assets/db1c49f2-732c-4d5c-abbc-22a604e3cc50

### DIFF
--- a/gtk2_ardour/ardour.menus.in
+++ b/gtk2_ardour/ardour.menus.in
@@ -935,6 +935,7 @@
       <menuitem action='split-region'/>
       <menuitem action='split-region-at-transients'/>
       <menuitem action='split-multichannel-region'/>
+      <menuitem action='replace-region-source'/>
       <menuitem action='close-region-gaps'/>
       <menuitem action='place-transient' />
       <menuitem action='show-rhythm-ferret'/>
--- a/gtk2_ardour/editor.h
+++ b/gtk2_ardour/editor.h
@@ -1171,6 +1171,7 @@ private:
 	void region_fill_track ();
 	void audition_playlist_region_standalone (std::shared_ptr<ARDOUR::Region>);
 	void split_multichannel_region();
+	void replace_source_for_selected_regions ();
 	void reverse_region ();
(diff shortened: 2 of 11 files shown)


GhostsonAcid:
I can't comment on your code, but having this sort of feature _built-in_ to Ardour would no doubt be very cool imo. 👍 

Last year I tackled the same functionality in my script here: [AudioClipboard - Manually Select Files To Use](https://github.com/GhostsonAcid/AudioClipboard/#manually-select-files-to-use) -In this case you basically just copy regions, delete them, then paste them back but point Ardour to use/embed/import a different source (or sources) to use.

Whether built-in to Ardour or not, however, some dev challenges include:

- How to handle files of a different sample-rate than the session and/or selected regions?
- How to handle importing vs embedding files? (-Give users a choice?)
- How to handle sources with different lengths?
- ...
- Probably more, but can't think right now... :P

-J

aivelon:
@GhostsonAcid Thanks for your input!!

> How to handle files of a different sample-rate than the session and/or selected regions?

This I need to look into. What do you think should happen in this case. I am thinking to not allow the operation in this case.

> How to handle importing vs embedding files? (-Give users a choice?)

Not really sure what the difference is here. I usually use the samples library to add audio to tracks and that is what I used here for source selection instead of a full file browser. So this implementation is basically same as dragging from the sources panel to the track to add audio region.

> How to handle sources with different lengths?

This I have handled. The region is resized based on the source length.

PR #1148, opened 2026-09-02
coreaudio: check _device_outs in the allocation guard

darkdi:
In `CoreAudioPCM::discover` the out-of-memory guard tests `_device_ins` twice, so a failed `calloc` for `_device_outs` walks past it. The `assert` on the line directly above already spells out the intended three, which is what made this an easy read rather than a guess. With NDEBUG the assert is gone and the null pointer reaches the device loop below, where `_device_outs[idx]` is written for every device found.

While I was there I noticed `libs/ardour/plugin_insert.cc:2401` has the same shape, `nout == 0 || inx.get(*t) == 0 || nout == 0`, but there the third operand is only redundant and nothing is missing, so I left it alone.


x42:
Thank you. Rebased and merged as 9.8-21 75a49c523c

PR #1151, opened 2026-09-06
VST3: do not treat a single .vst3 file as a module outside Windows

giang17:
`module_path_vst3()` skips `<name>.vst3/Contents/x86_64-win/<name>.vst3`, so that a single-file VST3 DLL is not picked up a second time inside its own bundle. That rule is compiled only on Windows (`#ifdef PLATFORM_WINDOWS`).

The same layout exists on Linux. yabridge's merged bundles place a symlink to the Windows DLL at `<name>.vst3/Contents/x86_64-win/<name>.vst3` (or `x86-win` for 32-bit plug-ins) next to the native `Contents/x86_64-linux/<name>.so`. Because `vst3_discover_from_path()` scans recursively, each such DLL is discovered as a separate single-file module: the scanner fails to `dlopen` it ("invalid ELF header"), the path lands in `vst3_x64_blacklist.txt`, and the Plugin Manager shows an error row next to the working bundle entry — one per bridged plug-in. The Plugin Manager's search matches name, type, creator, tags, file or path, not the scan status, so these rows cannot be hidden.

A single `.vst3` file was only ever a valid module on Windows (the legacy single dll, deprecated since VST 3.6.10), so on the other platforms every `.vst3` that is not a bundle directory is now ignored. The Windows code path, including the same-name rule, is unchanged.

Verified on Ubuntu 24.04 with 149 yabridge bundles carrying 150 bundled DLL symlinks (one of them `x86-win`): the start-up scan of the unpatched distribution package (8.4.0) lists all 150 DLL paths, a build of the 8.4 tag with this change lists none, and the 149 bundle entries are the same in both. The branch is based on `master` (ba38f08).

### DIFF
--- a/libs/ardour/vst3_scan.cc
+++ b/libs/ardour/vst3_scan.cc
@@ -369,6 +369,16 @@ ARDOUR::module_path_vst3 (string const& path)
 #endif
 			return "-1";
 		}
+#else
+		/* A single .vst3 file was only ever a valid module on Windows. Here it
+		 * is the Windows dll of a multi-platform bundle (Contents/x86_64-win/),
+		 * found because the folder is scanned recursively; the bundle itself is
+		 * used instead.
+		 */
+#ifndef NDEBUG
+		cerr << "Ignore .vst3 file (not a bundle) '" << path << "'\n";
+#endif
+		return "-1";
 #endif
 		module_path = path;
 	} else {


x42:
According to the VST3 spec this was only legal in early days on Windows, but never allowed on other platforms.

So it seems the problem here is with yabridge and old VST3s that are not in a bundle by itself?

giang17:
Partly. On the Windows side both kinds exist: of the 150 bridged Windows modules here (149 bundles; one plug-in ships a 32-bit and a 64-bit dll), 120 are legacy single-file dlls (`C:\…\Common Files\VST3\<name>.vst3` as a file) and 30 sit inside a bundle's `Contents\x86_64-win\` folder.

But yabridgectl does not expose either of them as such. For every plug-in it builds a merged bundle on the Linux side, `~/.vst3/yabridge/<name>.vst3/Contents/x86_64-linux/<name>.so` (`i386-linux` for 32-bit) plus a symlink to the Windows module at `Contents/x86_64-win/<name>.vst3` (`tools/yabridgectl/src/files.rs`, "part of the merged VST3 bundle"), regardless of how the Windows plug-in is laid out. That is the cross-platform bundle from the spec's *Plugin Format* page, with `x86_64-win` being the folder the spec assigns to the Windows dll. What Ardour then picks up is not a bare legacy dll but the Windows module *inside* a bundle, because the scan descends into the bundle and the Linux build accepts a `.vst3` file as a module.

Given that a single `.vst3` file was never a valid module outside Windows, the stricter fix would be to ignore every non-bundle `.vst3` on the other platforms and leave the Windows same-name rule alone. The branch has been updated to that form (Windows code path unchanged; on Linux/macOS a `.vst3` that is not a directory returns `-1`). Same result with the 149 bundles here: no error rows, 149 bundle entries unchanged, and a stray `.vst3` file outside any bundle is now ignored as well.

x42:
Fair. I expect it also does not hurt to apply the rule on other platforms.

PR #1152, opened 2026-09-09
update Chinese translation

brucekomike:
bring contents up-to-date since the partially translated version of ardour 6.x.


brucekomike:
this translation still needs validating.

PR #1157, opened 2026-09-17
Fixed /is_recording OSC feedback (prev. was not sent over network)

jzemgulys:
The /is_recording OSC message did not previously have a corresponding function call to actually send it over the network when feedback is enabled.


x42:
Rebased and merged as  9.8-41 d58bcb7bc5

PR #1159, opened 2026-09-22
Fix VST3 X11 resize requisition

DevJac:
The GtkSocket.queue_resize() docs say it "should be called when a widget for some reason has a new size request". This was not happening for widgets that had `view->canResize() == kResultTrue`.

This was blocking the ability for plugins to request a new size that was smaller than their initial size. This was sometimes causing a bug where plugins would initially be sized very large and could never be shrinked, even the plugin itself could not choose to shrink itself.

Oddly, it was only the views that had `canResize` set to false that could resize themselves.

Now we store the requested size, and also call `queue_resize()` unconditionally.

Personally, this fixed a bug I was having with Surge XT, where it would sometimes start very large, so large it couldn't fit on my screen. You can find other people online reporting the same problem. Surge XT is able to shrink itself, but Ardour would not allow it to. Ardour would not allow the plugin to have a window that was smaller than its initial window size.

This fixes that bug.

I've done a lot of research about this and I think there's a big mess related to properly integrating X11 and Ardour / GTK VST resizing. I don't know, it was all quite confusing to me with a lot of dead ends.

My ultimate impression was that the overall Ardour approach to VST windows with VST3s on X11 wasn't ideal, but despite trying, I wasn't able to come up with anything better, other than this small workaround, which I do think makes the situation a little better.

At the very least we can respect the plugin's own resize requests instead of stubbornly sticking to the initial size as the absolute minimum size forever.


x42:
>  there's a big mess related to properly integrating X11 and Ardour / GTK VST resizing

Indeed, the GTK concept of _size-request_, _size-allocate_ is conceptually at odds with VST3, particularly with `IPlugView::checkSizeConstraint`

Ideally `IPlugFrame::resizeView` would only result in a `gtk_widget_queue_resize()` call, which then asynchronously does a size-request and allocation, informing the plugin about the result...

It does seem odd that this fix works, in case of a resizable window `_gui_widget.queue_resize` should not be needed, there is an explicit call to set the window size `get_window()->resize(..)` which  overrides the size_request. `gtk_window_resize` does call `gtk_widget_queue_resize_no_redraw()` as well.

Could you check if this modified patch also helps:
https://gist.github.com/x42/34b5d20c673f95c71df4acab2912b6ab

DevJac:
The modified patch you proposed also fixes the issue I was having.

I made some style changes and force pushed your patch to this branch. I restructured the code slightly to avoid the early return nested inside of the 2 outer if-conditions. (I like an explicit if-else-statement, so long as it doesn't require putting lots of code nested inside the else block.)

It's a matter of style. If you prefer the way it was in your patch, just say so and I will update the PR to match your patch exactly.

Thank you for taking the time to look into this.

x42:
Thank you! Rebased and merged as 9.8-74 bc607be0b7

PR #1160, opened 2026-09-22
session-utils: add CLI export for named routes and buses

NickHooked:
## Motivation

I wanted to export some stems that I had forgotten to export while I still had access to the Ardour GUI. Later, with only SSH access to the machine, I found that this was not possible. This prompted me to add support for exporting individual routes and buses from the command line.

## Summary

- add a `--route` option to export a named route or bus without requiring a master bus
- add configurable plugin warmup time for asynchronously initialized instruments
- return failure for missing or non-audio routes instead of dereferencing a missing master bus
- include a local launcher and usage notes for the Ardour 9 runtime layout

## Verification

- built the export utility and dummy audio backend successfully
- exported named stereo and mono routes successfully
- verified exports containing an asynchronously initialized instrument using the warmup option
- compared headless output with the equivalent GUI export for matching duration and levels
- verified that a missing master bus or unknown route returns a failure status without crashing
- `git diff --check` passes


x42:
Looks good. 

Except for  the warmup time. 
Plugins should all be instantiated and running once the session is loaded.
What plugin(s) have you found that require this? What warmup time do you set for those?

There is a config variable _export-preroll_ (search for `get_export_preroll`) which is may handle this.  It is motivated as work-around by reverb plugins that don't report a tail-time. The default is 2 seconds.

NickHooked:
The plugin was Speedrum Lite, a VST3 drum sampler. Without a delay after `load_session()`, its routed output was completely silent. Waiting five seconds before starting the export produced the expected audio and matched the equivalent GUI export in duration and levels.

My understanding is that `export-preroll` is processing-time preroll after export starts, while this plugin appeared to need wall-clock time to finish asynchronous sample loading. I’ll verify whether `export-preroll` also solves this case. If it does, I’ll remove the separate warmup option.

NickHooked:
I tested this again on Ardour’s current `master`, using an isolated session containing Speedrum Lite routed directly to a regular stereo master bus. I used the unmodified `session_utils/export.cc` and the normal master-bus export path, so no `--route` code was involved.

Results:

- Default 2-second `export-preroll`, with no wall-clock delay after `load_session()`: silent output (`-91.0 dB`).
- `export-preroll` increased to 5 seconds, still without a wall-clock delay: silent output (`-91.0 dB`).
- Default `export-preroll` plus a 5-second wall-clock delay after `load_session()`: expected audio (`-44.5 dB` mean, `-6.6 dB` peak).

All exports had the same duration. This indicates that the issue already occurs on `master` with the existing master-bus export and is not caused by the new route-export path. Speedrum appears to load its samples asynchronously and needs actual elapsed time; offline export preroll is processed too quickly to provide that.

I can provide the test session if that would be useful.

pauldavisthefirst:
The .get() on the std::shared_ptr<IO> is unnecessary. a shared ptr will convert to boolean just like a pointer.

